PR #5, opened 2017-01-02
Moved logic from the CKEditorWebpackPlugin to resolveImportPath.

### LINKED ISSUE #4: Webpack plugin should be refactored
Logic should be moved to `resolveimportpath.js` to enable using this stuff in Rollup plugin.

ma2ciek:
Resolves #4.


packages/ckeditor5-dev-utils/lib/compiler/resolveimportpath.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {String} importPath Path to the imported file.
+ * @param {Array.<String>} packagePaths Array of node_modules directories and packages where packages will be looking for.
+ * @returns {Object|null}
+ */
+module.exports = function resolveImportPath( requesterPath, importPath, packagePaths ) {
+	const requestPackageName = getWorkspaceRelativePathInfo( importPath ).packageName;
+	let resolvedPath;
+
+	for ( let contextPackagePath of packagePaths ) {
+		if ( resolvedPath ) {
```
Reinmar: Unnecessary. You can return the resolved value in the previous loop iteration (at the end of this loop).

packages/ckeditor5-dev-utils/lib/compiler/resolveimportpath.js
```diff
@@ -0,0 +1,70 @@
...
+const fs = require( 'fs' );
+const getWorkspaceRelativePathInfo = require( './getworkspacerelativepathinfo' );
+
+/**
+ * Looks for the imported path in context of the requester path and returns informations about it when the request matches some file.
+ *
+ * @param {String} requesterPath Path to the requester.
+ * @param {String} importPath Path to the imported file.
+ * @param {Array.<String>} packagePaths Array of node_modules directories and packages where packages will be looking for.
+ * @returns {Object|null}
```
Reinmar: The return value misses documentation for the object properties.

packages/ckeditor5-dev-utils/lib/compiler/resolveimportpath.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+module.exports = function resolveImportPath( requesterPath, importPath, packagePaths ) {
+	const requestPackageName = getWorkspaceRelativePathInfo( importPath ).packageName;
```
Reinmar: This code makes little sense when you read it – import path can't be "workspace relative". Besides, it would be `requestedPackageName`.

So, if anything, there could be a generic function called `parseImportPath()`.

packages/ckeditor5-dev-utils/lib/compiler/resolveimportpath.js
```diff
@@ -0,0 +1,70 @@
...
+
+	for ( let contextPackagePath of packagePaths ) {
+		if ( resolvedPath ) {
+			break;
+		}
+
+		const chunks = contextPackagePath.split( path.sep );
+
+		// Current request package is the main package.
+		if ( chunks[ chunks.length - 1 ] === requestPackageName ) {
```
Reinmar: Why is this code outside `resolveImportPathInContext`? It means that a piece of logic is here in this one and a piece in the other function. It's, of course, ok to split code into multiple functions, but there must be a clear line what goes where.

packages/ckeditor5-dev-utils/lib/compiler/resolveimportpath.js
```diff
@@ -0,0 +1,70 @@
...
+
+const path = require( 'path' );
+const fs = require( 'fs' );
+const getWorkspaceRelativePathInfo = require( './getworkspacerelativepathinfo' );
+
+/**
+ * Looks for the imported path in context of the requester path and returns informations about it when the request matches some file.
+ *
+ * @param {String} requesterPath Path to the requester.
+ * @param {String} importPath Path to the imported file.
```
Reinmar: The path to be resolved.

packages/ckeditor5-dev-utils/lib/compiler/resolveimportpath.js
```diff
@@ -0,0 +1,70 @@
...
+ * @param {Array.<String>} packagePaths Array of node_modules directories and packages where packages will be looking for.
```
Reinmar: An array of directories where the algorithm will look for packages. You can either pass path to directories containing packages or to a package itself. If a name of the package to import matches the name of the passed directory, then this path will be used as a path to package. If not, then packages will be looked for in the specified directory.

Reinmar:
While working on https://github.com/ckeditor/ckeditor5-dev/issues/6 I realised that the algorithm we implemented, or rather, the configuration format that I proposed isn't that smart. Sorry :P.

The reason for that is support for scoped packages (where name of the directory won't match the name of the import (e.g. we import `@cke/cke5-x` but we keep this package in `packages/cke5-x/`). 

I think that something like this would make more sense:

```
packages: {
  'ckeditor5-foo': '/sth/sth/foo',
  '@ckeditor/ckeditor5-bar': '/sth/sth/packages',
  '@ckeditor/*: [
    '/sth/sth/packages',
    '/els/els/packages'
  ],
  '*': [
    '/sth/sth/packages', 
    ...
  ]
}

The idea is that if the key ends with '*' it means to look for packages in that place.

E.g., for the above configuration following matches can be made (assuming that specified directories exist in `sth/sth/packages/`):

* `@ckeditor/ckeditor5-foo` to `/sth/sth/packages/ckeditor5-foo` (using `'@ckeditor/*'`)
* `@foo/bar` to `/sth/sth/packages/@foo/bar` (using the `'*'`)
* `xyz` to `/sth/sth/packages/xyz` (using the `'*'`)

WDYT?

Reinmar:
We're not gonna need it, so let's close the PR (the ticket was closed too).

PR #8, opened 2017-01-04
Tool for generating changelog and creating releases

pomek:
Fixes: #2.


pomek:
I introduced a lot of tests for the new code but `tasks.createRelease` and `tasks.generateChangelog` still don't have tests.

The tasks are really big and tests will be bigger. We should consider whether it can be split into smaller modules.

Reinmar:
> I introduced a lot of tests for the new code but tasks.createRelease and tasks.generateChangelog still don't have tests.
> 
> The tasks are really big and tests will be bigger. We should consider whether it can be split into smaller modules.

Definitely. Those functions are *far* too big. It'd be super hard to tests them now. But we'll be working on this later.

Reinmar:
My approach to this review will be to read the code, but I won't try to validate these tools. We'll work on that when releasing the next version of CKEditor 5.

Reinmar:
OK, I changed my mind :D I'll do the review now, because there's a bit too many of the small things which will be hard to polish later.

packages/ckeditor5-dev-env/lib/utils/changelog.js
```diff
@@ -0,0 +1,249 @@
+/**
```
Reinmar: Let's not create such a big utils files. Split it into smaller ones, preferably a function per file (but sometimes it makes a bit more sense to group them). And do this after https://github.com/ckeditor/ckeditor5-dev/pull/8/files#r97547555, because then you'll see what's the best way to shape the code.

packages/ckeditor5-dev-env/lib/utils/changelog.js
```diff
@@ -0,0 +1,249 @@
...
+				'debug'
+			],
+
+			default: {
+				init: false,
+				debug: false
+			}
+		} );
+
+		delete options._;
```
Reinmar: Please remove this or comment it and write a test for it.

Hint: Since it'd be a waste of time to test it it means that this code makes no sense and could be removed :D

packages/ckeditor5-dev-env/lib/utils/changelog.js
```diff
@@ -0,0 +1,249 @@
...
+				}
+
+				changelog = changelog.replace( new RegExp( `^## \\[?${ currentTag }\\]?.*` ), '' ).trim();
+
+				return Promise.resolve( changelog );
+			} );
+	},
+
+	/**
+	 * @returns {Promise}
```
Reinmar: If the only thing this function does is reading a file from the system, please get rid of the promise. I've been writing this on and on in many PRs – use synchronous versions of Node.js methods.

PS. It took me 15s to try to understand what this function does. And I'm still unsure, which is exactly why KISS!

packages/ckeditor5-dev-env/lib/utils/changelog.js
```diff
@@ -0,0 +1,249 @@
...
+				resolve( content );
+			} );
+		} );
+	},
+
+	/**
+	 * @param {String} content
+	 * @returns {Promise}
+	 */
+	saveChangelog( content ) {
```
Reinmar: The same as above.

packages/ckeditor5-dev-env/lib/utils/changelog.js
```diff
@@ -0,0 +1,249 @@
...
+		const changelogFile = path.resolve( utils.changelogFile );
```
Reinmar: The same as above.

packages/ckeditor5-dev-env/lib/utils/changelog.js
```diff
@@ -0,0 +1,249 @@
...
+
+				return Promise.resolve( changelog );
+			} );
+	},
+
+	/**
+	 * @returns {Promise}
+	 */
+	getCurrentChangelog() {
+		const changelogFile = path.resolve( utils.changelogFile );
```
Reinmar: This doesn't seem to be necessary.

packages/ckeditor5-dev-env/lib/utils/changelog.js
```diff
@@ -0,0 +1,249 @@
...
+	 * @param {String} token Token used to authenticate with GitHub.
+	 * @param {Object} options
+	 * @param {String} options.repositoryOwner Owner of the repository.
+	 * @param {String} options.repositoryName Repository name.
+	 * @param {String} options.version Name of tag connected with the release.
+	 * @param {String} options.description Description of the release.
+	 * @param {String} options.debug Whether to display additional logs.
+	 * @returns {Promise}
+	 */
+	createGithubRelease( token, options ) {
```
Reinmar: This isn't a util. This is a specific task. It doesn't mean that it should end in the tasks.js file, but that it shouldn't be in the utils :P.

The utils hole is the last place where you want a file to end. And it must meet certain requirements, like having no requirements of its own other than other utils.

packages/ckeditor5-dev-env/lib/utils/exec-on-dependencies.js
```diff
@@ -0,0 +1,53 @@
+/**
```
Reinmar: Wrong file and function names. Should be `executeOnDependencies()`

packages/ckeditor5-dev-env/lib/utils/exec-on-dependencies.js
```diff
@@ -0,0 +1,53 @@
...
+		const directoryPackageJson = require( path.join( dependencyPath, 'package.json' ) );
+		const dependencyName = directoryPackageJson.name;
+
+		promise = promise.then( () => {
+			return functionToExecute( dependencyName, dependencyPath );
+		} );
+	}
+
+	return promise.then( () => {
+		process.chdir( currentCwd );
```
Reinmar: Why is it resetting the cwd if it's not changing it?

packages/ckeditor5-dev-env/lib/utils/exec-on-dependencies.js
```diff
@@ -0,0 +1,53 @@
...
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+const path = require( 'path' );
+const { workspace } = require( '@ckeditor/ckeditor5-dev-utils' );
+
+/**
+ * The module allows running a function on their locally installed dependencies.
```
Reinmar: It's not a module.

packages/ckeditor5-dev-env/lib/utils/proposed-deps-version.js
```diff
@@ -0,0 +1,43 @@
+/**
```
Reinmar: Wrong file and function names. `proposedDependenciesVersions()`

packages/ckeditor5-dev-env/tests/fixtures/package.json
```diff
@@ -0,0 +1,8 @@
+{
```
Reinmar: Don't assume that there will be just on fixture project needed. Wrap it in `basic/` or similar directory.

packages/ckeditor5-dev-env/tests/fixtures/package.json
```diff
@@ -0,0 +1,8 @@
...
+	"version": "3.1.2",
+	"dependencies": {
+		"ckeditor5-core": "ckeditor/ckeditor5-core",
```
Reinmar: Wrong names and versions (we're using scoped and versioned deps already).

packages/ckeditor5-dev-env/tests/fixtures/packages/ckeditor5-core/package.json
```diff
@@ -0,0 +1,3 @@
+{
+	"name": "ckeditor5-core"
```
Reinmar: Should be scoped.

packages/ckeditor5-dev-env/lib/utils/exec-on-dependencies.js
pomek: `functionToExecute` can change the cwd.

packages/ckeditor5-dev-env/tests/fixtures/package.json
pomek: This code has been written a few weeks before we started to use the scoped :D

packages/ckeditor5-dev-env/tests/fixtures/packages/ckeditor5-core/package.json
pomek: This is a fixture which checks whether the correct name is passed to function.

packages/ckeditor5-dev-env/lib/utils/exec-on-dependencies.js
Reinmar: So the function itself should fix the dir.

pomek:
Because `tasks.createRelease` is split into small modules, I wrote tests for it.

Unfortunately, I had problems with tests for `tasks.generateChangelog`. I couldn't create a stream mock which will be compatible with our `stream.noop()` helper.

Reinmar:
OK, I made a review. I haven't been able to check the whole code, because it's too much of it, but I checked how the tool works and I fixed some issues. I also aligned the setup to our needs.

However, I found one critical issue. According to the spec, the tool should only read commits made directly to the master branch, without going into commits done in in ticket branches (in those cases it should only check the merge commits). How the tool works now is completely the opposite – it checks every commit, including ticket branches, but it ignores merge commits. This needs to be fixed before we close this PR.

Reinmar:
image

```
(master 1dbd0e5) p@m /workspace/test-releases/ckeditor5> git log --first-parent master
commit 1dbd0e5da426974902cb5f973b4fd6e3e0918cce
Author: Piotrek Koszuliński <pkoszulinski@gmail.com>
Date:   Fri Jan 27 17:32:14 2017 +0100

    Removed unreleased changelog entries.

commit 6038d4c0cdfe1571c9bb4803471f79b536eb9b28
Merge: bd2e76b 56f0ebe
Author: Piotrek Koszuliński <pkoszulinski@gmail.com>
Date:   Fri Jan 27 17:30:18 2017 +0100

    Fix: This should count.

commit bd2e76b678de3ee86a2a10ce52d3431ee067d37b
Author: Piotrek Koszuliński <pkoszulinski@gmail.com>
Date:   Fri Jan 27 17:12:41 2017 +0100

    Fix: Aligned gulpfile to changes in ckeditor5-dev-env.

commit c9e65fc60a84504ad8ca51d758a66aaa3bee3edd
Author: Piotrek Koszuliński <pkoszulinski@gmail.com>
Date:   Fri Jan 27 16:14:37 2017 +0100

    Docs: Changelog.

```

Reinmar:
I'm only a bit unsure how to process merge commits. In the example above I deleted the original message, which normally look like this:

```
commit c41f115d11a6a092f82d42c1bf502b7f613d5d77
Merge: 1ea8f69 03dcb6b
Author: Piotrek Koszuliński <pkoszulinski@gmail.com>
Date:   Fri Dec 16 14:39:11 2016 +0100

    Merge pull request #382 from ckeditor/t/381

    Remove unnecessary tasks.
```

The "Merge pull request ..." line is pretty useful and I'd rather like to keep it, which means that our  merge commits should look like this:

```
Merge pull request #382 from ckeditor/t/381

Fix: This should be the message for the changelog. Fixes #123.
```

Let's see how to pass it to conventional changelog ;>

pomek:
I fixed it.

For the following tree:

image

I tried to modify the configuration for `gitRawCommits`.

### The first one - `{ merges: false }` (by default):

image

**Merge commits are skipped but ticket commits are available.**

### The second one - `{ merges: true }`:

image

**Merge commits are available and ticket commits are hidden.**

---

I've created another branch, pushed two commits to them:

image

Then I merged this branch into the master. I updated the message of merge commit:

image

And then I generated the changelog:

image

image

---

So, the ticket commits are skipped, merge commits are parsed and it works as we want.

pomek:
Finally, I did it. Git takes commits from a master and also merge commits into the master.

image

packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js
```diff
@@ -47,6 +48,16 @@ module.exports = {
 // - filters out the commit if it should not be visible in the changelog,
 // - makes links to issues and user's profiles on GitHub.
 function transformCommit( commit ) {
+	if ( commit.header.startsWith( 'Merge pull request' ) ) {
```
Reinmar: What about merges done from the console? The message is just "Merge ...", so I'll change this check to `startsWith( 'Merge' )`

pomek: Done.

Reinmar:
LGTM now. I have some followups to report, but the tasks do their job pretty much correctly now. Good job!

PR #54, opened 2017-01-23
Tool for saving history state from CKE5 deps.

pomek:
Before the merge, this line -  https://github.com/ckeditor/ckeditor5-dev/pull/54/files#diff-3ec487e2912c226fac4db47bfd56242bR10 - needs to be reverted.

---

```
Feature: Added script for saving revision log of CKEditor 5's dependencies. Resolves #54.
```


pomek:
Commits created by Travis: https://github.com/ckeditor/ckeditor5/compare/master-revisions

The process was triggered by [`ckeditor5-autoformat`](https://travis-ci.org/ckeditor/ckeditor5-autoformat/builds/194427851).

pomek:
Commit pushed to `ckeditor5#master-revisions` triggers a build for CKE5: https://travis-ci.org/ckeditor/ckeditor5/builds/194428720

pomek:
We need to set manually environment variables for each package on Travis:

* GITHUB_USER,
* GITHUB_PASSWORD.

packages/ckeditor5-dev-env/bin/save-revision.js
```diff
@@ -0,0 +1,94 @@
...
+				let packageRepo = packageName;
+
+				if ( packageRepo.startsWith( '@' ) ) {
+					packageRepo = packageName.slice( 1 );
+				}
+
+				json.dependencies[ packageName ] = packageRepo + '#' + depsHashes[ packageName ];
+			}
+
+			json.dependencies[ repository ] = repository + '#' + process.env.TRAVIS_COMMIT;
```
pomek: In order to avoid invalid keys in [`mgit.json`](https://github.com/ckeditor/ckeditor5/blob/master-revisions/mgit.json#L22),`repository` in `json. dependencies ` should starting with '@'.

pomek:
After introducing the `save-hashes` command in Mgit, this PR can be simplified. I'll work on it tomorrow.

pomek:
[Mgit has been released (v0.5.0)](https://github.com/cksource/mgit2/releases/tag/v0.5.0). This PR is ready to review.

pomek:
After successful build on Travis, we want to update repository hashes in `mgit.json` file in the main repository.

What does this task do?

1. Clone main repository.
2. Run `mgit boostrap`
3. Check out each package to the `#master`
4. Collect hashes (last commit for each package)
5. If the `mgit.json` has changed - commit the changes.

The whole process is executed after successful build on Travis.

pomek:
Commit created a moment ago - https://github.com/ckeditor/ckeditor5/commit/77a2a4c2eccac06a19be9c29c4b868039eee87ee.

Reinmar:
https://github.com/ckeditor/ckeditor5-dev/pull/54#issuecomment-274460552 Do we need that? Cause I see that Travis CI User can commit these: https://github.com/ckeditor/ckeditor5/commit/77a2a4c2eccac06a19be9c29c4b868039eee87ee

Reinmar:
Other than that, the task look like we could use it right away.

However, I'd do just one more thing – move it to ckeditor5-dev-tests. The reason for that is that then we won't need to install ckeditor5-dev-env on CI and ckeditor5-dev-env is pretty big (with dependencies) so it would slow our CI down.

pomek:
> #54 (comment) Do we need that? Cause I see that Travis CI User can commit these: ckeditor/ckeditor5@77a2a4c

Travis can commit because of that - https://github.com/ckeditor/ckeditor5-dev/pull/54/files#diff-3ec487e2912c226fac4db47bfd56242bR23

Reinmar:
> Travis can commit because of that - https://github.com/ckeditor/ckeditor5-dev/pull/54/files#diff-3ec487e2912c226fac4db47bfd56242bR23

But you wrote that these settings need to be set. So I don't get it. Can Travis commit out of the box or do we have to set anything?

pomek:
Travis can commit because we gave him access to our "bot" account (which has access to @ckeditor). And we need to pass a login and password because ssh key does not work (or I don't know how to set it correctly).

pomek:
https://travis-ci.org/ckeditor/ckeditor5-autoformat/builds/209711612

https://github.com/ckeditor/ckeditor5/commit/644438bc1240f79cc0a1e9a0a2071214235e1c47

packages/ckeditor5-dev-tests/bin/save-revision.js
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env node
+
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+const branch = 'master'; // process.env.TRAVIS_BRANCH;
```
pomek: Before merging - replace `master` to value from comment.

Reinmar: But we don't want to handle all branches now. Just master. Later on we can add `major` too. The tool should ignore all the other branches. Actually... This is missing now in this PR.

Reinmar: Oops. This is checked below: https://github.com/ckeditor/ckeditor5-dev/pull/54/files/85204c96227bfe75ae5aad04cd8f3ba56bffe61f#diff-d697025760f8f24c3177d8ba07dacdaeR12 :D

PR #60, opened 2017-01-30
Scripts for collecting, uploading and downloading translations.

### LINKED ISSUE #44: Create scripts which collect, upload and download translations
Required by https://github.com/ckeditor/ckeditor5/issues/387 (step 2.)

ma2ciek:
Resolves #44.
Resolves main part of the https://github.com/ckeditor/ckeditor5/issues/387.


ma2ciek:
I'll cut off few "working" commits and push branch again.

pomek:
I'm sensing a conflict after merging #8. Or, #8 will have a conflict after merging this PR.

ma2ciek:
@pomek I modified `packages/ckeditor5-dev-env/lib/index.js` file as well to add translation tasks. But it shouldn't be hard to merge these 2 versions. Probably `package.json` differs too.

Reinmar:
Let's close @pomek's changes first and see then what to do with this.

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
```diff
@@ -0,0 +1,202 @@
...
+		for ( const filePath of files ) {
+			const content =  fs.readFileSync( filePath, 'utf-8' );
+
+			translations.push( ...utils._getTranslationCallsFromFile( filePath, content ) );
+		}
+
+		return translations;
+	},
+
+	_getTranslationCallsFromFile( filePath, fileContent ) {
```
Reinmar: Private and protected functions should go after public ones.

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
```diff
@@ -0,0 +1,202 @@
...
+
+const ckeditor5PackagesDir = path.join( process.cwd(), 'packages' );
+const langContextSuffix = path.join( 'lang', 'contexts.json' );
+const corePackageName = 'ckeditor5-core';
+
+const utils = {
+	/**
+	 * Collect translations and returns array of translations.
+	 *
+	 * @returns {Array<Object>}
```
Reinmar: This is a correct format, but we use its alternative, with a period after `Array`: `Array.<Object>`.

Reinmar: Besides, you're mixing this too.

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
```diff
@@ -0,0 +1,202 @@
...
+
+			if ( errorMessage ) {
+				errors.push( errorMessage );
+			}
+		}
+
+		return errors;
+	},
+
+	maybeGetContextErrorMessage( contexts, translation ) {
```
Reinmar: Odd function. Couldn't be something like `validateContexts()` which logs the error or does something else (can return it too if it must)?

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
```diff
@@ -0,0 +1,202 @@
...
+		const packageContext = contexts.get( translation.package );
+		const corePackageContext = contexts.get( corePackageName );
+		let error;
+
+		if ( !corePackageContext ) {
+			error = `${corePackageName}/lang/contexts.json file is missing.`;
```
Reinmar: Missing spaces inside `${}`.

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
```diff
@@ -0,0 +1,202 @@
...
+	getRepeatedContextErrorMessages( contexts ) {
+		const errors = [];
+		const keys = new Set();
+
+		for ( const context of contexts.values() ) {
+			for ( const key in context.content ) {
+				if ( keys.has( key ) ) {
+					errors.push( `Context is duplicated for the key: ${ key }.` );
+				}
+				keys.add( key );
```
Reinmar: Missing empty line after if statement. That should be caught by JSCS.

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
```diff
@@ -0,0 +1,202 @@
...
+			].map( x => x + '\n' ).join( '' );
+		} ).join( '\n' );
+	},
+
+	savePotFile( packageName, fileContent ) {
+		const outputFilePath = path.join( process.cwd(), 'build', '.transifex', packageName, 'en.pot' );
+
+		fs.outputFileSync( outputFilePath, fileContent );
+
+		logger.info( `Created file: ${ outputFilePath }` );
```
Reinmar: Didn't I comment on this already? :D Maybe it was in @pomek's PR. If the console log says something different than the function name then you know that something's wrong. In this case it's the log message.

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
```diff
@@ -0,0 +1,202 @@
...
+	},
+
+	createPotFileHeader() {
+		return '# Copyright (c) Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.\n\n';
```
Reinmar: Get the end year from the system.

packages/ckeditor5-dev-webpack-plugin/lib/replacetranslationcallsforonelanguage.js
```diff
@@ -0,0 +1,50 @@
...
+
+		for ( const assetName in options.assets ) {
+			replaceTCalls( options.assets[ assetName ], allTranslations );
+		}
+
+		done();
+	} );
+};
+
+function replaceTCalls( assetContent, allTranslations ) {
```
Reinmar: This, or something similar should be exposed by dev-utils. Not a meaningless functions like `replaceFirstFunctionParameter`. It must be fairly easy to build a plugin for bundler X based on our utils. The smaller the interface exposed by dev-utils, the better.

packages/ckeditor5-dev-utils/lib/index.js
```diff
@@ -10,5 +10,8 @@ module.exports = {
 	logger: require( './logger' ),
 	tools: require( './tools' ),
 	stream: require( './stream' ),
-	workspace: require( './workspace' )
+	workspace: require( './workspace' ),
+	getAllTranslations: require( './translations/getalltranslations' ),
+	createTranslator: require( './translations/createtranslator' ),
```
Reinmar: What's `createTranslator()`? We need a function which for a given source code returns a new source code with all `t()` calls replaced. We may later think on how to integrate source map synchronisation with that – perhaps this function should accept a callback which is executed for each found `t()` or something, but it must provide you all data already. 

Besides, wrap the translation tools in an object, just like the other tools are wrapped.

packages/ckeditor5-dev-webpack-plugin/lib/replacetranslationcallsforonelanguage.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param {Object} compiler Webpack compiler.
+ * @param {String} language Language code, e.g en_US.
+ */
+module.exports = function replaceTranslationCallsForOneLangauge( compiler, language ) {
+	const packageNames = new Set();
+
+	compiler.plugin( 'after-resolvers', ( compiler ) => {
+		compiler.resolvers.normal.plugin( 'before-resolve', ( obj, done ) => {
+			const match = obj.request.match( /@ckeditor[\\\/]([^\\\/]+)/ );
```
Reinmar: Do we have to care for `\`? Won't it always be `/`?

packages/ckeditor5-dev-webpack-plugin/lib/replacetranslationcallsforonelanguage.js
```diff
@@ -0,0 +1,50 @@
...
+			if ( match ) {
+				packageNames.add( match[ 1 ] );
+			}
+
+			done();
+		} );
+	} );
+
+	compiler.plugin( 'emit', ( options, done ) => {
+		const allTranslations = getAllTranslations( packageNames, language );
```
Reinmar: The `getAllTranslations()` function cannot hardcode how the path to package folders is resolved. How do you know how Webpack has been configured? It should use the paths which Webpack uses (so resolved by Webpack).

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
ma2ciek: It should return errors, because only then It's possible to break execution of the rest code and prevents gathering wrong po files.

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
ma2ciek: It was in my other PR, sorry ;)

packages/ckeditor5-dev-env/lib/index.js
```diff
@@ -67,6 +67,26 @@ const tasks = {
 			.then( () => {
 				process.chdir( options.cwd );
 			} );
+	},
+
+	collectTranslations() {
```
Reinmar: Missing docs.

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
ma2ciek: But I rethink this and `Created file: workspace/file.js` sounds good to me. Even better than `Created file.js at workspace/`. And `outputFilePath` is still a correct name for the variable.

Reinmar:
So far so good. Gulp tasks work nearly perfectly (there's a small issue which I extracted to a followup: https://github.com/ckeditor/ckeditor5-dev/issues/91). I still need to check how editor builds, though.

Reinmar:
The Webpack plugin works nearly fine (I'll report a followup) which is another great news :).

We don't have time to polish the code now, so I'm not checking it too deeply. Let's first focus on potential issues and missing features.

Since I was able to build a localized editor:

image

And the tool is really nice to use:

https://gist.github.com/Reinmar/c8f34b34390b3e2d8797468e09376651

I'm merging this PR :) Congrats!

PR #75, opened 2017-02-07
Rebuilt the release tools.

### LINKED ISSUE #19: Automate version bumping for ckeditor5-* dependencies
Moved from: https://github.com/ckeditor/ckeditor5-dev-env/issues/13.

### LINKED ISSUE #57: Release tasks should also publish on npm changed packages
Currently, they only release them to GitHub.

### LINKED ISSUE #61: dev-env: Changelog task should not break if there's no `CHANGELOG.md` file present
It breaks now. It should create if from scratch.

### LINKED ISSUE #64: The release task needs to carefully choose packages to release
Right now it's gonna release every single package (dependency of the main repo). It needs to be way smarter and release packages which:

1. Had any changes (even internal ones – after all, it will be just a patch release). The idea is that we'd generally want to have  latest releases on npm, even if only things like docs have changed. Otherwise, we'd risk having e.g. an outdated README on npm for months (if not years – see e.g. how few commits to the code were made in https://github.com/ckeditor/ckeditor-dev/commits/master/plugins/basicstyles/plugin.js). Also, we may mistakenly miss documenting some change, so really – it's just safer to release it any commit was made.
2. If no commit was made, release a package if any of its dependencies (excluding devDependencies!) will be released (and needs to be updated in `package.json`). The idea, in this case, is that if something got fixed in a dependency, it also might affect the dependent package. It's also quite clear I think that it always needs to be a patch release because new features added to dependency do not affect the dependent package. However, a minor or major release of a dependency might bring patches so a depedent package needs to be bumped up too.

The last point doesn't seem very hard to implement, but it is... because currently the changelog task creates version entries in packages which haven't changed. So it must work a bit differently. It must not create new version entry for a package with no commits and this entry must be created by the release tool if it release the package anyway.

### LINKED ISSUE #65: dev-env: The release task should implement 3 steps
Steps:

1. Make all necessary checks for all packages at once. If any of these packages is in a wrong state abort.
2. Print what the task will do (which packages will be released to which versions) and wait for confirmation.
3. Do all the things. Nothing should break now.

pomek:
```
Feature: Introduced a brand new release tools with a new set of requirements. See #64.

* Added interactive mode for generating the changelog. A user can provide a new version or skip the process for given package.
* Release task - rebuilt module for collecting dependencies to release. 
* Release task - a new version to release will be read from CHANGELOG file.
* Changelog task - will not break if changelog does not exist. 
* Used `semver` package for bumping the version (instead of a custom module).

Closes #19.
Closes #57.
Closes #61.
Closes #64.
Closes #65.

BREAKING CHANGES: Removed the `getoptions` module.
BREAKING CHANGES: CLI parameters are not supported anymore.
BREAKING CHANGES: Simplified options required by `tasks.releaseDependencies()`.
```


pomek:
Screenshots which show how the tool looks now:

image

image

image

image

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
```diff
@@ -0,0 +1,140 @@
...
+		} );
+	},
+
+	/**
+	 * Asks a user for providing the GitHub token.
+	 *
+	 * @returns {Promise}
+	 */
+	provideToken() {
+		return new Promise( ( resolve, reject ) => {
```
Reinmar: No need for this promise.

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
```diff
@@ -0,0 +1,140 @@
...
+	/**
+	 * Asks a user for providing the new version.
+	 *
+	 * @param {String} packageName
+	 * @param {String} packageVersion
+	 * @param {String} releaseType
+	 * @returns {Promise}
+	 */
+	provideVersion( packageName, packageVersion, releaseType ) {
+		return new Promise( ( resolve, reject ) => {
```
Reinmar: No need for this pormise.

packages/ckeditor5-dev-env/lib/release-tools/tasks/createrelease.js
```diff
@@ -8,80 +8,120 @@
...
+			json.version = version;
+
+			return json;
 		} );
+
+		log.info( `Committing "package.json"...` );
+		exec( 'git add package.json' );
+		exec( `git commit --message="Release: v${ version }."` );
+
+		log.info( 'Creating tag...' );
```
Reinmar: Creating a tag...

or

Tagging...

packages/ckeditor5-dev-env/lib/release-tools/tasks/createrelease.js
```diff
@@ -8,80 +8,120 @@
...
+		exec( `git tag v${ version }` );
+		exec( `git push origin master v${ version }` );
+
+		if ( !options.skipNpm ) {
+			log.info( 'Publishing on NPM...' );
+			exec( 'npm publish' );
+		}
+
+		if ( !options.skipGithub ) {
+			log.info( 'Creating GitHub release...' );
```
Reinmar: Again, missing "a".

packages/ckeditor5-dev-env/lib/release-tools/tasks/createrelease.js
```diff
@@ -8,80 +8,120 @@
...
+			return createGithubRelease( options.token, releaseOptions )
+				.then( resolve )
+				.catch( reject );
+		}
+
+		resolve();
+	} );
+
+	return promise.then( () => {
+		log.info( `Release "v${ version }" has been created and published.\n` );
```
Reinmar: Make this green.

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
pomek: https://github.com/ckeditor/ckeditor5-dev-env/blob/master/lib/utils/inquiries.js#L15-L29 - a few weeks ago it has a promise. So what is wrong?

packages/ckeditor5-dev-env/lib/release-tools/tasks/createrelease.js
```diff
@@ -8,80 +8,120 @@
...
-	}
 
-	return getNewReleaseType()
-		.then( ( response ) => {
-			const bumpVersionCommand = `npm version ${ response.releaseType } --no-git-tag-version --force`;
-			version = tools.shExec( bumpVersionCommand, shExecParams ).trim();
+		if ( exec( 'git diff --name-only package.json' ).trim().length ) {
+			log.info( 'Updating dependencies...' );
+			exec( 'git add package.json' );
+			exec( 'git commit -m "Internal: Update dependencies."' );
```
Reinmar: Update*d* dependencies

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
Reinmar: First of all – it used a promise, because a callback was passed to that function. Either the API changed or someone missed it in the past that you can get a promise from the inquirer.

> So what is wrong?

There's unneeded complexity in the code. That's wrong.

pomek: Yea, I noticed. You're right. Sorry.

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelog.js
```diff
@@ -7,59 +7,86 @@
...
  */
-module.exports = function generateChangelog() {
+module.exports = function generateChangelog( newVersion = null ) {
 	const log = logger();
 
-	log.info( `Generating changelog for: ${ process.cwd() }` );
+	return new Promise( ( resolve ) => {
+		const packageJson = getPackageJson();
+
+		log.info( `Generating changelog for "${ packageJson.name }".` );
```
Reinmar: "the changelog" or "changelog entries"

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelog.js
```diff
@@ -7,59 +7,86 @@
...
+					const newReleaseType = ( hasCommitsFromLastRelease() ) ? response.releaseType : null;
 
-		return new Promise( ( resolve ) => {
-			// conventionalChangelog based on version in `package.json`.
-			conventionalChangelog( {}, null, { merges: undefined, firstParent: true }, parserOpts, writerOpts )
-				.pipe( saveChangelogPipe() );
+					return cli.provideVersion( packageJson.name, packageJson.version, newReleaseType );
+				} );
+		} else {
+			promise = promise.then( () => Promise.resolve( newVersion ) );
```
Reinmar: No need to use `Promise.resolve()` AFAICS.

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelog.js
```diff
@@ -7,59 +7,86 @@
...
+		}
 
-			function saveChangelogPipe() {
-				return stream.noop( ( changes ) => {
-					let currentChangelog = utils.getChangelog();
+		return promise
+			.then( ( version ) => {
+				if ( version === 'skip' ) {
```
Reinmar: Just checking – is the `version` trimmed at some point? What if I type something incorrect?

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
```diff
@@ -0,0 +1,140 @@
...
+	 * Asks a user for a confirmation for creating the releases.
+	 *
+	 * @param {Map} packages Packages to release.
+	 * @returns {Promise}
+	 */
+	confirmRelease( packages ) {
+		let message = 'Packages to release:\n';
+
+		for ( const [ packageName, packageDetails ] of packages ) {
+			message += `  - "${ packageName }": v${ packageDetails.version }\n`;
```
Reinmar: You need to remember about consistency. Aren't log entries listed as ` * <TEXT>`? Use the same indentation and list bullet all across the code.

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
```diff
@@ -0,0 +1,140 @@
...
+		}
+
+		message += 'Continue?';
+
+		return new Promise( ( resolve, reject ) => {
```
Reinmar: No need for this `Promise()`.

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelog.js
pomek: If you type anything other than `skip`, provided input will be checking by the `semver` package. Most probably it will be rejected because your input will be marked as a invalid version.

Reinmar: So that would be fine for now. Just let's make sure the value is trimmed.

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
```diff
@@ -0,0 +1,140 @@
...
+	 * @param {String} packageVersion
+	 * @param {String} releaseType
+	 * @returns {Promise}
+	 */
+	provideVersion( packageName, packageVersion, releaseType ) {
+		return new Promise( ( resolve, reject ) => {
+			const versionQuestion = {
+				type: 'input',
+				name: 'version',
+				default: ( releaseType ) ? semver.inc( packageVersion, releaseType ) : 'skip',
```
Reinmar: No need for parentheses around `releaseType`.

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
```diff
@@ -0,0 +1,140 @@
...
+				message: `New version for "${ packageName }" (currently ${ packageVersion }, type new version or "skip")?`,
```
Reinmar: This should be better:

currently "x.x.x", type the new version or "skip"

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
```diff
@@ -0,0 +1,140 @@
...
+				validate( input ) {
+					if ( input === 'skip' ) {
+						return true;
+					}
+
+					// TODO: Check whether provided version is available.
+					return semver.valid( input ) ? true : 'Please provide a valid version.';
```
Reinmar: `return semver.valid( input ) || 'error msg';`

Reinmar: This repeats in at least one more place.

Reinmar: Actually, this is even fine:

```js
input.length === 40 ? true : 'Please provide a valid token.'
```

Because you'd need parentheses around the comparison if you used `||` AFAIR.

pomek: Where? :p

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
```diff
@@ -0,0 +1,140 @@
...
+	},
+
+	/**
+	 * Asks a user for selecting services where packages will be released.
+	 *
+	 * If the user choices a GitHub, required token also has to be provided.
+	 *
+	 * @returns {Promise}
+	 */
+	configureReleaseOptions() {
```
Reinmar: This isn't a necessary step. After all, there were supposed to be the `--skip-npm` and `--skip-github` option. Let's not overcomplicate the tool unnecessarily.

packages/ckeditor5-dev-env/lib/release-tools/utils/packagestorelease.js
```diff
@@ -0,0 +1,79 @@
+/**
```
Reinmar: It took me a minute to decode this name :D I thought it's "packages to release" or "package store release" :D

packages/ckeditor5-dev-env/lib/release-tools/utils/packagestorelease.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Returns a list of packages to release.
+ *
+ * @param {Object} options
+ * @params {String} options.cwd Current work directory.
+ * @params {String} options.packages A relative path to the packages.
+ * @returns {Promise}
+ */
+module.exports = function packagesToRelease( options ) {
```
Reinmar: You need a verb in a function name. `getPackagesToRelease()`.

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
pomek: By introducing the `--skip-npm` and `--skip-github` this tool is more complicated because you need to remember about these options. 

In current solution, user can "configure" the release process and does not have to remember the all options.

packages/ckeditor5-dev-env/lib/release-tools/utils/packagestorelease.js
Reinmar: Remember about updating the file name.

packages/ckeditor5-dev-env/lib/release-tools/utils/versions.js
```diff
@@ -0,0 +1,50 @@
...
+
+		return ( matches ) ? matches[ 1 ] : null;
+	},
+
+	/**
+	 * Returns a name of the last created tag.
+	 *
+	 * @returns {String|null}
+	 */
+	getLastFromTag() {
```
Reinmar: This name makes no sense. If you have a problem with naming a function, use the docs as an inspiration.

pomek: It should be `getLastFromGit()`. My mistake.

packages/ckeditor5-dev-env/lib/release-tools/utils/versions.js
```diff
@@ -0,0 +1,50 @@
...
+
+const versions = {
+	/**
+	 * Returns a last created version in changelog file.
+	 *
+	 * @returns {String|null}
+	 */
+	getLastFromChangelog() {
+		const changelog = changelogUtils.getChangelog();
+		// TODO: Support for versions: alpha/beta/rc.
```
Reinmar: Please resolve this todo.

The safest regexp should be something like this:

```js
/\n## \[([^]]+)\]/
```

packages/ckeditor5-dev-env/lib/release-tools/utils/versions.js
```diff
@@ -0,0 +1,50 @@
...
+		const regexp = /## \[?(\d+\.\d+\.\d+)\]?/;
+
+		const matches = changelog.match( regexp );
+
+		return ( matches ) ? matches[ 1 ] : null;
```
Reinmar: Unnecessary parentheses.

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
Reinmar: Fair point. But I don't like to be asked for things which don't matter. You can leave it with the prompt, but then make sure that you haven't left some support for those CLI params. I saw `skipNpm` options around the code, but I can't tell whether they are leftovers or part of this prompt support.

packages/ckeditor5-dev-env/lib/index.js
```diff
@@ -5,7 +5,13 @@
 
 'use strict';
 
+const { logger } = require( '@ckeditor/ckeditor5-dev-utils' );
 const executeOnDependencies = require( './release-tools/utils/executeondependencies' );
+const packagesToRelease = require( './release-tools/utils/packagestorelease' );
+const validator = require( './release-tools/utils/releasevalidator' );
+const cli = require( './release-tools/utils/cli' );
+
+const BREAK_RELEASE_MESSAGE = 'Creating release has been aborted.';
```
Reinmar: Process of creating a release is also called "releasing" ;P

"Releasing has been aborted due to errors." would be better.

But what would be even better is if the error which is thrown had a proper message – i.e. containing a reason.

Instead of logging and throwing just throw and log the error in a `catch()`.

packages/ckeditor5-dev-env/lib/index.js
```diff
@@ -26,47 +32,114 @@ const tasks = {
...
 
-		const functionToExecute = ( repositoryName, repositoryPath ) => {
+		const errors = [];
+
+		return packagesToRelease( execOptions )
+			.then( ( dependencies ) => {
+				options.dependencies = dependencies;
+
+				if ( dependencies.size === 0 ) {
+					log.warning( 'Not found packages to release.' );
```
Reinmar: This error is so unclear that I had to analyse the code to understand why are you logging it.

When writing error messages, or any messages at all, think about the user, not about the implementation. The user does not know what "finding packages to release" means. Think about a more user oriented message. Perhaps something like:

None of the packages contains any changes since its last release. Aborting.

packages/ckeditor5-dev-env/lib/index.js
```diff
@@ -26,47 +32,114 @@ const tasks = {
...
 	releaseDependencies( options ) {
+		const log = logger();
+
 		const execOptions = {
 			cwd: options.cwd,
 			packages: options.packages
 		};
 
-		const functionToExecute = ( repositoryName, repositoryPath ) => {
+		const errors = [];
```
Reinmar: It took me a while to understand how are errors pushed to this array, so please leave some comment about that.

packages/ckeditor5-dev-env/lib/index.js
```diff
@@ -26,47 +32,114 @@ const tasks = {
...
+			.then( ( isConfirmed ) => {
+				if ( !isConfirmed ) {
+					throw new Error( BREAK_RELEASE_MESSAGE );
+				}
+
+				return executeOnDependencies( execOptions, validatePackages );
+			} )
+			.then( () => {
+				if ( errors.length ) {
+					log.error( 'The errors occur during release process.' );
```
Reinmar: Aborting due to errors:

packages/ckeditor5-dev-env/lib/index.js
```diff
@@ -26,47 +32,114 @@ const tasks = {
...
+				return Promise.resolve();
+			}
+
 			process.chdir( repositoryPath );
 
 			return tasks.createRelease( {
 				token: options.token,
-				dependencies: options.dependencies
+				dependencies: options.dependencies,
+				skipGithub: options.skipGithub,
```
Reinmar: Where do these two options come from?

pomek: These options come from object returned by `cli.configureReleaseOptions()`.

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
pomek: > I saw skipNpm options around the code, but I can't tell whether they are leftovers or part of this prompt support.

This prompt sets proper values for `skipNpm` and `skipGithub`. If `skipGithub == false`, a user has to provide a `token` in order to authorize a request to GH API.

PR #77, opened 2017-02-09
dev-tests: Added markdown and html file watching in manual tests

### LINKED ISSUE #52: Manual tests not watching .html and .md files
If I remember correctly this worked before - one could change .html or .md file and watcher updated manual tests.

ma2ciek:
Fixes #52.


packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -22,8 +22,10 @@ const writer = new commonmark.HtmlRenderer();
 const viewTemplate = fs.readFileSync( path.join( __dirname, 'template.html' ), 'utf-8' );
 
 module.exports = function compileManualTestHtmlFiles( buildDir, manualTestPattern ) {
-	const sourceMDFiles = globSync( path.join( manualTestPattern, '*.md' ) );
-	const sourceDirs = uniq( sourceMDFiles.map( file => path.dirname( file ) ) );
+	const sourceMdFiles = globSync( path.join( manualTestPattern, '*.md' ) );
```
Reinmar: MD not Md. Two letters abbrs should be uppercase.

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -32,34 +34,44 @@ module.exports = function compileManualTestHtmlFiles( buildDir, manualTestPatter
...
 	const log = logger();
+	const sourceMdFilePath = setExtension( sourceFilePathBase, 'md' );
+	const sourceHtmlFilePath = setExtension( sourceFilePathBase, 'html' );
+	const sourceJsFilePath = setExtension( sourceFilePathBase, 'js' );
+
+	const relativeHtmlFilePath = getRelativeFilePath( sourceHtmlFilePath );
+	const relativeJsFilePath = getRelativeFilePath( sourceJsFilePath );
 
-	const relativeFilePath = getRelativeFilePath( sourceFile );
+	console.log( relativeJsFilePath );
```
Reinmar: ?

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -32,34 +34,44 @@ module.exports = function compileManualTestHtmlFiles( buildDir, manualTestPatter
...
+	log.info( `Processing '${ gutil.colors.cyan( sourceFilePathBase ) }'...` );
 
 	// Compile test instruction (Markdown file).
-	const parsedMarkdownTree = reader.parse( fs.readFileSync( sourceFile, 'utf-8' ) );
+	const parsedMarkdownTree = reader.parse( fs.readFileSync( sourceMdFilePath, 'utf-8' ) );
 	const manualTestInstructions = '<div class="manual-test-sidebar">' + writer.render( parsedMarkdownTree ) + '</div>';
 
 	// Load test view (HTML file).
-	const htmlView = fs.readFileSync( `${ changeExtension( sourceFile, 'html' ) }`, 'utf-8' );
+	const htmlView = fs.readFileSync( `${ sourceHtmlFilePath }`, 'utf-8' );
```
Reinmar: It doesn't make sense to use a template string here.

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -80,8 +92,41 @@ function uniq( arr ) {
...
+		watchFile( filePath, onChange );
+	}
+}
+
+function watchFile( filePath, onChange ) {
+	const debouncedOnChange = debounce( () => onChange( filePath ), 500 );
+	fs.watch( filePath, debouncedOnChange );
+}
+
+function debounce( callback, delay ) {
```
Reinmar: Don't implement such methods yourself.

Either use a watcher which supports this or use lodash's deboune.

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -32,34 +34,44 @@ module.exports = function compileManualTestHtmlFiles( buildDir, manualTestPatter
 	sourceDirs.forEach( sourceDir => copyStaticFiles( buildDir, sourceDir ) );
 
 	// Generate real HTML files out of the MD + HTML files of each test.
-	sourceMDFiles.forEach( sourceFile => compileTestHtmlFile( buildDir, sourceFile ) );
+	sourceFilePathBases.forEach( sourceFilePathBase => compileTestHtmlFile( buildDir, sourceFilePathBase ) );
+
+	// Watch file and compile on change.
+	watchFiles( sourceMdFiles, ( mdFile ) => compileTestHtmlFile( buildDir, getFilePathWithoutExtension( mdFile ) ) );
+	watchFiles( sourceHtmlFiles, ( mdFile ) => compileTestHtmlFile( buildDir, getFilePathWithoutExtension( mdFile ) ) );
```
Reinmar: The name of the param is wrong. It's not an `mdFile`.

Reinmar:
Does this watcher handle additional test assets (other files, besides HTML and JS)?

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -80,8 +92,41 @@ function uniq( arr ) {
...
+	return path.join( dir, name );
+}
+
+function watchFiles( filePaths, onChange ) {
+	for ( const filePath of filePaths ) {
+		watchFile( filePath, onChange );
+	}
+}
+
+function watchFile( filePath, onChange ) {
```
Reinmar: This function is unnecessary. Don't split code into too many functions if not needed. I know this seems to be right to split code, but there are evidence that this reduces readability.

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -32,34 +34,44 @@ module.exports = function compileManualTestHtmlFiles( buildDir, manualTestPatter
...
 
 	// Load test view (HTML file).
-	const htmlView = fs.readFileSync( `${ changeExtension( sourceFile, 'html' ) }`, 'utf-8' );
+	const htmlView = fs.readFileSync( `${ sourceHtmlFilePath }`, 'utf-8' );
 
 	// Attach script file to the view.
 	const scriptTag =
 		'<body class="manual-test-container">' +
-			`<script src="./${ changeExtension( path.basename( sourceFile ), 'js' ) }"></script>` +
+			`<script src="/${ relativeJsFilePath }"></script>` +
```
Reinmar: Why an absolute path suddenly?

Reinmar:
First part of the review – let's work on the code quality. And please verify that the watcher also handles test assets (I haven't checked that). If it doesn't... then perhaps we can live with that for now, but please report a ticket for that (we can have a bug, but it should be reported).

Also, I was surprised that you used the `fs.watch()` function. If `fs-extra` doesn't improve the native `watch()` then I'd recommend using some specialised watcher package. I know there are many OS-specific issues with the default watcher.

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
ma2ciek: Hmm, maybe it's better to name the variable `absoluteToSeverRootJSFilePath`?
The path is relative to the ckeditor5 directory, but it's absolute in context of the server.

Reinmar: So it's just `absoluteFilePath` ;)

ma2ciek:
> Also, I was surprised that you used the `fs.watch()` function. If `fs-extra` doesn't improve the native watch() then I'd recommend using some specialised watcher package. I know there are many OS-specific issues with the default watcher.

I've done very similar to the previous implementation, didn't know the issue with `fs.watch()`. I've found this one: https://github.com/paulmillr/chokidar and it looks promising, so I'll add this package.
Actually, on the `fs-extra` page I found the following information: 
> File / Directory Watching
If you want to watch for changes to files or directories, then you should use chokidar.


> First part of the review – let's work on the code quality.

I'm sorry for that. Now I see big mistakes which I simply overlooked.

> And please verify that the watcher also handles test assets

Which assets do you mean? The static files?

pomek:
> I've done very similar to the previous implementation, didn't know the issue with fs.watch().

I didn't notice any issue with `fs.watch()` on three platforms - Windows, Linux and OSX. It works as I expected.

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
ma2ciek: And what with 3- and 4- letter long? is there a strict rule about this that we're using?

pomek:
https://github.com/paulmillr/chokidar/blob/master/lib/nodefs-handler.js#L37 :D

Reinmar:
> Which assets do you mean? The static files?

Yes, e.g. a test may use an png image (see e.g. ckeditor5-image's tests).

Reinmar:
> https://github.com/paulmillr/chokidar/blob/master/lib/nodefs-handler.js#L37 :D

Of course, at the end it uses `fs.watch()`. But what all the other code in this file does?

pomek:
I didn't analyze the code, so I don't know. 

I'm just saying that `fs.watch()` worked in the previous version properly.

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
Reinmar: https://github.com/ckeditor/ckeditor5-design/wiki/Code-Style-Naming-Guidelines#acronyms-and-proper-names

> However, two letter acronyms and proper names (if originally written uppercase) should be uppercase. So e.g. getUI (not getUi).

Longer acronyms should be `Xxxx`.

ma2ciek: So one line below there should be `mdFile` or `MDFile`? Because according to your quote it should be written using the second form, which doesn't look good for me.

ma2ciek:
> Yes, e.g. a test may use an png image (see e.g. ckeditor5-image's tests).

For now script watches only html and markdown files. It's a little bit tricky getting asset watching to work. Because with assets there's need to watch the entire directories, not the files itself. I'll create a ticket for that.

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -23,65 +25,88 @@ const viewTemplate = fs.readFileSync( path.join( __dirname, 'template.html' ), '
...
 
 	// Load test view (HTML file).
-	const htmlView = fs.readFileSync( `${ changeExtension( sourceFile, 'html' ) }`, 'utf-8' );
+	const htmlView = fs.readFileSync( sourceHtmlFilePath, 'utf-8' );
 
 	// Attach script file to the view.
 	const scriptTag =
 		'<body class="manual-test-container">' +
-			`<script src="./${ changeExtension( path.basename( sourceFile ), 'js' ) }"></script>` +
+			`<script src="/${ absoluteJsFilePath }"></script>` +
```
Reinmar: The same issue as with MD.

pomek:
@ma2ciek, could you write some tests for the new functionality?

ma2ciek:
> @ma2ciek, could you write some tests for the new functionality?

@pomek, @Reinmar fixed.

pomek:
It looks good but I still have a feeling that there is no test for a ticket which is resolving:  "Manual tests not watching .html and .md files".

In other words - we need to have a test which confirms that the test files will be compiled after changes.

packages/ckeditor5-dev-tests/tests/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -0,0 +1,112 @@
...
+		const outputFileSyncStub = sandbox.stub( fs, 'outputFileSync' );
+		const copySyncStub = sandbox.stub( fs, 'copySync' );
+		const chokidarWatchStub = sandbox.stub( chokidar, 'watch', () => ( { on: () => {} } ) );
+
+		mockery.registerMock( '../getrelativefilepath', getRelativeFilePathSpy );
+		mockery.registerMock( '../glob', globSyncSpy );
+		mockery.registerMock( 'dom-combiner', ( ...args ) => args.join( '\n' ) );
+		mockery.registerMock( '@ckeditor/ckeditor5-dev-utils', {
+			logger: () => ( {
+				info() { }
```
Reinmar: should be `{}`

packages/ckeditor5-dev-tests/tests/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -0,0 +1,112 @@
...
+		} );
+
+		const compileHtmlFiles = require( '../../../lib/utils/manual-tests/compilehtmlfiles' );
+		compileHtmlFiles( 'buildDir', 'manualTestPattern' );
+
+		sinon.assert.calledWithExactly( readFileSyncSpy, path.join( fakeDirname, 'template.html' ), 'utf-8' );
+		sinon.assert.calledWithExactly( readFileSyncSpy, path.join( 'path', 'to', 'file.md' ), 'utf-8' );
+		sinon.assert.calledWithExactly( readFileSyncSpy, path.join( 'path', 'to', 'file.html' ), 'utf-8' );
+		sinon.assert.calledWithExactly( globSyncSpy, path.join( 'manualTestPattern', '*.md' ) );
+		sinon.assert.calledWithExactly( mdParserSpy, '##Markdown header' );
```
Reinmar: Missing space after `##`.

Reinmar:
Yeah, I agree. A two simple tests are missing:

* one which checks that chokidar was called with proper arguments
* and another one that executing chokidar's callbalback triggers compilation of that specific file

BTW, the added test seems to be too complicated. It's trying to check too many things. I know that writing tests for code which interacts with IO isn't simple, so there are two options:

* to not go too crazy with what these tests do – the less they try to check, the less stubbing must be done,
* to create fixtures with pseudo-projects and run those tests on them.

ma2ciek:
I have to agree that this test is a little bit too complicated. I have some doubts whether to test some inner functions which I could make protected or to test whole file. Testing whole file gives more flexibility in rewriting some inner parts, because it's more a functional testing. 

Fixtures seams great to me. But do we have some guides or discussions how to write them? I saw @pomek one fixure in release tools. But these tests are no longer unit tests, so it should be described somewhere how much it's good to stub/mock and how much of node libraries that may for example parse markdown files can be trusted and not mocked. I'm still a little bit confused how to write good tests...

> one which checks that chokidar was called with proper arguments

https://github.com/ckeditor/ckeditor5-dev/pull/77/files#diff-5a21f688c7b7a80b04248ea9fe9803f4R108

> to not go too crazy with what these tests do – the less they try to check, the less stubbing must be done

It's impossible to stub less and still get output file content.

pomek:
In the release tool, I also created an integration test for one of the util, because I couldn't create a unit test which checks whether the tool works as we expected - https://github.com/ckeditor/ckeditor5-dev/pull/75/files#diff-896959e1de1b4e366d614cad7bb4d66cR16.

In the same PR, I removed all some of the stubs because weren't required anymore :D

Reinmar:
> Testing whole file gives more flexibility in rewriting some inner parts, because it's more a functional testing.

That should be the goal. It's not a strict rule, but if you see that you start stubbing too many things it means that perhaps a different approach should be used.

> Fixtures seems great to me. But do we have some guides or discussions how to write them?

No, we never discussed this. It's really a per-case thing. Sometimes it worth to stub more instead of spending a lot of time thinking how to work with fixtures and creating all the fixtures and cleaning them and so on... Sometimes things just get too complicated with that simple approach. 

In this case, I don't think that we should spend a lot more time on this now. If you see a room for quick improvements, then let's do them. If not, then let's close this ticket. In case of ckeditor5-dev we don't have to be that perfect for now – it was rather a comment for the future :).

ma2ciek:
Ok. That makes sense. I removed unnecessary assertions for now.

PR #94, opened 2017-03-03
Add language code mapping and old translations removing, prevent empty po files saving

### LINKED ISSUE #91: Configure Transifex language mapping
Currently we can find such files to be downloaded from Transifex:

```
[16:21:44] Saved ne_NP.po for ckeditor5-ui package
[16:21:44] Saved no.po for ckeditor5-ui package
[16:21:44] Saved nb.po for ckeditor5-ui package
[16:21:44] Saved oc.po for ckeditor5-ui package
[16:21:44] Saved fa.po for ckeditor5-ui package
[16:21:44] Saved pl.po for ckeditor5-ui package
[16:21:44] Saved pt.po for ckeditor5-ui package
[16:21:44] Saved pt_BR.po for ckeditor5-ui package
[16:21:44] Saved ro.po for ckeditor5-ui package
[16:21:44] Saved ru.po for ckeditor5-ui package
[16:21:44] Saved ru@petr1708.po for ckeditor5-ui package
[16:21:44] Saved sr.po for ckeditor5-ui package
[16:21:44] Saved sr@latin.po for ckeditor5-ui package
[16:21:44] Saved si_LK.po for ckeditor5-ui package
```

Unfortunately, these are not correct names in ISO639-1. According to @wwalc if we'd use a CLI version of Tx tool, we could provide such a mapping there as an option:

```
lang_map = en_AU:en-au,de_CH:de-ch,en_CA:en-ca,en_GB:en-gb,es_CO:es-co,fr_CA:fr-ca,ne_NP:ne,pt_BR:pt-br,ru@petr1708i:ru-petr1708,si_LK:si,sr@latin:sr-latn,ta_LK:ta,zh_CN:zh-cn,zh_HK:zh-hk,zh_TW:zh
```

Let's see how to do that if we're using the REST API. Also, let's check if this mapping is correct.

ma2ciek:
Fixes #91, #95.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/translations/download.js
+++ b/packages/ckeditor5-dev-env/lib/translations/download.js
@@ -49,7 +49,7 @@ function downlaodAndReplaceTranslations( loginConfig, packageNames ) {
 
 function removeOldTranslationForPackage( packageName ) {
 	const del = require( 'del' );
-	const glob = path.join( process.cwd(), 'packages', packageName, 'lang', 'translations', '**', '**' );
+	const glob = path.join( process.cwd(), 'packages', packageName, 'lang', 'translations', '**' );
 
 	return del( glob );
 }
--- a/packages/ckeditor5-dev-env/lib/translations/download.js
+++ b/packages/ckeditor5-dev-env/lib/translations/download.js
@@ -9,6 +9,7 @@ const fs = require( 'fs-extra' );
 const path = require( 'path' );
 const transifexService = require( './transifex-service' );
 const logger = require( '@ckeditor/ckeditor5-dev-utils' ).logger();
+const translationUtils = require( '@ckeditor/ckeditor5-dev-utils' ).translations;
 
 /**
  * Downloads translations from the Transifex for each package and language.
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-dev-env/lib/translations/download.js
```diff
@@ -46,8 +47,17 @@ function downlaodAndSaveTranslations( loginConfig, packageNames ) {
 	);
 }
 
+function removeOldTranslationForPackage( packageName ) {
+	const del = require( 'del' );
+	const glob = path.join( process.cwd(), 'packages', packageName, 'lang', 'translations', '**', '**' );
```
Reinmar: Shouldn't this end with `**, *`?

Reinmar:
Why is this PR changing the way how translations are saved? I mean – the ticket is about mapping. But here I can see some deleting and so on. Why such changes?

ma2ciek:
> Why is this PR changing the way how translations are saved? I mean – the ticket is about mapping. But here I can see some deleting and so on. Why such changes?

I added deletions to prevents creating new files with the existing one. without this change you will have old 'en_AU' and new 'en-au' file.

It will change the title to mention it.

Reinmar:
> I added deletions to prevents creating new files with the existing one. without this change you will have old 'en_AU' and new 'en-au' file.

We don't have any files yet, so this is not a problem for now. But ok, since you already implemented that logic there's no point in blocking it :) May be helpful in the future – e.g. if we change the mapping.

ma2ciek:
Now this PR fixes also #95.

packages/ckeditor5-dev-env/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "devDependencies": {
     "chai": "^3.5.0",
+    "del": "^2.2.2",
```
Reinmar: Well... devDeps again? ;)

PR #102, opened 2017-03-07
T/97: Breaking changes and new feature were ignored

### LINKED ISSUE #97: Breaking changes and new feature were ignored
image

Something is wrong here. The version bump should be minor, not patch.

The changelog section is ok:

```
## [0.7.1](https://github.com/ckeditor/ckeditor5-heading/compare/v0.7.0...v0.7.1) (2017-03-06)

### Features

* Enabled configuration and localization of available headings (see `config.heading.options`). Closes [#33](https://github.com/ckeditor/ckeditor5/issues/33). ([de07a0c](https://github.com/ckeditor/ckeditor5-heading/commit/de07a0c))

### Other changes

* Uploaded translations. ([50583b4](https://github.com/ckeditor/ckeditor5-heading/commit/50583b4))


### BREAKING CHANGES

* The `heading` command now accepts `id` option, not `formatId`.
```

pomek:
Fix: Recommended version was incorrect for merge commits. Closes #97.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js
@@ -100,6 +100,7 @@ function transformCommit( commit, displayLog = true ) {
 
 	const issues = [];
 
+	commit.rawType = commit.type;
 	commit.type = getCommitType( commit.type );
 
 	if ( commit.scope === '*' ) {
--- a/packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js
@@ -38,7 +38,6 @@ const typesOrder = {
 const packageJson = getPackageJson();
 const issuesUrl = ( typeof packageJson.bugs === 'object' ) ? packageJson.bugs.url : packageJson.bugs;
 const templatePath = path.join( __dirname, 'templates' );
-const log = logger();
 
 module.exports = {
 	transform: transformCommit,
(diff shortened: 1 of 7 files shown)


Reinmar:
I checked that the PR seems to fix the issue, but I have no idea what was wrong and what it actually does. The important part of every bug fix is explaining to others what was the problem. Otherwise the bug fix cannot be validated by others and others don't learn from that mistake.

I'm merging the PR, but please explain the PR in a comment.

pomek:
https://github.com/ckeditor/ckeditor5-dev/issues/97#issuecomment-284648272 :>

If you still don't understand - let me know, will try to describe it with more details.

Reinmar:
I can only guess that in case of merge commit we need to pass to some function a stripped commit message. But I can't see that change in the code. I can see lines like https://github.com/ckeditor/ckeditor5-dev/pull/102/files#diff-2dae7bc2c7cb13d0ff39f553ce400e02R65 or https://github.com/ckeditor/ckeditor5-dev/pull/102/files#diff-2dae7bc2c7cb13d0ff39f553ce400e02R103 or https://github.com/ckeditor/ckeditor5-dev/pull/102/files#diff-b291f4500070156fbf0bfe6fbf39fba4R45 which seem to be irrelevant.

Which means that either the PR contained irrelevant changes or that the fix was harder.

pomek:
https://github.com/ckeditor/ckeditor5-dev/pull/102/files#diff-b291f4500070156fbf0bfe6fbf39fba4R45 - it is required because in the issue you wrote:

> Something is wrong here. The version bump should be minor, not patch.

So I introduced a mechanism which detects whether current version is below the `1.0.0`. If the version is below and next release should be a `major`, the version will be bumped as `minor`, not `major`.

pomek:
Ofc for versions above the `1.0.0`, it will work according to SemVer.

Let's summarize:

- for a version below `1.0.0` (e.g.: `0.7.0`)
  * next `patch` => `0.7.1`,
  * next `minor` => `0.8.0`,
  * next `major` => `0.8.0`, ([according to semver](http://semver.org/#spec-item-4), API can change but the package is still non-public)
- for a version above or equal to `1.0.0` (e.g. `1.1.1`): 
  * next `patch` => `1.1.2`,
  * next `minor` => `1.2.0`,
  * next `major` => `2.0.0`.

Reinmar:
So, do you mean that the `semver` package doesn't handle this? This explains the PR a bit.

However, it doesn't explain the rest of the changes. If the root of the change is in `provideVersion()` then what all the other changes do? Why were they made?

pomek:
I don't understand this discussion. I fixed a bug which you reported. I changed files which behaved invalid. Everything works ok and I feel like you have a grudge because of that.

If the changes weren't required, most probably I haven't introduced it.

PR #106, opened 2017-03-07
T/88: Disable background throttling in Chrome

### LINKED ISSUE #88: Disable background throttling in Chrome
It turns out that there's a solution to that problematic behaviour if Chrome getting super slow when in bg: https://bugs.chromium.org/p/chromium/issues/detail?id=485425#c21

Let's enable it in our tests.

pomek:
Internal: Disabled background timer throttling for automated tests. Closes #88.


Reinmar:
Did you test it? It doesn't work for me.

Reinmar:
OK, it works after the second time ;| Odd.

Reinmar:
BTW, this is a fix, not an internal change.

PR #107, opened 2017-03-07
T/88: Disable background throttling in Chrome (2)

### LINKED ISSUE #88: Disable background throttling in Chrome

pomek:
Internal: Added `CHROME_LOCAL` browser which uses `--disable-background-timer-throttling` option for local environments. Fixes #88.

`CHROME_LOCAL` is used instead of `Chrome`.

---

The previous merge introduced the option only for CI.


Reinmar:
As I mentioned in https://github.com/ckeditor/ckeditor5-dev/issues/88 it doesn't work. But it's a bug in MacOS, so it makes sense to close this PR anyway.

PR #108, opened 2017-03-07
T/79: Use deps hoisting to speed up installation time on CI 

### LINKED ISSUE #79: Use deps hoisting to speed up installation time on CI
Just like in ckeditor5, let's use hoisting (I mean Lerna's feature). This speeds up installation time from ~6 to 3:30min.

pomek:
Internal: Added `hoist` option to `lerna.json` in order to improve the speed of installing packages on CI. Closes #79.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,7 @@
   "packages": [
     "packages/*"
   ],
+  "hoist": true,
+  "nohoist": "guppy-pre-commit",
   "version": "independent"
 }


pomek:
Before:

image

After:

image

Reinmar:
OK, this is indeed useful, but it doesn't solve #79 which was about speeding up CIs for ckeditor5-* packages.

I'll merge this PR, but #79 will remain open.

PR #109, opened 2017-03-08
T/79: Use deps hoisting to speed up installation time on CI (2)

### LINKED ISSUE #79: Use deps hoisting to speed up installation time on CI

pomek:
Internal: Use Lerna `hoist` option in order to improve the speed of installing packages on CI for ckeditor5-* packages. Closes #79.

### DIFF
--- a/packages/ckeditor5-dev-tests/bin/create-lerna-json.js
+++ b/packages/ckeditor5-dev-tests/bin/create-lerna-json.js
@@ -20,8 +20,7 @@ let json = {
 		'.'
 	],
 	'version': '0.0.0',
-	'hoist': true,
-	'nohoist': 'guppy-pre-commit',
+	'hoist': true
 };
 
 fs.writeFileSync( lerna, JSON.stringify( json, null, 2 ) + '\n', 'utf-8' );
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-dev-tests/bin/create-lerna-json.js
```diff
@@ -19,7 +19,9 @@ let json = {
 		'packages/*',
 		'.'
 	],
-	'version': '0.0.0'
+	'version': '0.0.0',
+	'hoist': true,
+	'nohoist': 'guppy-pre-commit',
```
Reinmar: For CI this can be hoisted too.

pomek: Fixed.

Reinmar:
This wasn't an internal change – for this package (ckeditor5-dev-tests) this is a feature/fix. It's internal if we talk about this in context of ckeditor5, because it concerns how it runs the tests (which is the internal aspect of this package). But since ckeditor5-dev-tests's role is to run tests, then how it run the tests is it's feature/API/public matter.

PR #112, opened 2017-03-08
T/110: Introduce release tasks to this package

### LINKED ISSUE #110: Introduce release tasks to this package
As in the title.

pomek:
Internal: Introduced release tasks. Closes #110.


Reinmar:
I see a problem here... we release the packages independently, so we can't have a single changelog for them. Babel release all packages together so they are versioned in the same way, hence we can't do this like they do: https://github.com/babel/babel/blob/7.0/CHANGELOG.md

I think that we need to generate changelog for each package separately, so our changelog generator needs some improvements (to get commits only affecting a certain directory). Then, I'd propose to use Lerna's `publish` method which is more suitable to how this package is shaped – e.g. there must be only one release commit – like: https://github.com/ckeditor/ckeditor5-dev/commit/3c429e5c12b2ef239210966276e9bf79bed0c87e

pomek:
At this moment I don't understand for what we created this repository and merged all packages here.

Reinmar:
> At this moment I don't understand for what we created this repository and merged all packages here.

We merged them into one repository to make it simpler to work on them. It gives us more flexibility and in this particular case it didn't make sense to keep them separately.

But these are still separate packages, so they should be released independently. Hence, they need separate changelogs. Just like in ckeditor5.

pomek:
So, the changelog task must (somehow) know which commits are proper for given package. 

[`transformCommit()`](https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js#L58-L125) can filter commits. But we have two cases. Changelog may be generated for:

1. a repository which has only one package (ckeditor5-*),
2. a repository which has multi packages (ckeditor5-dev).

Can we distinguish which mode should be taken for parse "package"?

## Examples:

#### Generate changelog for `ckeditor5-core`

- check whether the package is a dev package -> `false`,
- parse commits using existing rules.

#### Generate changelog for `ckeditor5-dev-env`

- check whether package is a dev package -> `true`
- for each commit from the whole repository does:
  * find all changed files for commit (`git diff-tree --no-commit-id --name-only -r COMMIT_ID`)
  * check whether all the files are in package directory:
    - if `true` -> the commit is valid and will be displayed in a changelog,
    - if `false` -> skip the commit.
  * generate the changelog based on filtered out commits.

Reinmar:
> Can we distinguish which mode should be taken for a package based on its a name?

That's the wrong direction, so I'm not even reading further, ok? :P 

Make this configurable. The config option should come from gulpfile or other "caller".

pomek:
Required ticket - https://github.com/ckeditor/ckeditor5-dev/issues/121.

pomek:
After merging #154 this PR won't be valid anymore.

PR #117, opened 2017-03-09
T/104: Fixed invalid URLs to Github issues.

### LINKED ISSUE #104: Changelog contains invalid URLs to issues
During generating the changelog, references to issues are manually replaced to URLs to Github.

Everything works fine for a single package. The problem appears when changelog is generating for few packages.

Unfortunately, [`package.json`](https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js#L38) is read once. It means, different packages have the same `package.json`. 

Look on [changelog from `ckeditor5-core`](https://raw.githubusercontent.com/ckeditor/ckeditor5-core/master/CHANGELOG.md) - URLs issues are incorrect. They lead to the `ckeditor5` (the whole generating changelog process was called from the cke5 directory). URLs should lead to `ckeditor5-core` (as it is for commits).

pomek:
Fix: Changelog contained invalid URLs to Github issues. Closes #104.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js
@@ -133,8 +133,7 @@ function linkGithubIssues( value, issues = null ) {
 	const issuesUrl = ( typeof packageJson.bugs === 'object' ) ? packageJson.bugs.url : packageJson.bugs;
 
 	if ( !issuesUrl ) {
-		throw new Error( `File "package.json" for package "${ packageJson.name }" ` +
-			'does not contain a proper set "bugs" property.' );
+		throw new Error( `The package.json for "${ packageJson.name }" must contain the "bugs" property.` );
 	}
 
 	return value.replace( /#([0-9]+)/g, ( _, issueId ) => {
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js
```diff
@@ -129,6 +128,9 @@ function linkGithubUsers( value ) {
 }
 
 function linkGithubIssues( value, issues = null ) {
+	const packageJson = getPackageJson();
+	const issuesUrl = ( typeof packageJson.bugs === 'object' ) ? packageJson.bugs.url : packageJson.bugs;
```
Reinmar: It should somehow throw or log an error if one of the repositories doesn't have the `bugs` property set.

packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js
```diff
@@ -153,3 +155,15 @@ function getCommitType( commit ) {
 			throw new Error( `Given invalid type of commit ("${ commit }").` );
 	}
 }
+
+function getPackageJson() {
```
Reinmar: Why the caching?

pomek: I would like to avoid reading the same file n times.

Reinmar: This is a premature optimisation. The performance of this tool (which is absolutely not important because it's a dev tool) would not even be affected in any significant manner by removing this caching because the OS/disk will cache this file itself.

KISS.

pomek:
Done.

PR #120, opened 2017-03-13
T/105: Release tasks should ignore some packages

### LINKED ISSUE #105: Release tasks should ignore some packages
Soon, we'll have builds repositories available in `ckeditor5`, but builds needs to be released separately (in the same way as `ckeditor5` itself). So `gulp changelog:dependencies` and `gulp release:dependencies` must skip it. This can be configured as a glob pattern passed to these tasks (the option can be hardcoded in `gulpfile.js` for now, later me may think about CLI).

Also, I often have some unofficial packages installed in ckeditor5, so they should be skipped as well. I think that the easiest way to do so will be to simply filter the contents of `packages/` directory by what's inside `package.json`'s dependencies. If a package is not there, it shouldn't be touched. NOTE: It'd be good to list those skipped packages so the releaser knows that something is ignored!

pomek:
Feature: Release task will ignore specified packages during the release process. Closes #105.

By passing list of packages as `options.skipPackages` you can define which packages won't be released during the release process. This option works with `tasks.releaseDependencies()` and `tasks.generateChangelogForDependencies()` tasks.


packages/ckeditor5-dev-env/lib/release-tools/utils/getpackagestorelease.js
```diff
@@ -79,4 +83,17 @@ module.exports = function getPackagesToRelease( options ) {
...
+
+	function displaySkippedPackages( skippedPackages ) {
+		if ( skippedPackages && !skippedPackages.length ) {
+			return;
+		}
+
+		const { logger } = require( '@ckeditor/ckeditor5-dev-utils' );
+
+		let message = 'Packages listed below have been skipped:\n';
+		skippedPackages.forEach( ( packageName ) => message += `  * ${ packageName }\n` );
```
Reinmar: This would be better as a `map()` + `join()`. You wouldn't need to trim it afterwards.

Reinmar:
> This option works with tasks.releaseDependencies() task.

As I wrote in the ticket:

> So gulp changelog:dependencies and gulp release:dependencies must skip it.

Those tasks are used in tandem. Having one configurable without the other one doesn't make much sense.

pomek:
Fixed.

Reinmar:
The skipped packages are only listed by the release task, am I right? This isn't perfect, but I guess it can work this way.

pomek:
No, you aren't.

image

Both tasks print the skipped packages.

Reinmar:
OK, I didn't notice this:

image

PR #129, opened 2017-03-15
T127: The save-revision script could log a link to a commit that it's made

### LINKED ISSUE #127: The save-revision script could log a link to a commit that it's made
Right now it doesn't log anything so you don't know if it worked and you can't go straight to that commit. So something like this would be nice:

> Successfully saved the revision under https://github.com/ckeditor/ckeditor5/commit/fc516a

pomek:
Internal: Changed commit message for commits made by script for saving revisions. Closes #127.


Reinmar:
But it doesn't log anything still. You just changed the message.

Reinmar:
For the future – it's not an internal change. It affects how this tool works for others.

PR #137, opened 2017-03-22
T/135 Fix problem with manual tests containing periods

### LINKED ISSUE #135: [TESTS] Manual test files cannot contain a period in name
My files structure looks like:

```
- ... 
- tests
 `- manual
   |- ckeditor.es6.html
   |- ckeditor.es6.js 
   `- ckeditor.es6.md
- ...
```

When I'm calling the `gulp test:manual`, it shows me an error:

image

I don't know why it searches for `ckeditor.md` instead of `ckeditor.es6.md`.

ma2ciek:
Fix: Manual tests containing period in file name should not crash. Closes #135.

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
+++ b/packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
@@ -48,9 +48,9 @@ module.exports = function compileHtmlFiles( buildDir, manualTestPattern ) {
 
 function compileHtmlFile( buildDir, sourceFilePathBase, viewTemplate ) {
 	const log = logger();
-	const sourceMDFilePath = setExtension( sourceFilePathBase, 'md' );
-	const sourceHtmlFilePath = setExtension( sourceFilePathBase, 'html' );
-	const sourceJSFilePath = setExtension( sourceFilePathBase, 'js' );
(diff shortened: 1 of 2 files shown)


ma2ciek:
Misses test for now. But should work already.

Reinmar:
I'd consider whether the `setExtension()` function is needed at all if it's used in just one place and whether this patch shouldn't fix that place too.

ma2ciek:
```
 setExtension( mdFile, 'html' )
```

It's the only usage and here it'll work perfectly because the `setExtension` function removes the current extension and adds new extension. I can do it manually, but I'll need to parsing manually the path string. I can name it  `replaceExtension` to clarify the usage.

pomek:
I would like to see a test which shows that the whole mechanism will work when files contain a period in their names. 

The test will provide that the bug (#135) will not appear once again.

pomek:
To be more precise - I mean a file which has more than one a period in name, e.g. `ckeditor.test.js`.

ma2ciek:
So I have already written it ;) https://github.com/ckeditor/ckeditor5-dev/blob/133aea73a86aeeb77f1b232e7f86a5e878272924/packages/ckeditor5-dev-tests/tests/utils/manual-tests/compilehtmlfiles.js#L114

pomek:
Yea, I did not notice it. Sorry and thanks!

PR #138, opened 2017-03-24
T/51: Add --files support for test:manual because it's too slow

pomek:
Feature: Added ability to compile specified manual tests. Closes #51.

PR #140, opened 2017-03-27
T/135 Fix problem with manual tests containing dots in file names

ma2ciek:
Fixes https://github.com/ckeditor/ckeditor5-dev/issues/135.

The glob pattern `globSync( path.join( sourceDir, '**', '*.!(js|html|md)' )` matches e.g. `someDir/ckeditor.es6.md`. It's hard for me to figure out the correct glob pattern for this match so I added the regexp exclusion.


ma2ciek:
The Glob should be a superset of the RegExp. it would be much easier to figure out the correct pattern :)

pomek:
Proposed changes are okay but would be nice to have a test which checks whether the files will be ignored. 

Because existing tests have been rewritten in #138, I propose to wait for #138 and then write the test.

ma2ciek:
Yes, I agree. It's hard to test it because it's glob-dependent and unfortunately it could be tested well with fixtures only.

Reinmar:
#138 was merged. So we're waiting here for a test?

pomek:
Yes.

ma2ciek:
I added the missing test.

PR #144, opened 2017-03-28
T/121: Extract the `transformCommit` function from writer-options.js

### LINKED ISSUE #121: [ENV] Extract the `transformCommit` function from writer-options.js
In order to do #110, we need to extract the `transformCommit()` function from [`writer-options.js`](https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-env/lib/release-tools/changelog/writer-options.js) file.

It allows replacing rules which are used to generate the changelog using parameters.

At this moment, the function is hardcoded and we aren't able to introduce other rules, e.g. dev packages (`ckeditor5-dev-*`) needs other rules than editor's packages (`ckeditor5-*`).

pomek:
Internal: Moved `transformCommit` function from `writer-options` to new file. Closes #121.

It allows replacing rules which are used to generate the changelog using parameters.


pomek:
In this PR I corrected a little the files structure.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommit.js
```diff
@@ -0,0 +1,96 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+const getPackageJson = require( './getpackagejson' );
+
+const transformCommit = {
```
ma2ciek: 1. The file name `transformCommit` is suggesting, that the default export should be a function, which is not. 
2. This file name should contains `-` between words according to our style guide :D

Maybe `transform-commit-utils`/ `transform-commit-helpers`?

pomek: :+1: for `transform-commit-utils`

packages/ckeditor5-dev-env/lib/release-tools/utils/getnewreleasetype.js
```diff
@@ -14,7 +15,7 @@ const { commitTypes, transform } = require( '../changelog/writer-options' );
  */
 module.exports = function getNewReleaseType() {
 	const conventionalRecommendedBump = require( 'conventional-recommended-bump' );
-	const parserOpts = require( '../changelog/parser-options' );
+	const parserOpts = require( './parseroptions' );
```
ma2ciek: Maybe `parserOptions`? You're saving 4 chars ;)
And now I don't know whether the file name is correct :D It's a plain object, so probably it shouldn't contain '-'.

ma2ciek: Ow. Now I see, that there is such a convention with `opts` across the committed files.

packages/ckeditor5-dev-env/lib/release-tools/utils/getnewreleasetype.js
```diff
@@ -5,7 +5,8 @@
 
 'use strict';
 
-const { commitTypes, transform } = require( '../changelog/writer-options' );
+const { availableTypes } = require( './transformcommit' );
```
ma2ciek: The name `availableTypes` becomes too generic when it's used without the `transformCommit` context.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
```diff
@@ -0,0 +1,141 @@
...
+	for ( const note of commit.notes ) {
+		if ( note.title === 'BREAKING CHANGE' ) {
+			note.title = 'BREAKING CHANGES';
+		}
+		note.text = transformCommit.linkGithubIssues( transformCommit.linkGithubUsers( note.text ) );
+	}
+
+	// Removes references that already appear in the subject.
+	commit.references = commit.references.filter( ( reference ) => {
+		return issues.indexOf( reference.issue ) === -1;
```
ma2ciek: It can be replaced with `includes()`

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
```diff
@@ -0,0 +1,141 @@
...
+	} );
+
+	return commit;
+};
+
+/**
+ * @typedef {Object} Commit
+ *
+ * @property {String} [type] Type of the commit.
```
ma2ciek: Should every property in type Commit be optional?

pomek: Yes, if commit does not match to our convention, these fields won't have a value.

pomek:
Fixed. Review requested.

pomek:
I found one issue with the tool, so please do not merge this PR.

pomek:
Eh. I cannot reproduce it. Everything works fine so it could be something on my side.

packages/ckeditor5-dev-env/lib/release-tools/utils/getnewreleasetype.js
```diff
@@ -5,7 +5,8 @@
 
 'use strict';
 
-const { commitTypes, transform } = require( '../changelog/writer-options' );
+const { availableCommitTypes } = require( './transform-commit-utils' );
+const transformCommitForCkeditor5Package = require( './transformcommitforckeditor5package' );
```
Reinmar: Another rule – it's `...ForCKEditor5Package`

pomek: Fixed in .5326cae

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
```diff
@@ -0,0 +1,139 @@
...
+ * Parses a single commit:
+ *   - displays a log when the commit has invalid format of the message,
+ *   - filters out the commit if it should not be visible in the changelog,
+ *   - makes links to issues and user's profiles on GitHub.
+ *
+ * @param {Commit} commit
+ * @param {Boolean} [displayLog=true] Whether to display the logs.
+ * @returns {Commit}
+ */
+module.exports = function transformCommitForCkeditor5Package( commit, displayLog = true ) {
```
Reinmar: Couldn't this function be called simpler? More general? Like `transformCommit`? ;) Do you expect to have different implementations in the future? With different rules?

pomek: Yes, for `ckeditor5-dev` we will have another `transformCommit` function.

pomek: To be honest, the whole refactoring is exactly for it.

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelog.js
```diff
@@ -66,10 +66,11 @@ module.exports = function generateChangelog( newVersion = null ) {
 					merges: undefined,
 					firstParent: true
 				};
-				const parserOpts = require( '../changelog/parser-options' );
-				const writerOpts = require( '../changelog/writer-options' );
+				const parserOptions = require( '../utils/parser-options' );
+				const writerOptions = require( '../utils/writer-options' );
+				writerOptions.transform = require( '../utils/transformcommitforckeditor5package' );
```
Reinmar: You must not touch an object which was returned to you by reference. This way you're affecting someone else's `writerOptions` too. So either make `parser-options` and `writer-options` functions which return a new instance every time or clone the object here.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
```diff
@@ -0,0 +1,139 @@
...
+	if ( hasCorrectType && isCommitIncluded ) {
+		logMessage += chalk.green( 'INCLUDED' );
+	} else if ( hasCorrectType && !isCommitIncluded ) {
+		logMessage += chalk.grey( 'SKIPPED' );
+	} else {
+		logMessage += chalk.red( 'INVALID' );
+	}
+
+	if ( commit.header.startsWith( 'Merge' ) && hasCorrectType ) {
+		const indentSize = '[XX:YY:ZZ] * 1234567 '.length;
```
Reinmar: What's this?

packages/ckeditor5-dev-env/lib/release-tools/utils/getnewreleasetype.js
```diff
@@ -5,7 +5,8 @@
 
 'use strict';
 
-const { commitTypes, transform } = require( '../changelog/writer-options' );
+const { availableCommitTypes } = require( './transform-commit-utils' );
+const transformCommitForCKEditor5Package = require( './transformcommitforckeditor5package' );
```
Reinmar: Since a specific "CKEditor 5 package" related "transform commit" function is used here... isn't this function related to "CKEditor 5 package" too?

Wouldn't it make sense if `transformCommit` was this function's param?

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
Reinmar: OK, I wanted to be sure.

Then, the second question is whether we expect to have more functions related to "CKE 5 packages" and "CKE 5 dev packages"? If so, then perhaps we should move this code somewhere else. Keep it closer to each other. E.g. the tasks available in `index.js` are also CKE5 related only so soon we'll have "ForCKEditor5Package" everywhere. Hence, I wanted to propose to try to better organise this code. But I don't know it too well, so nothing concrete.

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelog.js
pomek: Fixed in d9deb9d.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
pomek: This is a size of indent used in changelog tools. 

Changelog prints out the commits which will be used to generate changelog. For merge commit, we displayed "Merge ..." what was unreadable.

I added an additional note to the second line (below the "Merge ..." message) but it does not look good, so I needed to add the indent. 

I just could put the 21 spaces but nobody understands what I did it.

packages/ckeditor5-dev-env/lib/release-tools/utils/getnewreleasetype.js
pomek: In the future, this function will receive a parameter which allows choosing which "transform" function will be used. 

At this moment we have only one transform, so the parameter is not required.

OTOH, I don't know whether the parameter will have a sense. Maybe you are right now.

ATM I have a "YAGNI" feeling.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
ma2ciek: maybe it could be `INDENT_SIZE = '[XX:YY:ZZ] * 1234567 '.length // 21` at the head of this file (?)

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
pomek: > Then, the second question is whether we expect to have more functions related to "CKE 5 packages" and "CKE 5 dev packages"?

Is the question - do we want to use this tool for generating the changelog for the main (`ckeditor5`) package?

I would like to wait for the second "transformCommit" function. In the next PR (which will introduce release tool for this package) I will propose even better the files structure.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
pomek: I moved the `INDENT_SIZE` at the beginning of the file. Thanks, @ma2ciek.

I added also a short comment which explains what it this.

pomek: Fixed in 2bd2a63.

Reinmar: The value of this string was dragging my entire attention so I didn't even check what's after it. Code should be minimal also in this sense – that it doesn't contain distractions like such totally unnecessary values.

packages/ckeditor5-dev-env/lib/release-tools/utils/getnewreleasetype.js
Reinmar: So, if you have YAGNI now, then this entire refactoring may be YAGNI ;). It's the same category of problem - do we need to parametrize the `transformCommit`?

But, it'd also be more clear whether that specific function can be used directly here if the code was structured more clearly. As I mentioned before, you should physically separate it into general release helpers, tasks and helpers for CKEditor 5 packages and tasks and helpers for CKEditor 5 dev stuff. You'd then better see where this function landed. If in the first category, it should not know about `transformCommitForCKEditor5Package`.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
Reinmar: In such case, this refactoring shouldn't be a separate PR because we're not able to review it. It's a WIP – a state of the code in the middle of a refactoring process.

We can merge it like it looks now but it'd be much easier to make decisions if we've seen the entire picture.

packages/ckeditor5-dev-env/lib/release-tools/utils/getnewreleasetype.js
pomek: >  do we need to parametrize the transformCommit?

How do you want to generate changelog for the `ckeditor5-dev` package?

Because we talked about it via Slack, I know what we need to do. If you have a batter idea, let me know.

pomek: As you wanted, I've added a parameter to this function.

pomek:
I'm closing this PR. As you mentioned, this is some part of the whole mechanism.

I will create another PR when the whole mechanism will be ready.

ma2ciek:
Sorry @pomek for reviewing the closed stuff ;)

PR #145, opened 2017-03-28
T/134 Hide logs in translation scripts during `npm test` execution

### LINKED ISSUE #134: [ENV] Transifex logs shouldn't be visible during `npm test`
As in title, when I'm calling the `npm test` command, I would see the tests results. 

At this moment, tests results are mixed with transifex tasks logs:

image

ma2ciek:
Removed displaying logs in the console during `npm test` execution. Closes #134.

Added small test refactor according to our latest experience.


ma2ciek:
Misses the upload part for now.

pomek:
Missing proposal merge commit message.

packages/ckeditor5-dev-env/tests/translations/download.js
```diff
@@ -63,66 +92,44 @@ describe( 'download', () => {
...
-				en_AU: {
-					content: 'ckeditor5-ui-en-content'
-				}
+				en_AU: { content: 'ckeditor5-ui-en-content' }
 			}
 		};
 
-		const fileContents = {
+		fileContents = {
 			'ckeditor5-core-pl-content': {}, // empty
```
pomek: `// empty` :D 

We see that the object is empty. If you didn't explain why this object is empty, I suggest removing this comment.

packages/ckeditor5-dev-env/tests/translations/download.js
```diff
@@ -63,66 +92,44 @@ describe( 'download', () => {
...
 
 			sinon.assert.calledWithExactly(
-				outputFileSyncSpy,
+				stubs.fs.outputFileSync,
 				path.join( 'workspace', 'packages', 'ckeditor5-ui', 'lang', 'translations', 'en-au.po' ),
 				'ckeditor5-ui-en-content'
 			);
+		}, ( err ) => {
+			throw err;
 		} );
```
pomek: Indents in this code are strange.

I suggest creating a variable which will be passed to `download()` function. 

Then you can easily fix the indents:

```js
const credentials = {
	username: '...',
	password: '...'
};

return download( credentials )
	.then(
		() => { /* ... */ },
		() => { /* ... */ }
	);
```

packages/ckeditor5-dev-env/tests/translations/download.js
```diff
@@ -63,66 +92,44 @@ describe( 'download', () => {
...
 			);
 
 			sinon.assert.calledWithExactly(
-				outputFileSyncSpy,
+				stubs.fs.outputFileSync,
 				path.join( 'workspace', 'packages', 'ckeditor5-ui', 'lang', 'translations', 'en-au.po' ),
 				'ckeditor5-ui-en-content'
 			);
+		}, ( err ) => {
+			throw err;
```
pomek: You don't need to catch the error. If something went wrong, the promise will be rejected and test will fail.

packages/ckeditor5-dev-env/tests/translations/upload.js
```diff
@@ -10,67 +10,82 @@
...
 const mockery = require( 'mockery' );
-const transifexService = require( '../../lib/translations/transifex-service' );
-const fs = require( 'fs' );
 
 describe( 'upload', () => {
-	let sandbox;
+	const sandbox = sinon.sandbox.create();
 	let upload;
+	let packageNames;
+	let stubs;
```
pomek: https://github.com/ckeditor/ckeditor5-dev/pull/145/files#r108609189

packages/ckeditor5-dev-env/tests/translations/upload.js
```diff
@@ -10,67 +10,82 @@
...
+			warnOnReplace: false,
+			warnOnUnregistered: false,
+		} );
+
+		stubs = {
+			logger: {
+				info: sandbox.stub(),
+				warning: sandbox.stub(),
+				error: err => {
+					throw err;
```
pomek: https://github.com/ckeditor/ckeditor5-dev/pull/145/files#r108609636

packages/ckeditor5-dev-env/tests/translations/upload.js
```diff
@@ -10,67 +10,82 @@
...
+				getResources: sandbox.spy( () => Promise.resolve( [ {
+					slug: 'ckeditor5-core'
+				} ] ) ),
+				postResource: sandbox.spy( () => Promise.resolve( '[]' ) ),
+				putResourceContent: sandbox.spy( () => Promise.resolve( '{}' ) ),
+			},
+
+			fs: {
+				readdirSync: sandbox.spy( () => packageNames ),
+				createReadStream: sandbox.spy( path => `${ path } content` ),
```
pomek: [`fs.createReadStream`](https://nodejs.org/dist/latest-v6.x/docs/api/fs.html#fs_fs_createreadstream_path_options) returns an instance of ReadStream. It is not a string.

packages/ckeditor5-dev-env/tests/translations/upload.js
```diff
@@ -10,67 +10,82 @@
...
-		const putResourceContentSpy = sandbox.spy( () => Promise.resolve( '{}' ) );
-
-		sandbox.stub( transifexService, 'getResources', getResourcesSpy );
-		sandbox.stub( transifexService, 'postResource', postResourceSpy );
-		sandbox.stub( transifexService, 'putResourceContent', putResourceContentSpy );
-
-		const readDirSyncStub = sandbox.stub( fs, 'readdirSync', () => packageNames );
-		const createReadStreamStub = sandbox.stub( fs, 'createReadStream', ( path ) => `${path} content` );
-
 		return upload( { username: 'username', password: 'password' } ).then( () => {
```
pomek: Indents (https://github.com/ckeditor/ckeditor5-dev/pull/145/files#r108613656).

pomek:
In Travis logs, I found a message which should not appear (because you fix it in this PR)

https://travis-ci.org/ckeditor/ckeditor5-dev/builds/216042049#L1230

packages/ckeditor5-dev-env/tests/translations/upload.js
pomek: You can try to use a mock which I wrote for `conventionalChangelog()` function in `generateChangelog()` task - https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-env/tests/release-tools/tasks/generatechangelog.js#L26-L34.

packages/ckeditor5-dev-env/tests/translations/download.js
ma2ciek: Yes. You're right. I was getting strange errors in the test and I wanted to know whether this causes an issue. The issue was somewhere else and forget to remove that.

packages/ckeditor5-dev-env/tests/translations/download.js
ma2ciek: Actually the indentations are perfect for me:

```js
download( credentials ).then( () => {
   // resolved
}, () => {
  // rejected
} 
```

packages/ckeditor5-dev-env/tests/translations/upload.js
ma2ciek: Yes, they can be in the same line. I will change it.

But I noticed, that if you have few variable declarations in one line, you're not reading them all.

packages/ckeditor5-dev-env/tests/translations/upload.js
ma2ciek: This is a little different situation. Because of the promise catch at https://github.com/ckeditor/ckeditor5-dev/blob/15ccf2b3de21815e95918cf2b0a7037c2ac90da9/packages/ckeditor5-dev-env/lib/translations/upload.js Now I'm not sure if this catch should be there, because it complicates the tests. But from the other hand this function should log the error. Maybe I can throw the caught error and handle it outside this function.

packages/ckeditor5-dev-env/tests/translations/download.js
```diff
@@ -7,50 +7,79 @@
...
-				slug: 'ckeditor5-ui'
-			}
+		resources = [
+			{ slug: 'ckeditor5-core' },
+			{ slug: 'ckeditor5-ui' },
 		];
 
 		// jscs:disable requireCamelCaseOrUpperCaseIdentifiers
-		const resourcesDetails = {
+		resourcesDetails = {
```
ma2ciek: Because `available_languages` isn't a correct variable name.

packages/ckeditor5-dev-env/tests/translations/download.js
pomek: They are hard to read because [you don't see where the first function is ending and where the second function is starting](https://github.com/cksource/mgit2/blob/master/tests/commands/savehashes.js#L57-L99).

packages/ckeditor5-dev-env/tests/translations/download.js
pomek: Sure.

packages/ckeditor5-dev-env/tests/translations/collect-utils.js
```diff
@@ -16,16 +16,42 @@ const glob = require( 'glob' );
 const fs = require( 'fs-extra' );
 
 describe( 'collect-utils', () => {
-	let sandbox;
+	const sandbox = sinon.sandbox.create();
```
ma2ciek: Ok. I can move it inside `beforeEach`.
But I have one advantage of this approach - my IDE recognizes easily the type of the `sandbox` variable. Otherwise it becomes an `any` type.

packages/ckeditor5-dev-env/tests/translations/upload.js
pomek: Logger only writes the logs. Logger must not throw anything. 

In your case I would do something like that (if logging the error is important):

```js
return promise()
    .catch( ( err ) => {
        log.error( err );
        throw err;
    } );
```

packages/ckeditor5-dev-env/tests/translations/download.js
```diff
@@ -7,50 +7,79 @@
...
+				translations: {
+					createDictionaryFromPoFileContent: sandbox.spy( poFileContent => fileContents[ poFileContent ] ),
+					cleanPoFileContent: x => x,
+				},
+				logger() {
+					return stubs.logger;
+				}
+			},
+
+			path: path,
```
ma2ciek: I was playing with the mockery warnings and forgot to remove it.

packages/ckeditor5-dev-env/tests/translations/upload.js
ma2ciek: Ok, so like I have proposed in https://github.com/ckeditor/ckeditor5-dev/pull/145#discussion_r108626192 :)

The logger already logs here some information, so it's naturally to log here the errors as well.

packages/ckeditor5-dev-env/tests/translations/collect-utils.js
ma2ciek: At least we can remove `sandbox.restore()` if we're assigning new sandbox to the variable before each test.

packages/ckeditor5-dev-env/tests/translations/download.js
```diff
@@ -7,50 +7,79 @@
...
+			fs: {
+				outputFileSync: sandbox.spy(),
+			},
+
+			devUtils: {
+				translations: {
+					createDictionaryFromPoFileContent: sandbox.spy( poFileContent => fileContents[ poFileContent ] ),
+					cleanPoFileContent: x => x,
+				},
+				logger() {
```
ma2ciek: Sure.

packages/ckeditor5-dev-env/tests/translations/collect-utils.js
ma2ciek: Yep. You're right. Forgot about it.

ma2ciek:
Could you review it once again @pomek?
AFAIR I didn't change stream mocking, but TBH 100% correctly mocked tests with streams would be much harder to read and understand.

packages/ckeditor5-dev-env/tests/translations/upload.js
```diff
@@ -10,73 +10,91 @@
...
 const sinon = require( 'sinon' );
 const mockery = require( 'mockery' );
-const transifexService = require( '../../lib/translations/transifex-service' );
-const fs = require( 'fs' );
 
 describe( 'upload', () => {
-	let sandbox;
-	let upload;
+	let sandbox, stubs, upload;
+	let packageNames, serverResources, fileContents;
```
pomek: Merge this line with the above one.

ma2ciek: I don't like merging variables into a one big line ;)
In this case these 2 lines differs. In the first there're declared variables that are initialised in the `beforeEach()` hook, while in the second there're declared variables for the later initialisations in the tests.

ma2ciek:
Ow now, didn't catch that while testing locally.

PR #154, opened 2017-04-07
T/121: Changelog for multiple packages in single repository

### LINKED ISSUE #96: Improve changelog:dependencies task process
It now looks like this:

image

One thing to notice is that you have to decide to which version to bump before the changelog entries are printed. This is inconvenient. They should be printed before.

Second thing is that merge commits are printed without the significant line, so you don't know what was that exactly.

### LINKED ISSUE #110: Introduce release tasks to this package

### LINKED ISSUE #121: [ENV] Extract the `transformCommit` function from writer-options.js

### LINKED ISSUE #148: Previous version check leaks
```
[18:44:02] Starting 'release:dependencies'...
? Packages to release:
  * "@ckeditor/ckeditor5-autoformat": 0.4.1 => v0.5.0
  * "@ckeditor/ckeditor5-basic-styles": 0.7.1 => v0.8.0
  * "@ckeditor/ckeditor5-block-quote": 0.0.0 => v0.1.0 (2017-04-05)
### Features
* Introduced the block quote feature. Closes [#1
  * "@ckeditor/ckeditor5-build-classic": 0.0.1 => v0.1.0 (2017-04-05)
### Features
* Initial implementation. Closes [#1
  * "@ckeditor/ckeditor5-clipboard": 0.4.1 => v0.5.0
  * "@ckeditor/ckeditor5-core": 0.7.0 => v0.8.0
  * "@ckeditor/ckeditor5-editor-classic": 0.7.1 => v0.7.2
  * "@ckeditor/ckeditor5-editor-inline": 0.0.1 => v0.1.0 (2017-04-05)
### Features
* Introduced the inline editor. Closes: [#1
  * "@ckeditor/ckeditor5-engine": 0.8.0 => v0.9.0
  * "@ckeditor/ckeditor5-enter": 0.8.0 => v0.9.0
  * "@ckeditor/ckeditor5-heading": 0.8.0 => v0.9.0
  * "@ckeditor/ckeditor5-image": 0.4.0 => v0.5.0
  * "@ckeditor/ckeditor5-link": 0.5.1 => v0.6.0
  * "@ckeditor/ckeditor5-list": 0.5.1 => v0.6.0
  * "@ckeditor/ckeditor5-markdown-gfm": 0.4.1 => v0.4.2
  * "@ckeditor/ckeditor5-paragraph": 0.6.1 => v0.7.0
  * "@ckeditor/ckeditor5-presets": 0.1.1 => v0.2.0
  * "@ckeditor/ckeditor5-theme-lark": 0.6.1 => v0.7.0
  * "@ckeditor/ckeditor5-typing": 0.8.0 => v0.9.0
  * "@ckeditor/ckeditor5-ui": 0.7.1 => v0.8.0
  * "@ckeditor/ckeditor5-undo": 0.7.1 => v0.8.0
  * "@ckeditor/ckeditor5-utils": 0.8.0 => v0.9.0
  * "@ckeditor/ckeditor5-widget": 0.0.1 => v0.1.0 (2017-04-05)
### Features
* Initial implementation (the code was moved from the `ckeditor5-image` package). Closes [#1
Continue? (Y/n) (master c80166f) p@m /workspace/ckeditor5>
```

pomek:
### Merge message:

```
Feature: Introduced task for generating the changelog for multiple packages inside the single repository and fixed various issues related to generating the changelog. Closes #148. Closes #121. Closes #110. Closes #96.

BREAKING CHANGES: Function `executeOnDependencies()` is no longer available. Introduced `getSubPackagesPaths()` or `getSubRepositoriesPaths()` for gathering the package paths and `executeOnPackages()` for executing custom function on these packages.

BREAKING CHANGES: Task `tasks.generateChangelog()` has been renamed to `tasks.generateChangelogForSinglePackage()`.

BREAKING CHANGES: Task `generateChangelogForDependencies()` has been renamed to `tasks.generateChangelogForSubRepositories()`.

BREAKING CHANGES: Task `tasks.createRelease()` has been renamed to `tasks.createReleaseForSubRepository()`.

BREAKING CHANGES: Task `tasks.releaseDependencies()` has been renamed to `tasks.releaseSubRepositories()`.

NOTE: Introduced new task - `tasks.generateChangelogForSubRepositories()`.
```

Issues mentioned in the merge message: #148, #121, #110, #96.

---

In order to test this tool, I suggest cloning my repository – https://github.com/pomek/ckeditor5-dev.
Please remember you need to create a release commits (and tags) manually. 

This PR is a missing part of #144. 

---

First call this tool in `ckeditor5-dev` package will take around 10 min because the whole repository will be parsed. If you want to avoid it, you need to create a CHANGELOG file for each package with an entry for the latest version.


packages/ckeditor5-dev-env/lib/release-tools/utils/getnewreleasetype.js
```diff
@@ -5,66 +5,98 @@
...
+	// @param {Array} commits
+	// @returns {String}
+	function getNewVersionType( commits ) {
+		let hasNewFeatures = false;
+		let hasChanges = false;
+
+		for ( const item of commits ) {
+			const singleCommit = transformCommit( item, context );
 
-		conventionalRecommendedBump( options, parserOpts, ( err, response ) => {
```
pomek: Because `conventionalRecommendedBump` does not allow passing custom options to `git log` command, I had to remove this package and rewrite it. 

Thanks to that I could resolve #96.

packages/ckeditor5-dev-env/lib/release-tools/utils/hascommitsfromlastrelease.js
```diff
@@ -1,30 +0,0 @@
...
-
-const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
-
-/**
- * Checks whether the package has committed changes until the last tag.
- * If the last tag does not exist, the function checks whether any commit exists.
- *
- * @returns {Boolean}
- */
-module.exports = function hasCommitsFromLastRelease() {
```
pomek: It is not needed because [`getNewReleaseType()`]()https://github.com/ckeditor/ckeditor5-dev/pull/154/files#diff-81ddd240a176823f737da5a270374c71 knows all commits and can decide about the release type.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5devpackage.js
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Parses a single commit for `ckeditor5-dev` package.
+ *
+ * @param {Commit} commit
+ * @param {Object} {context}
+ * @param {Boolean} context.displayLogs Whether to display the logs.
+ * @param {Object} context.packageData Content from the 'package.json' for given package.
+ * @returns {Commit}
+ */
+module.exports = function transformCommitForCKEditor5DevPackage( commit, context ) {
```
pomek: We accept the commit only if its files are located in current processing package. [See the tests.](https://github.com/ckeditor/ckeditor5-dev/pull/154/files#diff-c4d9754e1f61e34742144bb2dcc0d345R62)

packages/ckeditor5-dev-env/lib/release-tools/utils/versions.js
```diff
@@ -17,11 +17,19 @@ const versions = {
...
+
+		const regexp = /\n## (\[([^\]]+)|[^ ]+)/;
 		const matches = changelog.match( regexp );
 
-		return matches ? matches[ 1 ] : null;
+		if ( !matches ) {
+			return null;
+		}
+
+		return matches[ 2 ] ? matches[ 2 ] : matches[ 1 ];
```
pomek: Now we handle both types of changelog entry:

1. For entry generated for the first time:
```md
CHANGELOG
==========

## version (date)
```

2. For the next entries:
```md
CHANGELOG
==========

## [version](URL to compare the changes) (date)
```

packages/ckeditor5-dev-env/lib/release-tools/utils/executeondependencies.js
```diff
@@ -26,7 +28,7 @@ const getPackageJson = require( './getpackagejson' );
  */
 module.exports = function executeOnDependencies( options, functionToExecute ) {
 	const packagesAbsolutePath = path.join( options.cwd, options.packages );
-	const directories = workspaceUtils.getDirectories( packagesAbsolutePath );
+	const directories = tools.getDirectories( packagesAbsolutePath );
```
pomek: Because we check whether the directory is a valid package, we can take everything from the CWD (instead of directories starting with `ckeditor5-`).

pomek:
I left a few comments which should explain why I did something this way.

packages/ckeditor5-dev-env/lib/release-tools/utils/versions.js
pomek: It fixes #148.

ma2ciek:
I'm going to make a big, big coffee :D

packages/ckeditor5-dev-env/lib/index.js
```diff
@@ -26,26 +26,57 @@ const tasks = {
...
 		return executeOnDependencies( execOptions, generateChangelogForSinglePackage )
 			.then( ( skippedPackages ) => {
 				displaySkippedPackages( skippedPackages );
 
 				process.chdir( options.cwd );
+			} )
+			.then( () => {
+				const packageNames = Object.keys( generatedChangelog );
+
+				if ( !packageNames.length ) {
```
ma2ciek: Don't we want to log here something as well?

packages/ckeditor5-dev-env/lib/index.js
```diff
@@ -26,26 +26,57 @@ const tasks = {
...
+	 * @param {Boolean} [options.checkPackageJson=true] If set to false, the mechanism will not check whether
+	 * the current package being specified in 'package.json' file.
 	 * @returns {Promise}
 	 */
 	generateChangelogForDependencies( options ) {
 		const execOptions = {
 			cwd: options.cwd,
 			packages: options.packages,
-			skipPackages: options.skipPackages || []
+			skipPackages: options.skipPackages || [],
```
ma2ciek: `skippedPackages`

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelog.js
```diff
@@ -91,12 +111,29 @@ module.exports = function generateChangelog( newVersion = null ) {
...
 
 				log.info( chalk.green( `Changelog for "${ packageJson.name }" (v${ version }) has been generated.` ) );
 
-				resolve();
+				resolve( version );
 			} );
 		}
 	} );
+
+	function getTransformFunction() {
```
ma2ciek: this function may be easily moved outside the main function

packages/ckeditor5-dev-env/lib/index.js
pomek: For what? "Skip packages" is a list of packages to skip.  "Skipped packages" means "list of packages that have been skipped".

pomek: Instead of `skipPackages`, it could be `packagesToSkip`. But it is not a part of this PR.

ma2ciek: AFAIK skip can be used as a verb / noun. But it's rather just a naming convention. `skipPackages` suggests that the variable is a function.

packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit-utils.js
```diff
@@ -0,0 +1,96 @@
...
+		'NOTE': 2
+	},
+
+	/**
+	 * Changes user's name to link that leads to the user's profile.
+	 *
+	 * @param {String} value
+	 * @returns {String}
+	 */
+	linkGithubUsers( value ) {
```
ma2ciek: A parameter could be more precise.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5devpackage.js
```diff
@@ -0,0 +1,45 @@
...
+'use strict';
+
+const transformCommitForCKEditor5Package = require( './transformcommitforckeditor5package' );
+const getChangedFilesForCommit = require( './getchangedfilesforcommit' );
+
+/**
+ * Parses a single commit for `ckeditor5-dev` package.
+ *
+ * @param {Commit} commit
+ * @param {Object} {context}
```
ma2ciek: small typo - `@param {Object} context`

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5package.js
```diff
@@ -0,0 +1,144 @@
...
+const INDENT_SIZE = 21;
+
+/**
+ * Parses a single commit:
+ *   - displays a log when the commit has invalid format of the message,
+ *   - filters out the commit if it should not be visible in the changelog,
+ *   - makes links to issues and user's profiles on GitHub.
+ *
+ * @param {Commit} commit
+ * @param {Object} {context}
```
ma2ciek: and here as well

packages/ckeditor5-dev-env/lib/index.js
pomek: Not sure if we need an information when we skipped all packages.

ma2ciek:
For me whole PR looks very good. Would you have a look @Reinmar?

pomek:
This PR is required for me before I start resolving other tickets. 

I modified a lot of files and I moved a lot of files from their original localization. I would like to avoid conflicts between changed files when we close another ticket that changed already moved a file in this PR.

Reinmar:
I'll be reviewing it today but most likely it will take me until the end of the week. You're on holidays anyway, so it'll be fine, right?

pomek:
Yes. Will be nice to have it next week.

packages/ckeditor5-dev-env/lib/index.js
Reinmar: I think that since this is a config option and not a property of some stateful object, it's fine that it's called like the `--skip` option. Besides, it may even become a `{Function|Array}` one day and what's then?

Reinmar:
The biggest issue with this PR is that it tries to squash everything into one function – `generateChangelogForDependencies()` and set parametrise it through options. This means that this single function becomes very complex and that these changes affect how it worked in the past. It's far harder now to validate these changes.

DRY doesn't mean that we must strive to create the least number of functions. It can actually mean the opposite if KISS is involved.

In this case, we have clear two entry points – generating changelog for sub repos and generating changelog for sub packages. While they may share a lot of the same logic (which can be extracted to some meaningful utils), they also differ significantly. You need to identify the parts which differ and implement them separately. Avoid config options – that should be your rule of thumb.

E.g if one of these functions need to verify something with `package.json`, then only this function need to call some `verifyIsDependency()` function. The other function shouldn't be concerned by it.

Another thing is actually finding the right "generalisation" of te problem you're dealing with. You introduced option `isDevPackage` forcing somehow that this is only used for dev packages. If, we'd have another repo which is similar to ckeditor5-dev, we shouldn't use that option... so what's then?

Hence, you need to do a step back and look at the situation. First of all, the current name `generateChangelogForDependencies()` already imposes something – it's about dependencies. Then you have `executeOnDependencies()` which should also be about dependencies. But are they in the case of ckeditor5-dev? Nope, they are not. You even needed to introduce the `checkPackageJson` option denying these functions' names.

So, what's a better abstraction? There are many options, but my first choice would be:

* `changelog.generateForSubRepositories()` and `changelog.generateForSubPackages()`
* `executeOnPackages()` and `getSubRepositories()` and `getSubPackages()`. Note that the difference between fetching sub repos and sub packages will be handled by the two last functions. The first function's job should be to run code on them which is in line with SRP and which actually may make it unnecessary.

Going further, we can see that we will have functions which are generic (like `executeOnPackages()`) and functions which will server one of the two modes – sub repos and sub packages. Depending on number of them it may be reasonable to e.g. split the `utils/*` to `utils/*`, `utils/packages/`, `utils/repositories/`. It's best to do such things once situation clarifies.

To sum up – I'll perform a small review, but the first thing to do is to clean up the code anyway. Only then it will be possible to check everything, so right now I'll just focus on more visible things.

packages/ckeditor5-dev-env/lib/index.js
```diff
@@ -26,26 +26,57 @@ const tasks = {
...
+			checkPackageJson: options.checkPackageJson === undefined ? true : options.checkPackageJson
```
Reinmar: Optional boolean flags should be `false` by default to avoid such a code and just because it's more intuitive that when you pass some option you enable something. So, in this case it should be `skipPackageJsonCheck`. But of course, the whole option will rather be unnecessary after refactor.

packages/ckeditor5-dev-env/lib/index.js
Reinmar: Let's leave it – it's very unlikely anyway.

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelog.js
```diff
@@ -91,12 +111,29 @@ module.exports = function generateChangelog( newVersion = null ) {
 
 				// Commit the changelog.
 				tools.shExec( `git add ${ utils.changelogFile }`, { verbosity: 'error' } );
-				tools.shExec( `git commit -m "Docs: Changelog. [skip ci]"`, { verbosity: 'error' } );
+
+				let commitMessage;
+
+				if ( options.isDevPackage ) {
+					commitMessage = `Changelog for "${ packageJson.name }".`;
```
Reinmar: Why committing every changelog separately? That's exactly one releasing sub packages should work differently than sub repos – in that case, there should be just one commit. Taken the number of code branching in this function you may have a perfect proof that it just shouldn't be the same function.

packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitforckeditor5devpackage.js
Reinmar: Isn't there more gitty way of retrieveing such commits? Like in http://stackoverflow.com/questions/16343659/how-to-show-git-log-history-for-a-sub-directory-of-a-git-repo?

packages/ckeditor5-dev-env/lib/release-tools/utils/getnewreleasetype.js
```diff
@@ -5,66 +5,98 @@
...
-module.exports = function getNewReleaseType() {
-	const conventionalRecommendedBump = require( 'conventional-recommended-bump' );
-	const parserOpts = require( '../changelog/parser-options' );
+module.exports = function getNewReleaseType( transformCommit, isDevPackage ) {
+	const packageJson = getPackageJson();
+	let fromVersion = versions.getLastFromChangelog();
+
+	if ( fromVersion ) {
+		if ( isDevPackage ) {
+			fromVersion = packageJson.name + '@' + fromVersion;
```
Reinmar: I'm not sure about the fact that in one case we use `vx.y.z` notation and in the other case there's no `v`. I see that Lerna doesn't use `v` in its tags... so we can follow that, but then let's this logic extracted somewhere because I can see that the same code is repeated twice.

packages/ckeditor5-dev-env/lib/release-tools/utils/versions.js
Reinmar: This condition should be unnecessary. It wasn't necessary when I patched this in https://github.com/ckeditor/ckeditor5-dev/compare/t/148?expand=1.

pomek:
Guys (@Reinmar, @ma2ciek), ready to review. I will modify the proposed merge message.

Reinmar:
I merged the code and... it turned out that the new task doesn't work. Running `gulp changelog` in ckeditor5-dev does nothing now. It worked on t/121 though. The task may have problems with reading merge commits.

Reinmar:
Hm... I don't know why it works, but the commit is apparently rejected at some point. `gitRawCommits()` return the proper stuff:

```
(master fefc1de ?1) p@m /workspace/ckeditor5-dev> git log --format="%B%n-hash-%n%H%n------------------------ >8 ------------------------" @ckeditor/ckeditor5-dev-env@4.4.3..HEAD --first-parent
Merge pull request #154 from ckeditor/t/121

Feature: A task for generating changelogs in a monorepo was introduced. Several other improvements were made on the occasion. Closes #148. Closes #121. Closes #110. Closes #96.

BREAKING CHANGES: Task `tasks.generateChangelog()` has been renamed to `tasks.generateChangelogForSinglePackage()`.

BREAKING CHANGES: Task `generateChangelogForDependencies()` has been renamed to `tasks.generateChangelogForSubRepositories()`.

BREAKING CHANGES: Task `tasks.createRelease()` has been renamed to `tasks.releaseRepository()`.

BREAKING CHANGES: Task `tasks.releaseDependencies()` has been renamed to `tasks.releaseSubRepositories()`.

NOTE: Introduced a new task – `tasks.generateChangelogForSubRepositories()`.
-hash-
fefc1de92d89761e30cbc1b23d09ece0ce6403f8
------------------------ >8 ------------------------
```

So, the problem lays somewhere later. Most likely, in getting the list of changed files:

```
# normal commit:
(master fefc1de ?1) p@m /workspace/ckeditor5-dev> git show --pretty="" --name-only 118fccb2
packages/ckeditor5-dev-env/lib/release-tools/utils/versions.js
packages/ckeditor5-dev-env/tests/release-tools/utils/changelog.js
packages/ckeditor5-dev-env/tests/release-tools/utils/versions.js
# merge commit:
(master fefc1de ?1) p@m /workspace/ckeditor5-dev> git show --pretty="" --name-only fefc1de
```

PR #166, opened 2017-04-18
Added support for the --server option

### LINKED ISSUE #165: Allow passing empty --browsers option to `gulp test`
Currently, it doesn't seem possible to just start the Karma server without launching any browser. This is inconvenient if I want to run test on a browser which I want to launch myself (e.g. on a VM).

I don't know, though, how to tell `gulp test` to not start any browser. `gulp test --browser=""`? :D Or `gulp test --server`?

Reinmar:
Feature: Added support for the `--server` option to the automated tests runner. Closes #165.

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
+++ b/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
@@ -42,18 +42,6 @@ module.exports = function getKarmaConfig( options ) {
 		}
 	}
 
-	let browsers = [];
-
-	if ( options.browsers ) {
-		browsers = options.browsers.map( ( browser ) => {
-			if ( browser === 'Chrome' ) {
-				return 'CHROME_LOCAL';
-			}
-
-			return browser;
-		} );
-	}
-
 	const karmaConfig = {
 		// Base path that will be used to resolve all patterns (eg. files, exclude).
 		basePath: process.cwd(),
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
```diff
@@ -174,3 +158,22 @@ module.exports = function getKarmaConfig( options ) {
...
 };
 
+// Returns the value of Karma's browser option.
+// @returns {Array|null}
+function getBrowsers( options ) {
+	if ( process.env.TRAVIS ) {
+		return [ 'CHROME_TRAVIS_CI' ];
+	}
+
+	if ( options.server || !options.browsers ) {
```
ma2ciek: I see, that you handled both `--server` and `--browser=''` :)

Reinmar: I just wanted to keep this: https://github.com/ckeditor/ckeditor5-dev/pull/166/files#diff-31df1471c2e1fbb9040249aaaf4d952aL47, but I changed the behaviour a bit during refactoring. That wasn't intentional, but it seems that I, indeed, handled `--browsers=''` :D

ma2ciek:
As I made a quick research I found the `--no-brosers` option in the [docs](http://karma-runner.github.io/1.0/config/configuration-file.html) to Karma`s command line. IMO it sounds better and we'll keep the convention in the option names.

Reinmar:
Hm... with `--no-browsers` we'd have to also use `-w` (starting Karma in single run mode without browsers doesn't seem particularly useful).

ma2ciek:
Yes, we'd have to. So maybe we should keep the current version with `--server`

PR #169, opened 2017-04-25
Make TranslationService more generic

### LINKED ISSUE #167: Webpack plugin should be able to parse and translate other files as well
For now webpack plugin is able to translate the `ckeditor-*/**` files only. It would be nice if it could translate other files, e.g. some custom 3rd party plugins as for sure there will be such a need.

ma2ciek:
Closes #167.

Provided an optional parameter (`getPathToPoFile` ) in `TranslationService` constructor instead of complicating the plugin, so now it can be used outside the `ckeditor5-dev` e.g. for some custom plugins.


packages/ckeditor5-dev-utils/lib/translations/translationservice.js
```diff
@@ -95,6 +99,8 @@ module.exports = class TranslationService {
 	// Loads translations from the po file.
 	_loadPoFile( pathToPoFile ) {
 		if ( !fs.existsSync( pathToPoFile ) ) {
+			console.warn( `PO file doesn't exists under the path: ${ pathToPoFile }.` );
```
pomek: PO file doesn't exist…

packages/ckeditor5-dev-utils/tests/translations/translationservice.js
```diff
@@ -17,7 +17,7 @@ const mockery = require( 'mockery' );
 describe( 'translations', () => {
 	describe( 'TranslationService', () => {
 		const sandbox = sinon.sandbox.create();
-		let translationService, stubs;
+		let translationService, TranslationService, stubs;
```
pomek: `translationService` and `TranslationService` are similar to each other. The first one is an instance of `TranslationService`, the second one is the class definition. In order to avoid troubles, I would like to remove the `translationService` variable from this place and move it to test scenario. Most probably it will be repeated in each test scenario but the whole file will be easier to read.

ma2ciek: > Most probably it will be repeated in each test scenario but the whole file will be easier to read.

Yes, I wanted to avoid it, but your idea is probably better.

ma2ciek:
Fixed

Reinmar:
I can't test all the tools now due to https://github.com/ckeditor/ckeditor5-dev/issues/174 but the change LGTM too.

PR #170, opened 2017-04-25
Do not output html when using jsdoc. Modified json output path. Close…

m-turek:
…s #168.

### DIFF
--- a/packages/ckeditor5-dev-docs/lib/index.js
+++ b/packages/ckeditor5-dev-docs/lib/index.js
@@ -5,8 +5,6 @@
 
 'use strict';
 
-const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
-const path = require( 'path' );
 const gulp = require( 'gulp' );
 const jsdoc = require( 'gulp-jsdoc3' );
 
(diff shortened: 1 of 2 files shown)


packages/jsdoc-plugins/lib/utils/doclet-logger.js
```diff
@@ -13,6 +13,6 @@ const fsExtra = require( 'fs-extra' );
  */
 exports.handlers = {
 	processingComplete( e ) {
-		fsExtra.outputFileSync( process.cwd() + '/build/docs/output.json', JSON.stringify( e, null, 4 ) );
+		fsExtra.outputFileSync( process.cwd() + '/docs/api-docs/output.json', JSON.stringify( e, null, 4 ) );
```
Reinmar: To not repeat ourselves – I'd propose `docs/api/output.json`.

packages/ckeditor5-dev-docs/lib/index.js
```diff
@@ -20,36 +18,32 @@ module.exports = {
...
-					require.resolve( '@ckeditor/jsdoc-plugins/lib/validator/validator' ),
-					require.resolve( '@ckeditor/jsdoc-plugins/lib/utils/doclet-logger' )
-				]
-			};
+	const jsDocConfig = {
+		opts: {
+			encoding: 'utf8',
+			recurse: true,
+			access: 'all',
+			template: 'templates/silent'
```
Reinmar: Does it have to be added to deps or does it come with JSDoc?

m-turek: - `encoding`is not needed as `utf8` is default value

- `recurse` is not needed because `docs` task doesn't traverse directories (https://github.com/ckeditor/ckeditor5/blob/master/gulpfile.js#L21-L31)

- `access` should stay as `all` because by default `private` symbols are skipped by jsdoc

- `template` should stay cause `silent` isn't default

PR #192, opened 2017-05-10
T/191: Switch to ESLint

### LINKED ISSUE #11: Each task should get options separately
Moved from: https://github.com/ckeditor/ckeditor5-dev-env/issues/27

### LINKED ISSUE #191: Switch to ESLint
Described here: https://github.com/ckeditor/ckeditor5/issues/366 and here https://github.com/ckeditor/ckeditor5-dev/pull/189#issuecomment-300157024.

pomek:

Feature: Introduced ESLint instead of JSHint. Closes #191. Closes #11.

BREAKING CHANGES: Each task will get options separately.

---

### Additional information

~~After merging this PR, a command `gulp lint` executed in this repository will show a lot of errors that will be fixed in another PR.~~ _Fixed in this PR._


.eslintrc
```diff
@@ -126,7 +126,7 @@
     "no-trailing-spaces": "error",
     "max-len": [
       "error",
-      150
+      120
```
Reinmar: In jscsrc it is 140.

Reinmar:
Uuuuh... That was a lot of work, but I reviewed all the rules. Turned out that many were not enabled. I'll now do some checks how the config I created behaves for ckeditor5-* packages.

Reinmar:
Okey dokey. So, the rules look rather good. I think that in this specific repo (ckeditor5-dev) we only need to look for some remaining JSHint and JSCS comments in the code (`global(s)`, `jscs:disable`, etc.). After that, it should be ready.

However, before we close it, it'd be good to also have a PR for https://github.com/ckeditor/ckeditor5-engine/issues/953. Which, unfortunately, means about 2000 errors. **FORTUNATELY** ESLint's fixer works very well. So the first step will be to run there:

```
eslint  --fix src tests
```

and check the changes it proposed. If you'll do this, make sure to commit this part of work as a separate commit. After that, you'll need to work on the other issues (with JSCS can't fix).

Reinmar:
Damn... I've just looked at the Travis logs and it turns out that my changes in this repo caused two tests to fail ;|. Either I did some mistake or changes proposed by ESLint weren't transparent. I'm afraid you'll need to check it.

pomek:
Something is wrong with the ESLint config.

I fixed (and pushed) failed tests - https://github.com/ckeditor/ckeditor5-dev/pull/192/commits/77b40082a74d09b10ee4735e2c0bfba4eb18575f. Unfortunately, linter did not allow committing the changes.

image

Reported errors change the behavior of the code…

pomek:
>  I think that in this specific repo (ckeditor5-dev) we only need to look for some remaining JSHint and JSCS comments in the code (global(s), jscs:disable, etc.).

Done in b455219.

Reinmar:
This is big WTH... I need to look into this ;|

Reinmar:
Turns out that you reverted too many changes: https://github.com/ckeditor/ckeditor5-dev/pull/192/commits/53178e2496a90582c625485a8d6d934902a3429e.

Reinmar:
And the method shorthand behaviour is a big surprise: https://twitter.com/reinmarpl/status/862956665330642944.

pomek:
Tests passed! :D

packages/ckeditor5-dev-env/tests/release-tools/tasks/creategithubrelease.js
```diff
@@ -13,6 +11,7 @@ const mockery = require( 'mockery' );
 
 describe( 'dev-env/release-tools/tasks', () => {
 	describe( 'createGithubRelease()', () => {
+		// eslint-disable-next-line no-unused-vars
```
Reinmar: Why was this comment added here? Why declaraing unused vars?

packages/ckeditor5-dev-utils/tests/stream.js
```diff
@@ -44,7 +42,7 @@ describe( 'stream', () => {
 		} );
 
 		it( 'should wait until a promise returned by the callback is resolved', ( ) => {
-			let resolve, resolved;
+			let resolve, resolved; // eslint-disable-line
```
Reinmar: Why is this needed?

packages/jsdoc-plugins/lib/longname-fixer/fixers/fix-short-refs.js
```diff
@@ -15,7 +15,7 @@ const assign = Object.assign;
  * @returns {Options}
  */
 function fixShortRefs( options ) {
-	let { doclet, lastInterfaceOrClass } = options;
+	let { doclet, lastInterfaceOrClass } = options; // eslint-disable-line prefer-const
```
Reinmar: Again, why is this needed?

Reinmar:
OK, I've checked some `eslint-disable*` usage and I can tell that most of them should be removed. The code was wrong there. The only ones which make sense are `max-len` disabling but only for hardly avoidable things (like long strings – I saw some imports like this).

pomek:
@Reinmar, could you review it once again? 

I fixed the code according to your points.

Reinmar:
On it!

packages/ckeditor5-dev-env/tests/release-tools/utils/generatechangelogfromcommits-integration.js
```diff
@@ -74,10 +74,11 @@ describe( 'dev-env/release-tools/utils', () => {
 			sandbox.restore();
 		} );
 
-		it( 'generates a changelog for the first time', () => {
+		// Done callback is called manually because these tests are slow and sometimes Mocha throws a timeout error.
+		it( 'generates a changelog for the first time', done => {
```
Reinmar: You should not do these changes in a PR which is completely unrelated. I couldn't understand what ESLint rule might require such changes ;|

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
```diff
@@ -97,7 +97,7 @@ const utils = {
 		}
 
 		return [ ...usedContextMap ]
-			.filter( ( [ key, usage ] ) => !usage ) // eslint-disable-line no-unused-vars
```
Reinmar: I'd actually revert this change. It's very strange that ESLint doesn't understand that you need only `usage` out of these 2 – just like with callback arguments. Sometimes you define more than you use.

packages/ckeditor5-dev-env/tests/release-tools/utils/generatechangelogfromcommits-integration.js
pomek: These tests fail randomly. We can revert it and restart the build on CI when they fail if we don't want to fix it in this PR.

Reinmar: The right way to go would to be to fix this on master and merge master into this branch. But since you already did this change here, we can live with it.

packages/ckeditor5-dev-env/lib/translations/collect-utils.js
Reinmar: Turns out you can write `[ , usage ]`. But it's a pity that ESLint doesn't support `args: 'after-used'` for this case too.

packages/ckeditor5-dev-env/tests/release-tools/utils/generatechangelogfromcommits-integration.js
pomek: Thanks and sorry.

PR #197, opened 2017-05-11
T/147: If there is not tag in the repo release:dependencies throws

### LINKED ISSUE #147: If there is not tag in the repo release:dependencies throws
```
(master c80166f) p@m /workspace/ckeditor5> gulp release:dependencies
[18:37:35] Using gulpfile /workspace/ckeditor5/gulpfile.js
[18:37:35] Starting 'release:dependencies'...
[18:37:36] Error while executing git describe --abbrev=0 --tags 2> /dev/null:
```

pomek:

Fix: A util `getPackagesToRelease()` will not break the release process when some package is released for the first time. Closes #147.


pomek:
I need to test this change manually. Please give me few mins.

pomek:
Works fine!

image

image

https://github.com/pomek/ckeditor5-d/releases/tag/v0.0.2

packages/ckeditor5-dev-env/tests/release-tools/utils/getpackagestorelease.js
```diff
@@ -192,5 +192,43 @@ describe( 'dev-env/release-tools/utils', () => {
 					expect( packages.get( '@ckeditor/ckeditor5-utils' ) ).to.equal( undefined );
 				} );
 		} );
+
+		it( 'does not throw when packages are released for the first time', () => {
```
Reinmar: This test does not throw without the fix. So. what is it checking?

Reinmar:
Missing an integration test. One of the new tests doesn't check this case and the other is very low level. How do you know that the code really works?

pomek:
As you reported in a ticket:

> ```
> [18:37:36] Error while executing git describe --abbrev=0 --tags 2> /dev/null:
> ```

This error was thrown by one of the utils (`version.js`) so I fixed it and `getPackagesToRelease()` will not break the release if tag will not exist.

Reinmar:
How do you know that something else won't break? Perhaps you tested it, but I don't want to create a new repo just to test it and repeat that N times in the future every time we'll change something in this code.

Unit tests may verify a piece of logic. Integration tests verify features. I'd like this feature to be tested well because it's a complex piece of code and we had issues in it.

pomek:
Will be fixed in #216 because it delivers an integration test.

PR #208, opened 2017-05-18
T/184: Add "internal" option to the version picker 

### LINKED ISSUE #184: Add "internal" option to the version picker
When picking a version suggested by the changelog task, sometimes you'd like to say that all changelog entries should be ignored (because they are relevant to other subpackages, not this one) and that this should be a patch release for an internal change (with that nice "Internal changes only (updated dependencies, documentation, etc.)." text).

So, this is like picking the patch bump option, but with discarding all the changelog entries.

pomek:

Feature: Added `internal` option to the version picker. Closes #184.

If you type `internal` as a new version during generating the changelog, all commits were made will be ignored.

---

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/templates/template.hbs
+++ b/packages/ckeditor5-dev-env/lib/release-tools/templates/template.hbs
@@ -1,5 +1,9 @@
 {{> header}}
 
+{{#if isInternalRelease}}
+Internal changes only (updated dependencies, documentation, etc.).
+
+{{else}}
 {{#each commitGroups}}
 {{#if title}}
 ### {{title}}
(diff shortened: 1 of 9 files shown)


packages/ckeditor5-dev-env/lib/release-tools/templates/template.hbs
```diff
@@ -1,5 +1,9 @@
 {{> header}}
 
+{{#if isInternalRelease}}
```
pomek: I needed to duplicate the sentence because Handlebars does not support expression like this `{{#if condition1 || condition2}}`.

Reinmar: Srsly? ::||

PR #218, opened 2017-05-26
T/217: Utils - Bundler - "createEntryFile" should use "config" key from configuration build

### LINKED ISSUE #217: Utils - Bundler - "createEntryFile" should use "config" key from configuration build
I don't know how to describe the issue. Changes are required for https://github.com/ckeditor/ckeditor5-build-classic/issues/10#issuecomment-303398682.

What we need to do is change `./packages/ckeditor5-dev-utils/lib/bundler/createentryfile.js`:

```diff
- config: require( '${ configPath }' )
+ config: require( '${ configPath }' ).config
```

and update the docs (and tests).

pomek:

Other: `createEntryFile()` util will use the "config" key from configuration file. Closes #217.

[Read more why it is required and important.](https://github.com/ckeditor/ckeditor5-build-classic/issues/10#issuecomment-303398682)

BREAKING CHANGE: `createEntryFile()` function does not accept the `configPath` argument any more. You need to pass the configuration object as `options.config` instead of it.

---

### Additional information

Required for https://github.com/ckeditor/ckeditor5-build-classic/pull/13.

### DIFF
--- a/packages/ckeditor5-dev-utils/lib/bundler/geteditorconfig.js
+++ b/packages/ckeditor5-dev-utils/lib/bundler/geteditorconfig.js
@@ -19,13 +19,6 @@ module.exports = function getEditorConfig( config ) {
 	}
 
 	return javascriptStringify( config, null, '\t' )
-		.split( '\n' )
-		.map( ( line, index ) => {
-			if ( index === 0 ) {
-				return line;
-			}
-
-			return `\t${ line }`;
-		} )
-		.join( '\n' );
+		// Indent all but the first line (so it can be easily concatenated with `config = ${ editorConfig }`).
+		.replace( /\n/g, '\n\t' );
 };
--- a/packages/ckeditor5-dev-utils/lib/bundler/geteditorconfig.js
+++ b/packages/ckeditor5-dev-utils/lib/bundler/geteditorconfig.js
@@ -1,0 +1,39 @@
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+/**
+ * Transforms specified configuration to a string which can be inserted
+ * to generated an entry file.
+ *
+ * Keys and values will be wrapped in apostrophes instead of quotation marks.
+ *
+ * Keys which contains letters or/and numbers or/and underscore or/and a Dollar sign
+ * will not be wrapped in apostrophes.
+ *
+ * @param {Object} config
+ * @returns {String}
+ */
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-dev-utils/lib/bundler/geteditorconfig.js
```diff
@@ -0,0 +1,39 @@
...
+		.map( ( line, index ) => {
+			if ( index === 0 ) {
+				return line;
+			}
+
+			return `\t${ line }`;
+		} )
+		.join( '\n' )
+		// Replace quotation marks to apostrophes.
+		.replace( /"/g, '\'' )
```
Reinmar: This is potentially breaking. You can have `"\""` string and you'll end up with `'\''` which is a different string.

pomek: Good one. Will fix it.

Reinmar:
You could check this library: https://www.npmjs.com/package/javascript-stringify

Reinmar:
It will be safer than using JSON.stringify and operating on that.

pomek:
Done. `javascript-stringify` works fine.

packages/ckeditor5-dev-utils/lib/bundler/geteditorconfig.js
```diff
@@ -0,0 +1,37 @@
...
+
+'use strict';
+
+const javascriptStringify = require( 'javascript-stringify' );
+
+/**
+ * Transforms specified configuration to a string which can be inserted
+ * to generated an entry file.
+ *
+ * Keys and values will be wrapped in apostrophes instead of quotation marks.
```
Reinmar: I wanted to comment on this documentation previously that it goes into implementation details rather than saying what it tries to achieve (which is – printing the object in our code style).

And now, notice how much outdated this documentation is after you changed the implementation ;)

Reinmar:
Please improve the API docs.

pomek:
Done.

packages/ckeditor5-dev-utils/lib/bundler/geteditorconfig.js
```diff
@@ -0,0 +1,31 @@
...
+ * @param {Object} config
+ * @returns {String}
+ */
+module.exports = function getEditorConfig( config ) {
+	if ( !config ) {
+		return '{}';
+	}
+
+	return javascriptStringify( config, null, '\t' )
+		.split( '\n' )
```
Reinmar: `.replace( /\n/g, '\n\t' )` seems to be an equivalent. I'll check it.

PR #225, opened 2017-06-05
T/224: [Lint] `getGitIgnore()` should transform directories as the glob patterns for Gulp

### LINKED ISSUE #224: [Lint] `getGitIgnore()` should transform directories as the glob patterns for Gulp
Described here: https://github.com/ckeditor/ckeditor5/issues/456#issuecomment-305742908.

pomek:

Feature: `getGitIgnore()` will transform directories to glob patterns. Closes #224.

Required in order to [use a more standard `"directory/"` notation in `.gitignore`](https://github.com/ckeditor/ckeditor5/issues/456).

### DIFF
--- a/packages/ckeditor5-dev-lint/lib/utils/getgitignore.js
+++ b/packages/ckeditor5-dev-lint/lib/utils/getgitignore.js
@@ -26,10 +26,10 @@ module.exports = function getGitIgnore() {
 		.map( i => '!' + i )
 		// Add "**" to folders to have everything inside them ignored.
 		.map( path => {
-			if ( !path.endsWith( '/' ) ) {
-				return path;
+			if ( path.endsWith( '/' ) ) {
+				return path + '**';
 			}
 
-			return path + '**';
+			return path;
 		} );
 };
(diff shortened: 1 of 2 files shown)


Reinmar:
What we should check on this occasion is whether we really need dev-lint package.

Let's say that ESLint can check `.gitignore` itself (a reasonable assumption, I think) and that it also accepts some other patterns (we need to ignore the `src/lib/` directory). This would mean that everything the `dev-lint` package does could be easily replaced with ESLint, rendering this package unnecessary.

Let's think about this because I'd gladly remove this package and moved the linting script to package's gulpfiles.

PS. I forgot about getting `pre-commit` files. I don't think that ESLint can do that out of the box. But nevertheless. So... ignore my comment :D

Reinmar:
There's `lint-staged` https://medium.com/@okonetchnikov/make-linting-great-again-f3890e1ad6b8 and I quite like this:

```
// package.json
...
"scripts" {
 "lint-staged": "lint-staged",
 "stylelint-staged": "stylelint --syntax scss"
},
"lint-staged": {
 "*.js": "eslint",
 "*.scss": "stylelint-staged"
},
"pre-commit": "lint-staged"
...
```

This would let us remove the `gulp` depdendency from all small packages and `gulpfile.js` and `dev-lint` of course.

This is an interesting direction... I'll report a separate ticket for it because it's a bigger topic of course.

PR #227, opened 2017-06-08
Add JSDoc plugin for static members inheritance

### LINKED ISSUE #226: Update JSDoc plugin for static members inheritance.
Related to #201 . 
Previous plugin did not handle multiple levels of inheritance and added only static members from direct parent class.
Static members were also not added from implemented interfaces or mixed mixins.

Changed the plugin to handle multiple levels of inheritance, implemented interfaces and mixed mixins.

m-turek:

Feature: JSDoc plugin enabling inheritance of static members. Closes #226 .

---

### Additional information

* Handles multiple levels of inheritance.


m-turek:
Tests pass, one failed cause of timeout.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+const isNonEmptyArray = require( './utils/isnonemptyarray' );
+const cloneDeep = require( 'lodash' ).cloneDeep;
+
+/**
+ * Adds missing doclets of static members coming from implemented interfaces and extended classes.
+ * JSDoc does not support inheritance of static members.
+ * This module requires the input to be processed by 'relationbuilder' first.
+ *
+ * @param {Array.<Doclet>} data
+ * @returns {Array.<Doclet>} data
```
ma2ciek: The `data` is unnecessary.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+ */
+module.exports = data => {
```
ma2ciek: This function is overcomplicated, it contains many nested functions, scopes and some shadow variables, e.g. `newDoclets`.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+	const dotIndex = parentDoclet.longname.lastIndexOf( '.' );
+	const hashIndex = parentDoclet.longname.lastIndexOf( '#' );
+	const name = parentDoclet.longname.slice( Math.max( dotIndex, hashIndex ) );
+
+	return childDoclet.longname + name;
+}
+
+function checkExplicitlyInherits( doclets ) {
+	for ( const doclet of doclets ) {
+		if ( typeof doclet.inheritdoc !== 'undefined' || typeof doclet.override !== 'undefined' ) {
```
ma2ciek: you can just check with `===`

packages/jsdoc-plugins/lib/relation-fixer/index.js
```diff
@@ -5,14 +5,11 @@
 
 'use strict';
 
-/**
- * @see http://usejsdoc.org/about-plugins.html
- */
-
-const augmentStatics = require( './augmentstatics' );
+const relationBuilder = require( './relationbuilder' );
```
ma2ciek: This should be a verb.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: Unnamed functions at the top level are hard to debug. You lose the stack trace.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+		return d.kind === 'class' || d.kind === 'interface' || d.kind === 'mixin';
+	} );
+
+	// Creates new doclets which given childDoclet is missing.
+	function getNewDoclets( childDoclet, relation, options ) {
+		const docletsToAdd = getDocletsToAdd( childDoclet, relation, options );
+		const newDoclets = [];
+
+		docletsToAdd.forEach( d => {
+			const cloned = cloneDeep( d );
```
ma2ciek: `clonedDoclet` would be easier to understand later.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+	const entitiesWhichNeedNewDoclets = data.filter( d => {
+		return d.kind === 'class' || d.kind === 'interface' || d.kind === 'mixin';
+	} );
+
+	// Creates new doclets which given childDoclet is missing.
+	function getNewDoclets( childDoclet, relation, options ) {
+		const docletsToAdd = getDocletsToAdd( childDoclet, relation, options );
+		const newDoclets = [];
+
+		docletsToAdd.forEach( d => {
```
ma2ciek: The `for of` function is more relevant to functions with some side-effects.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+			scope: 'static'
+		} ) );
+	} );
+
+	data.push( ...newDoclets );
+
+	return data;
+};
+
+function getNewLongname( parentDoclet, childDoclet ) {
```
ma2ciek: It's not enough descriptive.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+		newDoclets.push( ...getNewDoclets( d, 'implementsNested', {
+			scope: 'static'
+		} ) );
+		// Add missing static members from extended classes.
+		newDoclets.push( ...getNewDoclets( d, 'augmentsNested', {
+			scope: 'static'
+		} ) );
+	} );
+
+	data.push( ...newDoclets );
```
ma2ciek: Modifying the original input is not the best way to do it.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: The third issue is, that the whole file should be easy to read from the top to the bottom. And in your code functions execute from the bottom to the top.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+	} );
+
+	// Creates new doclets which given childDoclet is missing.
+	function getNewDoclets( childDoclet, relation, options ) {
+		const docletsToAdd = getDocletsToAdd( childDoclet, relation, options );
+		const newDoclets = [];
+
+		docletsToAdd.forEach( d => {
+			const cloned = cloneDeep( d );
+			let docletsWithSameLongname = [];
```
ma2ciek: This variable is overridden later.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+
+function getNewLongname( parentDoclet, childDoclet ) {
+	const dotIndex = parentDoclet.longname.lastIndexOf( '.' );
+	const hashIndex = parentDoclet.longname.lastIndexOf( '#' );
+	const name = parentDoclet.longname.slice( Math.max( dotIndex, hashIndex ) );
+
+	return childDoclet.longname + name;
+}
+
+function checkExplicitlyInherits( doclets ) {
```
ma2ciek: `checkIfExplicitlyInherits`?

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: `data` is not descriptive. The parameter should be named `doclets` or `originalDoclets`.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+
+	// Gets doclets from entities related to current doclet ( e.g. implemented by it ) and matching criteria given in options.
+	function getDocletsToAdd( doclet, relation, options = {} ) {
+		if ( !isNonEmptyArray( doclet[ relation ] ) ) {
+			return [];
+		}
+
+		const parents = doclet[ relation ];
+		const docletsToAdd = [];
+		parents.forEach( ln => {
```
ma2ciek: What is the `ln`?

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+			const toAdd = data.filter( d => {
+				let optionalCondition = true;
```
ma2ciek: `optionalCondition` is not descriptive.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: I would use the `for of` loop or `Array.prototype.reduce`.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: Or if you want to modify it (I saw some code, that also affects the `data`), you shouldn't return anything.

packages/jsdoc-plugins/lib/relation-fixer/index.js
```diff
@@ -5,14 +5,11 @@
...
+const addMissingDoclets = require( './addmissingdoclets' );
 
 exports.handlers = {
 	processingComplete( e ) {
-		augmentStatics( e.doclets );
+		e.doclets = addMissingDoclets( relationBuilder( e.doclets ) );
```
ma2ciek: https://github.com/ckeditor/ckeditor5-dev/pull/227/files#r121330658

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+				}
+			} );
+		}
+
+		return result;
+	}
+
+	entitiesWhichNeedNewDoclets.forEach( d => {
+		// Add missing static members from interfaces.
+		newDoclets.push( ...getNewDoclets( d, 'implementsNested', {
```
ma2ciek: As far as I understand, it should be imported from `packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js`

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
...
+			scope: 'static'
+		} ) );
+		// Add missing static members from extended classes.
+		newDoclets.push( ...getNewDoclets( d, 'augmentsNested', {
```
ma2ciek: same here https://github.com/ckeditor/ckeditor5-dev/pull/227/files#r121331270

packages/jsdoc-plugins/lib/relation-fixer/relationbuilder.js
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Checks ascendants of every doclet and adds them to relation chain.
+ * Handles nested inheritance, mixins and implementation of interfaces.
+ * Also adds descendants to doclet.
+ * For example: If ClassB extends ClassA and ClassA implements InterfaceC,
+ * ClassB and ClassA will have a property 'implementsNested': [ 'InterfaceC' ],
+ * also InterfaceC will have a property 'descendants': [ 'ClassA', 'ClassB' ] etc.
+ *
+ * @param {Array.<Doclet>} data
+ * @returns {Array.<Doclet>} data
```
ma2ciek: Same as in the previous file.

packages/jsdoc-plugins/lib/relation-fixer/relationbuilder.js
```diff
@@ -0,0 +1,96 @@
...
+ */
+module.exports = data => {
```
ma2ciek: This function is overcomplicated too.

packages/jsdoc-plugins/lib/relation-fixer/relationbuilder.js
```diff
@@ -0,0 +1,96 @@
...
+							result[ key ].push( ...ancestorsResult[ key ] );
+						}
+					} );
+				} );
+			}
+		} );
+
+		Object.assign( doclet, result );
+
+		return result;
```
ma2ciek: You shouldn't return here anything neither.

packages/jsdoc-plugins/lib/relation-fixer/relationbuilder.js
```diff
@@ -0,0 +1,96 @@
...
+		const hasAscendants = isNonEmptyArray( item.augments ) ||
+			isNonEmptyArray( item.implements ) ||
+			isNonEmptyArray( item.mixes );
+
+		return isOfProperKind && hasAscendants;
+	} );
+
+	// Creates arrays of items which are implemented, mixed, augmented by a doclet.
+	// Creates an array of doclet descendants.
+	function handleChain( doclet, relations = [], options = {} ) {
```
ma2ciek: What does it really mean?

ma2ciek: `relations` parameter doesn't need to have default value.

packages/jsdoc-plugins/lib/relation-fixer/relationbuilder.js
```diff
@@ -0,0 +1,96 @@
...
+				doclet[ r ].forEach( ln => {
+					const items = data.filter( d => d.longname === ln );
+
+					items.forEach( item => {
+						const ancestorsResult = handleChain( item, relations, {
+							descendants
+						} );
+
+						// Push relation chain of doclet's ancestors to current doclet.
+						for ( const key of Object.keys( result ) ) {
```
ma2ciek: `for in`?

packages/jsdoc-plugins/lib/relation-fixer/utils/isnonemptyarray.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @license Copyright (c) 2017, CKSource - Frederico Knabben. All rights reserved.
+ * Licensed under the terms of the MIT License (see LICENSE.md).
+ */
+
+'use strict';
+
+module.exports = obj => {
```
ma2ciek: Also I would use a normal and named function.

packages/jsdoc-plugins/tests/relation-fixer/relationfixerintegration.js
```diff
@@ -0,0 +1,87 @@
...
+/* jshint mocha:true */
+
+'use strict';
+
+const chai = require( 'chai' );
+const expect = chai.expect;
+const relationBuilder = require( '../../lib/relation-fixer/relationbuilder' );
+const addMissingDoclets = require( '../../lib/relation-fixer/addmissingdoclets' );
+const cloneDeep = require( 'lodash' ).cloneDeep;
+const testDoclets = [
```
ma2ciek: This code + `cloneDeep` should be added to the `beforeEach` hook

packages/jsdoc-plugins/tests/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,82 @@
...
+'use strict';
+
+const chai = require( 'chai' );
+const expect = chai.expect;
+const addMissingDoclets = require( '../../lib/relation-fixer/addmissingdoclets' );
+const cloneDeep = require( 'lodash' ).cloneDeep;
+
+describe( 'Plugin adds missing doclets from relation chain', () => {
+	it( 'should add missing doclets coming from interfaces', () => {
+		const testDoclets = require( './test-data/interface' );
```
ma2ciek: This is really danger as the second require will contain modified data!

ma2ciek: This is why the functions with side-effects or modifying the input are so hard to debug.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
Reinmar: Agree. It's unclear what's "new". In case of problems with describing it clearly you can also write a short doc what it does.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -0,0 +1,145 @@
+/**
+ * @license Copyright (c) 2017, CKSource - Frederico Knabben. All rights reserved.
+ * Licensed under the terms of the MIT License (see LICENSE.md).
+ */
+
+'use strict';
+
+const isNonEmptyArray = require( './utils/isnonemptyarray' );
```
Reinmar: I have mixed feelings regarding implementing special modules for such basic functions. Fixing duplicated code is easier than fixing broken abstraction. It's simply cheaper at the beginning to have `isNonEmptyArray` defined in two places where it's used and it's cheaper later on when it turns out that after some changes it's used in one or zero places :P.

packages/jsdoc-plugins/tests/relation-fixer/relationfixerintegration.js
```diff
@@ -0,0 +1,87 @@
+/**
+ * @license Copyright (c) 2016-2017, CKSource - Frederico Knabben. All rights reserved.
+ * Licensed under the terms of the MIT License (see LICENSE.md).
+ */
+
+/* jshint mocha:true */
```
Reinmar: We don't use `jshint` anymore.

pomek:
> Tests pass. One unrelated test failed due to a timeout.

@m-turek, could you merge `master` into this branch? I fixed the error with the timeout (#232).

m-turek:
I've commited fixes. Please have a look again.

Reinmar:
I'm afraid it doesn't work. I rebuilt the docs using this branch and they are not complete.

---

These are properties of `Plugin`:

image

---

And these are properties of `Bold` which should inherit everything from `Plugin`:

image

---

Other things are incomplete too:

image

## Master

To compare with master, the only thing I noticed that was missing on master were these properties (`Plugin`):

image

Reinmar:
My wild guess is that the `Bold` class and other concrete plugin classes define the `get pluginName()` and `get requires()` and these empty doc strings take precedence (which doesn't happen in `Plugin` itself which doesn't define those getters). However, `Bold` uses `@inheritDoc`:

```js
	/**
	 * @inheritDoc
	 */
	static get requires() {
		return [ BoldEngine ];
	}

	/**
	 * @inheritDoc
	 */
	static get pluginName() {
		return 'basic-styles/bold';
	}
```

So it should work fine.

Reinmar:
The same may be true for the empty `init()` and `afterInit()`.

Reinmar:
BTW, was this PR supposed to fix inheriting events? If so, then it didn't do that.

m-turek:
I've added a fix, reported errors should be fixed now. Also most validation errors should be gone.
There are still 4 validation errors now but I think they are real errors which were not discovered previously.

```Incorrect param type: Array.<(module:engine/view/document~MutatatedText|module:engine/view/document~MutatatedChildren)>
	at /Users/maciejturek/cke5-workspace/ckeditor5/packages/ckeditor5-typing/src/input.js (line 92)

Incorrect param type: Array.<(module:engine/view/observer/mutationobserver~MutatedText|module:engine/view/observer/mutationobserver~MutatatedChildren)>
	at /Users/maciejturek/cke5-workspace/ckeditor5/packages/ckeditor5-typing/src/input.js (line 133)

Incorrect param type: Iterable.<module:engine/model/delta~Delta>
	at /Users/maciejturek/cke5-workspace/ckeditor5/packages/ckeditor5-engine/src/model/range.js (line 408)

Incorrect event name: module:typing/deletecommand~DeleteCommand#event:execute in @fires tag
	at /Users/maciejturek/cke5-workspace/ckeditor5/packages/ckeditor5-typing/src/deletecommand.js (line 60)
```

Other things to keep in mind:
* Mixins are still not handled, the plugin doesn't create any new doclets for mixed members (except for events) - but I wanted to push something working before the weekend

Reinmar:
Interesting that those errors poped out now. It was clean before (except that event errors). It means that perhaps some docs were not previously validated so that would be good :) I'll check it.

Reinmar:
> Mixins are still not handled, the plugin doesn't create any new doclets for mixed members (except for events) - but I wanted to push something working before the weekend

OK, what I noticed is that events are differently inherited than other props/methods from mixins. E.g. in a `LinkEngine` plugin you now have:

image

So, the event is mixed from `ObservableMixin` but methods of this mixin are not. I reported recently that we need a different filter/label for "mixed" stuff and mixed events should behave like mixed properties there – they should only be displayed when a special filter is enabled. But of course they need to be always available during validation.

---

Another issue is that in `Plugin` you now have only these methods:

image

When there should be all methods of `PluginInterface`:

image

---

I can see in `Bold`:

image

and `BoldEngine` these properties:

image

Again, `BoldEngine` should have all properties of `Plugin` and `Plugin` should have all properties of `PluginInterface`. But what we get here is that only the properties which were really defined in a certain class are available.

---

To sum up, it should all work like normal inheriting:

* a child class has all properties method and events which its parent class and all mixed objects have,
* the description of child class's items should either be taken:
  * that class itself, if defined,
  * from the parent class/mixins if not defined in the child class.

Of course, this is how I imagine this working functionally, but I know nothing about JSDoc. I do start have a feeling that things are not so rosy with JSDoc. I guess we should talk about this after the weekend.

Reinmar:
I checked all the errors that you listed in https://github.com/ckeditor/ckeditor5-dev/pull/227#issuecomment-308507936 and they weren't previously caught by the validator. So the changes that you made made the validator more thorough.

Reinmar:
> Again I wasn't sure if we wanted all the mixed stuff so I added only events for now. I will add everything then. Sorry for the confusion

Please keep in mind that it will require implementing the "mixes" filter in Umberto. Otherwise, before we have it we'll make the API docs hard to browse because we use rich mixins in many classes.

So, it's even fine that only events are mixed for a while – there are not many mixed events so we don't need the filter yet.

m-turek:
Added missing stuff from interfaces and mixins (Umberto has mixins filter now).

Reinmar:
🎉 Congrats. It works great now and seems to be 100% complete. I think that it's very useful how it's rendered in Umberto now. E.g., here you have some command's methods:

image

You can clearly see which one is implemented directly in it and which are inherited. I can also show mixed ones:

image

And we can also see that overriding item descriptions work fine:

image

The `value` is defined in this command and has its own description, even though it was also defined in the parent class.

Great job!

packages/jsdoc-plugins/lib/validator/doclet-validator.js
```diff
@@ -316,6 +312,9 @@ class DocletValidator {
 	 * @returns {Boolean}
 	 */
 	_isCorrectType( type ) {
+		// JSDoc converts `Type.<Function>` to `Type.<function()>` for some reason...
+		type = type.replace( 'function()', 'function' );
```
Reinmar: I'm not sure about this line. I noticed that in some places we had a strange `function()` param... E.g:

image

The documentation of this method looks like this:

```
	/**
	 * Loads a set of plugins and adds them to the collection.
	 *
	 * @param {Array.<Function|String>} plugins An array of {@link module:core/plugin~PluginInterface plugin constructors}
	 * or {@link module:core/plugin~PluginInterface.pluginName plugin names}. The second option (names) work only if
	 * `availablePlugins` were passed to the {@link #constructor}.
	 * @param {Array.<String|Function>} [removePlugins] Names of plugins or plugin constructors
	 * which should not be loaded (despite being specified in the `plugins` array).
	 * @returns {Promise} A promise which gets resolved once all plugins are loaded and available into the
	 * collection.
	 * @returns {Promise.<Array.<module:core/plugin~PluginInterface>>} returns.loadedPlugins The array of loaded plugins.
	 */
```

And I'd like there the `Function` type, not `function()` nor `function`.

m-turek: JSDoc converts `Function` param to `function()` on its own. I don't know why at the moment.
`type = type.replace( 'function()', 'function' );` - this is to avoid errors thrown by `doclet-validator`

Reinmar: OK... so let's leave it like this and get back to it in a separate ticket.

PR #251, opened 2017-08-10
Add Window type to JSDoc validator

### LINKED ISSUE #250: JSDoc validator: Add Window as a valid type
ATM there's no way to document `@param {Window}` because it throws `Incorrect param type: Window`. It should be added [here](https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/jsdoc-plugins/lib/validator/types.js).

A follow-up of https://github.com/ckeditor/ckeditor5-utils/pull/174.

ma2ciek:

Internal: Added Window type to JSDoc validator. Closes #250.

### DIFF
--- a/packages/jsdoc-plugins/lib/validator/types.js
+++ b/packages/jsdoc-plugins/lib/validator/types.js
@@ -35,6 +35,7 @@ const BASIC_TYPES = [
 	'ClientRect',
 
 	// Web APIs
+	'Window',
 	'File',
 	'DataTransfer'
 ];


packages/jsdoc-plugins/lib/validator/types.js
```diff
@@ -35,6 +35,7 @@ const BASIC_TYPES = [
 	'ClientRect',
 
 	// Web APIs
```
Reinmar: Ekhm... shouldn't it belong to the previous category?

ma2ciek: Hmm, but that's something more than just DOM. I should create another category or write it at the top level.

Reinmar: Window is part of DOM: https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model

ma2ciek: I would say that it implements some DOM interfaces, but not belongs to, as it contains huge amount of other things

PR #274, opened 2017-09-06
T/269: Autolinking GitHub repositories in the commit message

### LINKED ISSUE #269: The changelog generator incorrectly turns @ckeditor/ckeditor5-foo into links to profiles
It generated this:

```
* Introduced `AttributeCommand`. It was moved from `[@ckeditor](https://github.com/ckeditor)/ckeditor5-core/src/command/toggleattributecommand`. Closes [#47](https://github.com/ckeditor/ckeditor5-basic-styles/issues/47). ([0301d4a](https://github.com/ckeditor/ckeditor5-basic-styles/commit/0301d4a))
```

pomek:

Fix: If a part of the comment matches to "organization/repository#id", this part will be replaced with a link which leads to the issue in the specified repository. Closes #269.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
@@ -45,94 +45,47 @@ const transformCommitUtils = {
 	 * @returns {String}
 	 */
 	linkToGithubUser( comment ) {
-		return comment.replace( /@([0-9A-Z_-]+\/?)/ig, ( matchText, nickName ) => {
-			if ( nickName.endsWith( '/' ) ) {
-				return matchText;
+		return comment.replace( /(.?)@([0-9A-Z_-]+)(\/)?/ig, ( matchedText, charBeforeAt, nickName, charAfterUser ) => {
+			// Most probably the matched value is an email address.
+			if ( charBeforeAt && /[A-Z0-9_]/i.test( charBeforeAt ) ) {
+				return matchedText;
 			}
 
-			return `[@${ nickName }](https://github.com/${ nickName })`;
+			if ( charAfterUser === '/' ) {
+				return matchedText;
+			}
+
+			charAfterUser = charAfterUser || '';
+
+			return `${ charBeforeAt }[@${ nickName }](https://github.com/${ nickName })${ charAfterUser }`;
 		} );
 	},
 
 	/**
 	 * Replaces reference to issue (#ID) with a link to the issue.
+	 * If comment matches to "organization/repository#ID", link will lead to the specified repository.
 	 *
 	 * @param {String} comment
 	 * @returns {String}
 	 */
 	linkToGithubIssue( comment ) {
-		const packageJson = getPackageJson();
-		const issuesUrl = ( typeof packageJson.bugs === 'object' ) ? packageJson.bugs.url : packageJson.bugs;
-
-		if ( !issuesUrl ) {
-			throw new Error( `The package.json for "${ packageJson.name }" must contain the "bugs" property.` );
-		}
-
-		return comment.replace( /(.?)#([0-9]+)/g, ( matchText, charBeforeIssue, issueId ) => {
-			// Don't replace anything if the '#ID' belongs to another part of the comment.
-			if ( charBeforeIssue && /[A-Z0-9_-]/i.test( charBeforeIssue ) ) {
-				return matchText;
-			}
-
-			return `${ charBeforeIssue }[#${ issueId }](${ issuesUrl }/${ issueId })`;
-		} );
-	},
-
-	/**
-	 * Replaces reference to repository (organization/repository) with a link to the repository.
-	 *
-	 * If the reference contains an issue, link will lead to the issue.
-	 *
-	 * @param {String} comment
-	 * @returns {String}
-	 */
-	linkToGithubRepository( comment ) {
-		return comment.replace( /(@|\.)?([A-Z0-9-_/]+)(#(\d+)?)?/gi, ( matchText, charBeforeRepo, repository, issueMark, issueId ) => {
-			if ( !isValidRepository() ) {
-				return matchText;
-			}
-
-			if ( issueId ) {
-				return `[${ repository }#${ issueId }](https://github.com/${ repository }/issues/${ issueId })`;
-			}
-
-			return `[${ repository }](https://github.com/${ repository })`;
-
-			function isValidRepository() {
-				// If the repository starts with '@', it means the package link should lead to NPM.
-				if ( matchText.startsWith( '@' ) ) {
-					return false;
+		return comment.replace( /(\/?[A-Z0-9-_]+\/[A-Z0-9-_]+)?#([0-9]+)/ig, ( matchedText, maybeRepository, issueId ) => {
+			if ( maybeRepository ) {
+				if ( maybeRepository.startsWith( '/' ) ) {
+					return matchedText;
 				}
 
-				// If the repository starts with '.', it means the repository is part of other link.
-				if ( matchText.startsWith( '.' ) ) {
-					return false;
-				}
-
-				// If the issue hash (#) occurs in repository but the issue id misses, don't modify it.
-				if ( issueMark && !issueId ) {
-					return false;
-				}
-
-				// If the repository contains more than two slashes, it means the repository is a path.
-				return repository.split( '/' ).length === 2;
-			}
-		} );
-	},
-
-	/**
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-dev-env/tests/release-tools/utils/transform-commit/transform-commit-utils.js
```diff
@@ -41,46 +41,170 @@ describe( 'dev-env/release-tools/utils/transform-commit', () => {
...
+				expect( transformCommit.linkToGithubRepository( '@ckeditor/ckeditor5-dev' ) )
+					.to.equal( '@ckeditor/ckeditor5-dev' );
+			} );
+
+			it( 'does not make a link from a comment which is a path', () => {
+				expect( transformCommit.linkToGithubRepository( 'i/am/a/path' ) )
+					.to.equal( 'i/am/a/path' );
+			} );
+
+			it( 'does not make a link if a comment does not match to "@organization/repository"', () => {
```
ma2ciek: `organization/repository`

ma2ciek: In that case, I would enable the link.

pomek: I would like to create a link only if the string matches in 100%, so `org/repo` is ok but `org/repo/` isn't ok. Maybe other people will help us with the decision (@Reinmar, I'm looking at you :D)

Reinmar:
Sorry, this branch has conflicts now.

packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
```diff
@@ -37,31 +37,100 @@ const transformCommitUtils = {
...
+	 * Replaces reference to user (@name) with a link to his profile.
 	 *
-	 * @param {String} sentence
+	 * @param {String} comment
 	 * @returns {String}
 	 */
-	linkGithubUsers( sentence ) {
-		return sentence.replace( /@([\w\d_-]+)/g, '[@$1](https://github.com/$1)' );
+	linkToGithubUser( comment ) {
+		return comment.replace( /@([0-9A-Z_-]+\/?)/ig, ( matchText, nickName ) => {
```
Reinmar: `matchedText`

Reinmar: `/@\w+(?!\/)/g` and you don't need that `if ()` below

packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
```diff
@@ -39,31 +39,100 @@ const transformCommitUtils = {
...
 
 		if ( !issuesUrl ) {
 			throw new Error( `The package.json for "${ packageJson.name }" must contain the "bugs" property.` );
 		}
 
-		return sentence.replace( /#([0-9]+)/g, ( _, issueId ) => {
-			return `[#${ issueId }](${ issuesUrl }/${ issueId })`;
+		return comment.replace( /(.?)#([0-9]+)/g, ( matchText, charBeforeIssue, issueId ) => {
+			// Don't replace anything if the '#ID' belongs to another part of the comment.
+			if ( charBeforeIssue && /[A-Z0-9_-]/i.test( charBeforeIssue ) ) {
```
Reinmar: You can put this regexp in the first one – just replace `(.?)` with it.

packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
```diff
@@ -39,31 +39,100 @@ const transformCommitUtils = {
...
+				return matchText;
+			}
+
+			if ( issueId ) {
+				return `[${ repository }#${ issueId }](https://github.com/${ repository }/issues/${ issueId })`;
+			}
+
+			return `[${ repository }](https://github.com/${ repository })`;
+
+			function isValidRepository() {
```
Reinmar: You can define this function outside this scope cause it's making the code harder to read.

packages/ckeditor5-dev-env/tests/release-tools/utils/transform-commit/transform-commit-utils.js
```diff
@@ -41,46 +41,170 @@ describe( 'dev-env/release-tools/utils/transform-commit', () => {
 			} );
 		} );
 
-		describe( 'linkGithubUsers()', () => {
-			it( 'marks profile as a link to GitHub profile', () => {
-				expect( transformCommit.linkGithubUsers( '@CKSource' ) )
+		describe( 'linkToGithubUser()', () => {
```
Reinmar: Just two tests? Be more creative and also more realistic – these tests check cases completely detached from reality. Those strings will be used in many contexts.

packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
```diff
@@ -39,31 +39,100 @@ const transformCommitUtils = {
...
+	},
+
+	/**
+	 * Replaces scoped package name with a link which will land to the package in NPM repository.
+	 *
+	 * @param {String} comment
+	 * @returns {String}
+	 */
+	linkToNpmScopedPackage( comment ) {
+		return comment.replace( /(@[A-Z0-9-_/]+)#?/gi, ( matchText, repository ) => {
```
Reinmar: Again `(?!#)`

Reinmar: And if you want to have exactly two segments separated with `/` just write this in the pattern.

Reinmar:
A general issue with all this replacing is that it will double link stuff which is already linked. E.g.

```
Big [thanks to @foo for contribution](http://...).
```

Will become:

```
Big [thanks to [@foo](https://github.com/foo) for contribution](http://...).
```

The same with this (which is actually how we format package names most of the time):

```
This change was published in the `@ckeditor/bam` package.
```

Will become:

```
This change was published in the `[@ckeditor/bam](https://...)` package.
```

Which is not going to work.

When making PR like this one you really need to take examples from the real life. Otherwise, you create an unrealistic implementation which doesn't really work.

What I'm getting at is that there are two options:

1. Make a list of commit messages and look for interesting examples. Use them as test cases. I can already tell you that there will be a lot of untrivial cases like the ones I described above and that you may sink in this ticket for a week, completely reimplementing these methods.
2. Take a short path and do the minimum. Fix the mentioned bugs and report the rest as followups (you did more than fixing bugs in this PR).

packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
pomek: It does not help. I leave it as it is.

packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
pomek: As above, it does not help. I had to leave it.

packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
pomek: This function has been removed.

pomek:
> A general issue with all this replacing is that it will double link stuff which is already linked.

Unfortunately, if we would like to fix these issues with replacing already existing links, the whole mechanism has to be rewritten. It can take a lot of time…

I removed part of the code which replaces `org/repo` with a link to the repository (includes the `@org/repo` with a link to npm). Now we will replace `org/repo#id` with a link to the issue in the specified repository. 

I also improved a little the tests.

Could you look on this once again?

packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
```diff
@@ -39,30 +39,52 @@ const transformCommitUtils = {
 	},
 
 	/**
-	 * Changes user's name to link that leads to the user's profile.
+	 * Replaces reference to user (@name) with a link to his profile.
```
Reinmar: Why "his"?

PR #279, opened 2017-09-11
T/278: Dependency Checker

### LINKED ISSUE #278: Add "dependency-checker" before running tests on CI
Would be nice to have a tool which could checks whether all required dependencies (and dev-dependencies) are specified in `package.json`.

pomek:

Feature: Adds "dependency-checker" which allows validating whether required dependencies are specified in package.json. Closes #278.

---

<details>
    <summary>I run this tool for our packages. The result is below:</summary>

`editor-inline`

Error: The files listed below contain dependencies which are not defined in "package.json":

  - ./tests/manual/tickets/23/1.js

Missing dev-dependencies:

  - @ckeditor/ckeditor5-enter
  - @ckeditor/ckeditor5-typing
  - @ckeditor/ckeditor5-heading
  - @ckeditor/ckeditor5-undo

---

`engine`

Error: The files listed below contain dependencies which are not defined in "package.json":

  - ./tests/manual/tickets/880/1.js
  - ./tests/manual/tickets/887/1.js

Missing dev-dependencies:

  - @ckeditor/ckeditor5-essentials

---

`image`

Error: The files listed below contain dependencies which are not defined in "package.json":

  - ./tests/manual/autohoist.js
  - ./tests/manual/caption.js
  - ./tests/manual/tickets/127/1.js

Missing dev-dependencies:

  - @ckeditor/ckeditor5-essentials
  - @ckeditor/ckeditor5-list
  - @ckeditor/ckeditor5-link

---

`list`

Error: The files listed below contain dependencies which are not defined in "package.json":

  - ./tests/manual/nestedlists.js

Missing dev-dependencies:

  - @ckeditor/ckeditor5-link
</details>

### DIFF
--- a/packages/ckeditor5-dev-tests/bin/check-dependencies.js
+++ b/packages/ckeditor5-dev-tests/bin/check-dependencies.js
@@ -29,11 +29,16 @@ for ( const filePath of glob.sync( globPattern ) ) {
 		continue;
 	}
 
-	fs.readFileSync( filePath, 'utf-8' )
-		.split( '\n' )
-		.forEach( line => {
-			// Find required package name.
-			const matchedImport = /^import[^;]+from '(@ckeditor\/[^/]+)[^']+';/mg.exec( line );
+	const fileContent = fs.readFileSync( filePath, 'utf-8' );
+	const matchedImports = fileContent.match( /^import[^;]+from '(@ckeditor\/[^/]+)[^']+';/mg );
+
+	if ( !matchedImports ) {
+		continue;
+	}
+
+	matchedImports
+		.forEach( importLine => {
+			const matchedImport = importLine.match( /(@ckeditor\/[^/]+)/ );
 
 			if ( !matchedImport ) {
 				return;
--- a/packages/ckeditor5-dev-tests/bin/check-dependencies.js
+++ b/packages/ckeditor5-dev-tests/bin/check-dependencies.js
@@ -1,0 +1,88 @@
+#!/usr/bin/env node
+
+/**
+ * @license Copyright (c) 2003-2017, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+const fs = require( 'fs' );
+const path = require( 'path' );
+const glob = require( 'glob' );
+
+const cwd = process.cwd();
+const packageJson = require( path.join( cwd, 'package.json' ) );
+
+const currentPackageName = packageJson.name;
+const dependencies = Object.keys( packageJson.dependencies || {} );
+const devDependencies = Object.keys( packageJson.devDependencies || {} );
+
+const missingDependencies = new Set();
+const missingDevDependencies = new Set();
+const invalidFiles = new Set();
+
+const globPattern = path.join( cwd, '@(src|tests)/**/*.js' );
+
+for ( const filePath of glob.sync( globPattern ) ) {
+	fs.readFileSync( filePath, 'utf-8' )
+		.split( '\n' )
+		.forEach( line => {
+			// Find required package name.
+			let packageName = /^import .* from '(@ckeditor\/[^/]+)\/.*/.exec( line );
+
+			if ( !packageName ) {
+				return;
+			}
+
+			packageName = packageName[ 1 ];
+
+			// Current package cannot be added as dependency.
+			if ( currentPackageName === packageName ) {
+				return;
+			}
+
+			// Check whether the package is defined as dependency and dev-dependency.
+			const containPackageAsDependency = dependencies.includes( packageName );
+			const containPackageAsDevDependency = devDependencies.includes( packageName );
+
+			// If so, current checking line is fine.
+			if ( containPackageAsDependency || containPackageAsDevDependency ) {
+				return;
+			}
+
+			// If found package is missing and current file is "source" file, add the package as a missing dependency.
+			if ( !containPackageAsDependency && filePath.match( /\/src\// ) ) {
+				missingDependencies.add( packageName );
+			}
+
+			// If found package is missing and current file is "test" file, add the package as a missing dev-dependency.
+			if ( !containPackageAsDevDependency && filePath.match( /\/tests\// ) ) {
+				missingDevDependencies.add( packageName );
+			}
+
+			// Mark current file as invalid (it can help for manual debugging).
+			invalidFiles.add( filePath.replace( cwd + '/', './' ) );
+		} );
+}
+
+if ( invalidFiles.size ) {
+	let errorMessage = 'The files listed below contain dependencies which are not defined in "package.json":\n\n';
+	errorMessage += [ ...invalidFiles ].map( formatLine ).join( '\n' );
+
+	if ( missingDependencies.size ) {
+		errorMessage += '\n\nMissing dependencies:\n\n';
+		errorMessage += [ ...missingDependencies ].map( formatLine ).join( '\n' );
+	}
+
+	if ( missingDevDependencies.size ) {
+		errorMessage += '\n\nMissing dev-dependencies:\n\n';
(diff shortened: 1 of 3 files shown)


pomek:
The following branches should be together with this PR:

- https://github.com/ckeditor/ckeditor5-engine/compare/t/ckeditor5-dev/278
- https://github.com/ckeditor/ckeditor5-image/compare/t/ckeditor5-dev/278
- https://github.com/ckeditor/ckeditor5-list/compare/t/ckeditor5-dev/278
- https://github.com/ckeditor/ckeditor5-editor-inline/compare/t/ckeditor5-dev/278

Reinmar:
Haha :) I see you implemented it from scratch. I hope it's stable because that's actually fine to have  it like this. https://www.npmjs.com/package/npm-check has quite a lot of deps and I don't think that https://www.npmjs.com/package/check-dependencies supports ES6 imports.

packages/ckeditor5-dev-tests/bin/check-dependencies.js
```diff
@@ -0,0 +1,90 @@
...
+const invalidFiles = new Set();
+
+const globPattern = path.join( cwd, '@(src|tests)/**/*.js' );
+
+for ( const filePath of glob.sync( globPattern ) ) {
+	fs.readFileSync( filePath, 'utf-8' )
+		.split( '\n' )
+		.forEach( line => {
+			// Find required package name.
+			let packageName = /^import .* from '(@ckeditor\/[^/]+)\/.*/.exec( line );
```
Reinmar: This is not a package name here. It's a match for import. If you'd use `const` you wouldn't make this mistake ;)

Reinmar: Also, use `[^ ]+` instead of `.*`.

packages/ckeditor5-dev-tests/bin/check-dependencies.js
```diff
@@ -0,0 +1,90 @@
...
+
+const currentPackageName = packageJson.name;
+const dependencies = Object.keys( packageJson.dependencies || {} );
+const devDependencies = Object.keys( packageJson.devDependencies || {} );
+
+const missingDependencies = new Set();
+const missingDevDependencies = new Set();
+const invalidFiles = new Set();
+
+const globPattern = path.join( cwd, '@(src|tests)/**/*.js' );
```
Reinmar: This will also check `src/lib` which it should not (we keep there a lodash build).

packages/ckeditor5-dev-tests/bin/check-dependencies.js
```diff
@@ -0,0 +1,90 @@
...
+			let packageName = /^import .* from '(@ckeditor\/[^/]+)\/.*/.exec( line );
+
+			if ( !packageName ) {
+				return;
+			}
+
+			packageName = packageName[ 1 ];
+
+			// Current package cannot be added as dependency.
+			if ( currentPackageName === packageName ) {
```
Reinmar: You actually shouldn't have a non-relative import within a package itself. So this is a bug as well. I'm too lazy to look for bugs caused by this now, but there were a couple. https://github.com/ckeditor/ckeditor5/issues/437 mentions this.

Reinmar:
In general it looks really cool. Just minor issues to polish and we'll be able to close two issues at once.

packages/ckeditor5-dev-tests/bin/check-dependencies.js
pomek: It will fail on imports a few modules (`import { getData, setData } from ...`).

Reinmar: Right. I've completely forgotten about this case.

Reinmar: So at least make it `.+` which is more correct :D

Besides, I've just thought about multi-line imports:

```js
import {
	stringify as stringifyModel,
	setData as setModelData,
	getData as getModelData
} from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
```

Which means that you can't split the lines like that.

This should work though:

```js
/^import[^;]+from '(@ckeditor\/[^/]+)[^']+';/mg/
```

E.g.:

```js
// fs.readFileSync( 'packages/ckeditor5-clipboard/tests/clipboard.js' ).toString().match( /^import[^;]+from '(@ckeditor\/[^/]+)[^']+';/mg );

[ 'import VirtualTestEditor from \'@ckeditor/ckeditor5-core/tests/_utils/virtualtesteditor\';',
  'import Paragraph from \'@ckeditor/ckeditor5-paragraph/src/paragraph\';',
  'import {\n\tstringify as stringifyView,\n\tparse as parseView\n} from \'@ckeditor/ckeditor5-engine/src/dev-utils/view\';',
  'import {\n\tstringify as stringifyModel,\n\tsetData as setModelData,\n\tgetData as getModelData\n} from \'@ckeditor/ckeditor5-engine/src/dev-utils/model\';',
  'import ViewDocumentFragment from \'@ckeditor/ckeditor5-engine/src/view/documentfragment\';',
  'import ViewText from \'@ckeditor/ckeditor5-engine/src/view/text\';' ]
```

pomek:
@Reinmar, could you look on this one more time?

packages/ckeditor5-dev-tests/bin/check-dependencies.js
pomek: I have no idea how to write a glob pattern which:

- skips `src/lib/**/*.js`,
- accepts `src/file.js` and `src/dir/file.js`.

I had a problem with `src/file.js` and I decided to check the path in the `for...of` loop.

Reinmar:
R- again. You still split content by lines before matching it.

Reinmar:
Matching the content of the entire file with the regexp I gave you returns all `import` statements in the file at once.

pomek:
(?).

May I'm doing something wrong but the regexp does not work:

image

```js
for ( const filePath of glob.sync( globPattern ) ) {
	// Skip "src/lib" directory.
	if ( filePath.match( /\/src\/lib\// ) ) {
		continue;
	}

	const fileContent = fs.readFileSync( filePath, 'utf-8' );
	const matchedImport = /^import[^;]+from '(@ckeditor\/[^/]+)[^']+';/mg.exec( fileContent );

	console.log( filePath );
	console.log( matchedImport );

	break;

	/* ... */
}
```

Tested with: https://github.com/ckeditor/ckeditor5-core/blob/599681ab8218cc7944b30f2a9f666a9112adacf5/src/command.js.

Reinmar:
> May I'm doing something wrong but the regexp does not work:

Yes, you changed `match()` to `exec()` ;)

pomek:
:D

pomek:
Now it should look ok.

I don't remove the `R-` label because I need to fix invalid imports in `core`, `ui` and `utils`. These errors occur after introducing changes after review.

pomek:
I've just updated the comment with required changes (https://github.com/ckeditor/ckeditor5-dev/pull/279#issuecomment-328765053).

Ready to review.

packages/ckeditor5-dev-tests/bin/check-dependencies.js
Reinmar: Where's this condition now?

pomek: I removed it. If a package requires some files from itself, it should use a relative path. I fixed the paths on `t/ckeditor5-dev/278` branches in several repositories.

Reinmar: So please bring it back and make the tool report them in a clear (I guess they are reported now as pkg x is missing in pkg x's deps). This is a bit different case and could be reported separately.

pomek: https://github.com/ckeditor/ckeditor5-dev/issues/282.

PR #281, opened 2017-09-18
T/272: Changed order of calling commands (release tool)

### LINKED ISSUE #272: The release tool should not mix committing with pushing and with publishing on GH and npm
The process should look like this:

* first do all commits,
* then do all pushes,
* then publish all on npm,
* then publish all released on GH.

Right now, the tool mixes the first two steps together (commit+push) and it also mixes publishing which is unsafe because for 5 minutes (according to https://gist.github.com/Reinmar/8ae3db219d385df06ccb94cd6d17cfd6) we have an incomplete release on npm. This also breaks builds CI because pushes to the repo start builds which try to install packages which are not yet installed on npm. But it's hard to prevent that.

### LINKED ISSUE #292: Commits with deps update should be skipped by CI
"Internal: Updated dependencies." committed by `gulp release:dependencies`  cause CI to fail: https://travis-ci.org/ckeditor/ckeditor5-core/builds/282705735

Right after that we commit the new version (see e.g. https://travis-ci.org/ckeditor/ckeditor5-core/builds/282707730) and those builds pass, so let's just mark the first commit as `[skip ci]`.

pomek:

Fix: Changed order of commands in release tool. Closes #272. Closes #292.

Due to releasing packages one after another, the CI could break the builds. Now release tool will:

* do all commits (generate missing changelogs or/and  update dependencies' versions),
* publish all packages on NPM (all packages will contain proper versions),
* do all pushes (CI will not crash because all versions are valid),
* make the GitHub releases.

Function `validatePackageToRelease()` will not return an error if the branch is ahead with the remote.

---

### Additional information

I had to change the `validatePackageToRelease()` function because commit with updating dependencies cannot be pushed before releasing packages on npm.


packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
```diff
@@ -208,28 +220,79 @@ module.exports = function releaseSubRepositories( options ) {
...
-	function releasePackages( releaseOptions ) {
+	function pullAndPushPackages() {
+		return executeOnPackages( pathsCollection.packages, repositoryPath => {
+			process.chdir( repositoryPath );
+
+			const packageJson = getPackageJson( repositoryPath );
+			log.info( `Synchronizing a local repository with the remote for "${ packageJson.name }"...` );
+
+			// Push updated dependencies and changelog if was generated.
+			exec( 'git pull && git push' );
```
Reinmar: We can't pull after tagging a new version. That would break the tag.

packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
```diff
@@ -208,28 +220,79 @@ module.exports = function releaseSubRepositories( options ) {
 		} );
 	}
 
-	function releasePackages( releaseOptions ) {
+	function pullAndPushPackages() {
```
Reinmar: Why is this function located before `bumpVersion()`? Sort the code in the order your read it.

packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
```diff
@@ -7,14 +7,15 @@
...
+const parseGithubUrl = require( 'parse-github-url' );
 const cli = require( '../utils/cli' );
+const createGithubRelease = require( '../utils/creategithubrelease' );
 const displaySkippedPackages = require( '../utils/displayskippedpackages' );
 const executeOnPackages = require( '../utils/executeonpackages' );
 const generateChangelogForSinglePackage = require( './generatechangelogforsinglepackage' );
 const getPackageJson = require( '../utils/getpackagejson' );
 const getPackagesToRelease = require( '../utils/getpackagestorelease' );
 const getSubRepositoriesPaths = require( '../utils/getsubrepositoriespaths' );
-const releaseRepository = require( '../utils/releaserepository' );
```
Reinmar: If the new code in this file does all the releasing why haven't you removed the `releaseRepository` file?

packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
pomek: Sure. Pull should be done before tagging.

packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
pomek: We still have a task for releasing the single repository which uses this util.

pomek:
@Reinmar, `pull` will be now done before the tagging.

Could you review once again?

packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
Reinmar: Those commands should not try to pull anything. The assumption is that no one will do any changes during the process. In other words – let's assume that the releaser pulled everything, (optionally) validate that there are no new changes before committing anything, and then just push.

pomek: I removed pulling and didn't introduce your optionally validation step.

Let's assume that everything is pulled before starting release.

Reinmar:
All worked great. I commented out `npm publish` and `git push` calls and the rest went smooth!

PR #284, opened 2017-09-18
Provide error tag detection.

### LINKED ISSUE #280: Errors documentation
*(The mechanisms described in this issue were originally designed in https://github.com/ckeditor/ckeditor5-design/issues/14.)*

Since the beginning, we mark every error with `/** @error */` docstring containing additional information about the issue. E.g.:

```js
/**
 * There was a problem processing the configuration of the toolbar. The item with the given
 * name does not exist so it was omitted when rendering the toolbar.
 *
 * This warning usually shows up when the {@link module:core/plugin~Plugin} which is supposed
 * to provide a toolbar item has not been loaded or there is a typo in the configuration.
 *
 * Make sure the plugin responsible for this toolbar item is loaded and the toolbar configuration
 * is correct, e.g. {@link module:basic-styles/bold~Bold} is loaded for the `'bold'` toolbar item.
 *
 * @error toolbarview-item-unavailable
 * @param {String} name The name of the component.
 */
log.warn(
	'toolbarview-item-unavailable: The requested toolbar item is unavailable.',
	{ name }
);
```

Or:

```js
/**
 * Command does not exist.
 *
 * @error commandcollection-command-not-found
 * @param {String} commandName Name of the command.
 */
throw new CKEditorError( 'commandcollection-command-not-found: Command does not exist.', { commandName } );
```

This was meant to allow us to keep error names short (and even minify them, by stripping the text after `:`). However, the crucial point here is to have these errors listed (with full descriptions) in the API docs site. Also, when `CKEditorError` or `log.*` are used, they should log also a *working* link to that error.

This means writing some kind of plugin for JSDoc or Umberto and figuring out how these errors should be listed. On one page in some special subpage? In the modules in which they are defined? What if an error may be thrown by multiple modules? This we need to answer first.

Secondly, later on, we can think on the following improvements:

* minifying the error names (as mentioned above – by stripping the textual part and leaving error code only) – ticket: https://github.com/ckeditor/ckeditor5/issues/30,
* validation that all errors (`CKEditorError` and `log.*` usage) are documented.

ma2ciek:

Feature: Provided `@error` tag detection in JSDoc. Closes #280.

### DIFF
--- a/packages/ckeditor5-dev-docs/lib/index.js
+++ b/packages/ckeditor5-dev-docs/lib/index.js
@@ -44,6 +44,7 @@ function build( config ) {
 		plugins: [
 			require.resolve( 'jsdoc/plugins/markdown' ),
 			require.resolve( '@ckeditor/jsdoc-plugins/lib/export-fixer/export-fixer' ),
+			require.resolve( '@ckeditor/jsdoc-plugins/lib/custom-tags/error' ),
 			require.resolve( '@ckeditor/jsdoc-plugins/lib/relation-fixer' ),
 			require.resolve( '@ckeditor/jsdoc-plugins/lib/longname-fixer/longname-fixer' ),
 			require.resolve( '@ckeditor/jsdoc-plugins/lib/event-extender/event-extender' ),
--- a/packages/jsdoc-plugins/lib/custom-tags/error.js
+++ b/packages/jsdoc-plugins/lib/custom-tags/error.js
@@ -1,0 +1,27 @@
+/**
+ * @license Copyright (c) 2016-2017, CKSource - Frederico Knabben. All rights reserved.
+ * Licensed under the terms of the MIT License (see LICENSE.md).
+ */
+
+'use strict';
+
+module.exports = {
+	// See http://usejsdoc.org/about-plugins.html#tag-definition.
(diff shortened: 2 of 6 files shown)


ma2ciek:
This PR adds error doclets to the list of doclets:

Example:
```json
{
    "comment": "/**\n\t\t\t\t * This item's id should be a string.\n\t\t\t\t *\n\t\t\t\t * @error collection-add-invalid-id\n\t\t\t\t */",
    "meta": {
        "range": [
            3270,
            3370
        ],
        "filename": "collection.js",
        "lineno": 120,
        "path": "/Users/maciejbukowski/workspace/ckeditor5/packages/ckeditor5-utils/src",
        "code": {}
    },
    "description": "<p>This item's id should be a string.</p>",
    "name": "collection-add-invalid-id",
    "longname": "module:errors/collection-add-invalid-id",
    "kind": "error",
    "memberof": "module:errors",
    "scope": "inner"
}
```

m-turek:
Looks good to me.

ma2ciek:
It should be noted, that it's only the first step in creating documentation for errors. The display part should be done in the Umberto.

Reinmar:
Perfect! I checked that it works beautifully.

PR #291, opened 2017-10-03
T/289: Task for generating a summary changelog

### LINKED ISSUE #289: Introduce "summary changelog" generator for ckeditor5 and the builds
We need to automate creating this changelog: https://github.com/ckeditor/ckeditor5/blob/master/CHANGELOG.md

because similar changelogs are needed in the builds. The builds' changelogs should have the same format, but should have filtered packages to only contain those packages on which this build depends (the whole tree, not only the first level).

pomek:

Feature: Introduced a task which allows generating a summary changelog for a repository. Closes #289.

NOTE: Changelog will propose `internal` bump version instead of `skip` if a package contains any commit.
NOTE: If a package does not contain any commit, the changelog will propose `skip` bump version.
NOTE: `internal` bump will increase the release version if the current version is specified as a release (`v1.0.0-alpha.1` becomes `v1.0.0-alpha.2`).

BREAKING CHANGES: Release tool won't generate changelogs for skipped packages anymore. This task will be handled by changelog generator.

---

Added as a part of `changelog` task in the main repository: https://github.com/ckeditor/ckeditor5/compare/t/ckeditor5-dev/289


pomek:
@Reinmar, ready to review.

Reinmar:
> +Besides changes in the dependencies, this build also contains these bug fixes:

This message makes no sense in the main package.json. Let's change it to:

> +Besides changes in the dependencies, this version also contains these bug fixes:

Reinmar:
The builds' and main pkg's changelogs miss the `### Dependencies` header. You can see it being added in the builds' changelogs.

Reinmar:
Most importantly... I've run the command for the current state of the project and... I'll let you check what's wrong ;>. Hint – the list of updated deps is not complete.

pomek:
Hm. It will be a long post with a lot of images.

### Release everything

I released all dependencies. New version is equal to bump major version for previous package. Builds always have version `900.0.0`.

**ckeditor5**:

image

**ckeditor5-build-balloon**:

image

**ckeditor5-build-classic**:

image

**ckeditor5-build-inline**:

image

Because all dependencies have been released, all packages are mentioned in packages' changelog. IMO ok.

### Release `core` package only

I released only the `ckeditor5-core` package which is a base dependency for all plugins.

**ckeditor5**:

image

**ckeditor5-build-balloon**:

image

**ckeditor5-build-classic**:

image

**ckeditor5-build-inline**:

image

`ckeditor5` repository is ok because `core` is defined as deps in pkg.json.

`builds` repositories are ok because `core` is specified as a dependency for at least one dependency for builds' dependencies. 
Other words – builds require `ckeditor5-basic-styles` which requires `ckeditor5-core` so `core` package should be mentioned in CHANGELOG for builds.

Based on assumptions from ticket:

> The builds' changelogs should have the same format, but should have filtered packages to only contain those packages on which this build depends (the whole tree, not only the first level).

Because `basic-styles` won't be released, this package isn't mentioned in changelogs.

### Release `editor-*` packages

I released packages which are the base dependency for builds.

**ckeditor5**:

image

**ckeditor5-build-balloon**:

image

**ckeditor5-build-classic**:

image

**ckeditor5-build-inline**:

image

The main repository contains entries related to `editor-*` and `build-*`. It's ok because all these packages will be released.

Builds contain only entry related to the editor which is a base for a build (build-inline => editor-inline, etc).

---

Because pasted scenarios look good to me, I have a few questions:

- Am I miss anything?
- Do we misunderstand? 
- Could you help me reproduce a situation which allows me checking whether all dependencies are added to changelogs proper?

Reinmar:
When testing, you should use realistic scenario. The one I checked simply meant running the tool on the current CKE5 code base without changing any suggested new versions. It didn't work well.

pomek:
For the current state of the project:

image

image

Changelogs were generated using default version proposed by the tool.

**ckeditor5:**

image

**ckeditor5-build-balloon:**

image

**ckeditor5-build-classic:**

image

**ckeditor5-build-inline:**

image

IMO everything is ok except the version bumper for packages but this is a topic for another ticket.

Reinmar:
Will only the listed packages be release? No – also other packages which depend on them. *All* of them need to be listed there. This is task **1**.

Another problem which I spotted here is that the changelog generator skips packages which had only internal changes in them. That might've been our intentional decision in the past, but I think now that it's wrong.

I'd expect now one of the below:

* The changelog generator should propose the "internal" release for them and generate the changelog right away.
* The changelog generator should skip them but the release tool needs to release them anyway (just like it releases packages which depend on other released packages).

I don't actually know how the release tool work now – perhaps it does what the 2nd option says. But the fact that I'm confused here means that the 2nd option is bad – It leaves everything for the release process which is harder for me to control.

Therefore, I'd prefer if the changelog generator proposed "internal" for those packages. This is task **2** and seems unrelated to this ticket.

However, if we go with **2**, then it also makes a perfect sense if the changelog generator automatically generated empty release notes for packages which depend on other packages to be released. Right now it's the release tool which does this, but this, again, is hard for me to control. If we'll do this change, then the changelog generator will have no troubles to include all packages that are going to be released in "summary changelog" which will, ultimately, fix **1**.

So, in other words – changelog generator should:

* generate "internal release" positions in packages which have only internal/test/docs changes,
* generate "internal release" positions in packages which have no changes at all but which depend on other packages to be released,
* generate "summary changelogs" which will include all packages which are going to be released and on which the starting package depends (directly and indirectly).

pomek:
Ok, I finally finish fixing R-.

I introduced few changes:

* release tool **will not generate changelogs** for skipped packages anymore,
* changelog generator will check whether changelogs for skipped packages should be generated automatically (this task was executed by release tool in the previous version)
* changelog will propose `internal` bump version instead of `skip` if package contains any commits
* if package does not contain any commit, changelog will propose `skip` bump version,
* `internal` bump will increase the release version if the current version is specified as a release (`v1.0.0-alpha.1` becomes `v1.0.0-alpha.2`).

I will modify the commit message. I guess it's worth to mention about these changes

Reinmar:
It seems to work great, although, I tested it in ckeditor5 only.

The changelog generator prints out a bit confusing logs as it first goes through all the packages and lists skipped once just to process them later on, but that's okeysh. We can live with that.

PR #294, opened 2017-10-04
Provide support for `@observable` tag

### LINKED ISSUE #285: Add support for @observable
We use it in the API docs 88 times, so this is pretty important to have this supported.

We mark with this tag properties which changes fire events.

So, the minimum will be to mark them with API docs with "observable" keyword (just like "protected" and "private").

The optimum will be to automatically add events which accompany these properties. E.g. if `Bar#foo` is observable, then `Bar` should fire a `change:foo` event.

Sometimes, those events are already documented so don't override this by mistake. Custom docs have precedence.

ma2ciek:

Feature: Provided support for `@observable` tag. Closes #285.

### DIFF
--- a/packages/jsdoc-plugins/lib/custom-tags/observable.js
+++ b/packages/jsdoc-plugins/lib/custom-tags/observable.js
@@ -1,0 +1,28 @@
+/**
+ * @license Copyright (c) 2016-2017, CKSource - Frederico Knabben. All rights reserved.
+ * Licensed under the terms of the MIT License (see LICENSE.md).
+ */
+
+'use strict';
+
+const addMissingEventDoclets = require( '../observable-event-fixer/addmissingeventdoclets' );
+
+module.exports = {
+	// See http://usejsdoc.org/about-plugins.html#tag-definitions.
+	defineTags( dictionary ) {
+		dictionary.defineTag( 'observable', {
+			mustNotHaveDescription: false,
+			canHaveType: true,
+			canHaveName: true,
+			onTagged( doclet ) {
+				doclet.observable = true;
+			}
+		} );
+	},
+
+	handlers: {
+		processingComplete( e ) {
+			e.doclets = addMissingEventDoclets( e.doclets );
+		}
+	}
+};
--- a/packages/jsdoc-plugins/lib/custom-tags/observable.js
+++ b/packages/jsdoc-plugins/lib/custom-tags/observable.js
@@ -9,9 +9,6 @@ module.exports = {
 	// See http://usejsdoc.org/about-plugins.html#tag-definitions.
 	defineTags( dictionary ) {
 		dictionary.defineTag( 'observable', {
-			mustNotHaveDescription: false,
-			canHaveType: true,
-			canHaveName: true,
 			onTagged( doclet ) {
 				doclet.observable = true;
 			}
--- a/packages/jsdoc-plugins/lib/custom-tags/observable.js
+++ b/packages/jsdoc-plugins/lib/custom-tags/observable.js
@@ -5,8 +5,6 @@
 
 'use strict';
 
-const addMissingEventDoclets = require( '../observable-event-fixer/addmissingeventdoclets' );
-
 module.exports = {
 	// See http://usejsdoc.org/about-plugins.html#tag-definitions.
 	defineTags( dictionary ) {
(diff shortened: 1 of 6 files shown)


packages/jsdoc-plugins/tests/observable-event-provider/addmissingeventdocletsforobservables.js
```diff
@@ -0,0 +1,251 @@
...
+			const inputDoclets = [ {
+				comment: '...',
+				meta: {
+					range: [
+						1861,
+						2061
+					],
+					filename: 'command.js',
+					lineno: 56,
+					path: '/Users/maciejbukowski/workspace/ckeditor5/packages/ckeditor5-core/src',
```
pomek: You should not use real paths in tests. What in case I will change something in this file. Shall I correct the paths because my workspace is different?

Let's use `/workspace/ckeditor5/packages/ckeditor5-core/src` instead.

packages/jsdoc-plugins/lib/observable-event-provider/addmissingeventdocletsforobservables.js
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		const newDoclet = cloneDeep( doclet );
+		const eventName = newDoclet.memberof + '#event:change:' + newDoclet.name;
+
+		if ( !newDoclet.fires ) {
+			newDoclet.fires = [];
+		}
+
+		if ( !newDoclet.fires.includes( eventName ) ) {
```
m-turek: The class should fire an event, not the property.

packages/jsdoc-plugins/tests/observable-event-provider/addmissingeventdocletsforobservables.js
ma2ciek: Hmm, right, didn't catch this. Tests don't use them at all, so they'll work everywhere. I just wanted to add a real case example.

packages/jsdoc-plugins/tests/observable-event-provider/addmissingeventdocletsforobservables.js
```diff
@@ -0,0 +1,251 @@
...
+				kind: 'event',
+				longname: 'module:core/command~Command#event:change:isEnabled'
+			} ];
+
+			const outputDoclets = addMissingEventDocletsForObservables( inputDoclets );
+
+			expect( outputDoclets.length ).to.equal( 2 );
+		} );
+
+		it( 'should add a proper event name to the fixes property', () => {
```
m-turek: "fires" instead of "fixes"

packages/jsdoc-plugins/lib/observable-event-provider/addmissingeventdocletsforobservables.js
m-turek: Ok then 👍

packages/jsdoc-plugins/lib/observable-event-provider/addmissingeventdocletsforobservables.js
```diff
@@ -0,0 +1,122 @@
+/**
+ * @license Copyright (c) 2016-2017, CKSource - Frederico Knabben. All rights reserved.
+ * Licensed under the terms of the MIT License (see LICENSE.md).
+ */
+
+'use strict';
+
+const {	cloneDeep } = require( 'lodash' );
```
Reinmar: Spacing.

packages/jsdoc-plugins/lib/custom-tags/observable.js
```diff
@@ -0,0 +1,20 @@
...
+ * Licensed under the terms of the MIT License (see LICENSE.md).
+ */
+
+'use strict';
+
+module.exports = {
+	// See http://usejsdoc.org/about-plugins.html#tag-definitions.
+	defineTags( dictionary ) {
+		dictionary.defineTag( 'observable', {
+			mustNotHaveDescription: false,
```
Reinmar: Can it have description?

packages/jsdoc-plugins/lib/custom-tags/observable.js
```diff
@@ -0,0 +1,20 @@
...
+			canHaveType: true,
```
Reinmar: Can have a type?

packages/jsdoc-plugins/lib/custom-tags/observable.js
```diff
@@ -0,0 +1,20 @@
...
+			canHaveName: true,
```
Reinmar: Not true as well.

Reinmar:
It generally seems to work but I left a couple of comments anyway ;)

ma2ciek:
Thanks. I didn't catch those ones.

PR #298, opened 2017-10-13
T/296: Remove gulp dependency.

### LINKED ISSUE #296: Remove gulp dependency
This ticket is related to: https://github.com/ckeditor/ckeditor5/issues/473.

I've found that `gulp` is used only in `ckeditor-dev-docs` package and it should be easily removed by using `jsdoc` directly.

Other thing to check is if we want remove `gulp=util` also. So far I've found that we only use `gutil.colors` so maybe using smaller library for CLI color would be better.

jodator:

Other: Remove gulp dependency. Closes #296.

Now all packages are independent of gulp. Depending on usage you might either create "script " entry in pacakge.json to invoke bin executables or require library into script.

* Package `ckeditor5-dev-env` exposes new `translations` binary.
* Removed `gulp-jsdoc3` from building API documentation. Now `jsdoc` is invoked directly.
* Removed `options` param from logger methods. Logger no longer uses `gutil.log` method.

BREAKING CHANGE: Gulp is no longer required to run tasks from ckeditor5-dev-* packages use npm scripts instead.
BREAKING CHANGE: Gulp task are removed. New npm scripts were introduced.

---

Required for https://github.com/ckeditor/ckeditor5/pull/602.

### DIFF
--- a/packages/ckeditor5-dev-docs/lib/index.js
+++ b/packages/ckeditor5-dev-docs/lib/index.js
@@ -84,7 +84,7 @@ function build( config ) {
 					try {
 						tools.shExec( `${ cmd } -c ${ tmpConfig.name }`, { verbosity: 'info' } );
 					} catch ( error ) {
-						return reject( 'Error during JSDoc generation.' );
+						return reject( `Error during JSDoc generation: ${ error.message }` );
 					}
 
 					console.log( `Documented ${ files.length } files!` );
--- a/packages/ckeditor5-dev-docs/lib/index.js
+++ b/packages/ckeditor5-dev-docs/lib/index.js
@@ -5,8 +5,12 @@
 
 'use strict';
 
-const gulp = require( 'gulp' );
-const jsdoc = require( 'gulp-jsdoc3' );
+const fs = require( 'fs' );
+const tmp = require( 'tmp' );
+const map = require( 'map-stream' );
+const vfs = require( 'vinyl-fs' );
+
+const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
 
 module.exports = {
 	build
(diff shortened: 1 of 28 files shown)


packages/ckeditor5-dev-docs/lib/index.js
```diff
@@ -34,35 +38,59 @@ function build( config ) {
...
+					}
+
+					const cmd = require.resolve( 'jsdoc/jsdoc.js' );
+
+					console.log( 'JSDoc started...' );
+
+					try {
+						tools.shExec( `${ cmd } -c ${ tmpConfig.name }`, { verbosity: 'info' } );
+					} catch ( error ) {
+						return reject( 'Error during JSDoc generation.' );
```
pomek: Would be nice to attach message from `error`. It should help understand what went wrong.

packages/ckeditor5-dev-env/package.json
```diff
@@ -34,6 +34,9 @@
     "node": ">=6.0.0",
     "npm": ">=3.0.0"
   },
+  "bin": {
```
pomek: Why did you extract translations tasks from `env/index.js`?

jodator: I could either leave it as is or create a `bin` script (`translations`). I didn't see a point of duplicating `require().*Translation()` in ckeditor5 repo.

This script accepts `collect`, `upload` or `download` param.

Reinmar:
BTW, @jodator, please configure your editor to trim trailing whitespaces.

jodator:
@Reinmar double checked that now - currently I have this on. Might be off for some time as I've reset my editor settings during that time.

PR #299, opened 2017-10-16
Fix paths in webpack-plugin for Windows environment

### LINKED ISSUE #297: Webpack Plugin - replaceTCalls() should work on Windows
The problem occurred when @onurozkan wanted to build an editor on Windows (https://github.com/ckeditor/ckeditor5-build-classic/issues/26).

The build does not work on Windows because regexp used in `replaceTCalls()` function works on Unix systems only.

https://github.com/ckeditor/ckeditor5-dev/blob/b6c059b6d8b6b840cdbd85b703564ca7521fe811/packages/ckeditor5-dev-webpack-plugin/lib/replacetcalls.js#L29

We should detect which separator path should be used: `/` for Unix, `\` for Windows. 

I remember we did it a few months ago when we were introducing Karma.

ma2ciek:

Fix: Fixed paths in webpack-plugin for Windows environment. Closes #297.

### DIFF
--- a/packages/ckeditor5-dev-webpack-plugin/lib/replacetcalls.js
+++ b/packages/ckeditor5-dev-webpack-plugin/lib/replacetcalls.js
@@ -7,6 +7,7 @@
 
 const path = require( 'path' );
 const {	TranslationService } = require( '@ckeditor/ckeditor5-dev-utils' ).translations;
+const utils = require( './utils' );
 
 /**
  * Replaces all function call parameters with translated strings for the t function.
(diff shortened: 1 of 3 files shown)


pomek:
Are we able to provide at least single test for a Windows and Unix paths?

ma2ciek:
Do you mean unit tests? I can export those function with an underscore and test them whether their Regexps are correct.

pomek:
Should be fine.

ma2ciek:
I added tests to the RegExps only as these are the most vulnerable places for that PR. Whole functions are Webpack-dependent and shouldn't be tested by unit tests IMO.

PR #303, opened 2017-11-09
Make webpack plugin being able to run inside the ckeditor5-* package

### LINKED ISSUE #302: Translations are missing on the fresh cloned and built repo ckeditor5-build-classic
## Steps to reproduce:
```
git clone https://github.com/ckeditor/ckeditor5-build-classic
npm i
npm run build
```
## Error:
screen shot 2017-11-09 at 11 01 52
...

ma2ciek:

Fix: Made webpack plugin being able to run inside the ckeditor5-* package. Closes #302.

### DIFF
--- a/packages/ckeditor5-dev-webpack-plugin/lib/replacetcalls.js
+++ b/packages/ckeditor5-dev-webpack-plugin/lib/replacetcalls.js
@@ -6,7 +6,7 @@
 'use strict';
 
 const path = require( 'path' );
-const {	TranslationService } = require( '@ckeditor/ckeditor5-dev-utils' ).translations;
+const { TranslationService } = require( '@ckeditor/ckeditor5-dev-utils' ).translations;
 const utils = require( './utils' );
 
 /**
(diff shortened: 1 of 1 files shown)


ma2ciek:
I'm considering moving the functions into TranslationService class for being able to test them.

pomek:
Would be good to have at least one test which checks whether the paths are correct. But I don't know how much changes it requires.

ma2ciek:
@pomek. I'm working on the https://github.com/ckeditor/ckeditor5/issues/624, so I'll be probably rewriting those part to make TranslationService and the plugin's core part more generic. So I think that we can wait for the incoming changes before rewriting that part for tests. I'll keep that in mind.

pomek:
> Error: original.line and original.column are not numbers -- you probably meant to omit the original mapping entirely and only map the generated position. If so, pass null for the original mapping instead of an object with empty or null values.

But it isn't a part of this ticket.

PR #304, opened 2017-11-10
T/290: Tests are executed as a single entry point

### LINKED ISSUE #290: Check if tests would work better if we had one entry point in Karma
Right now we tell Karma to load all our test files, which is something around ~200 entry points. This is a lot and makes Webpack compile editor 200 times – so when running `npm t` we load 200 entry points.

When we've been setting up Karma I noticed that some guys recommend to have just one entry point, generated on-the-fly. This entry point should simply import all the tests which should be loaded.

It's worth checking if this could help cause our 7000 tests can only be run reliably in Chrome and Safari (https://github.com/ckeditor/ckeditor5/issues/573).

pomek:

Feature: Changed a way how the tests are executed. Karma will receive a single entry point which will import all tests. Closes #290.

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
+++ b/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
@@ -17,7 +17,7 @@ module.exports = function runAutomatedTests( options ) {
 		const config = getKarmaConfig( options );
 
 		if ( !config ) {
-			return reject();
+			return reject( 'Cannot load the configuration file because specified options are invalid.' );
 		}
 
 		const server = new KarmaServer( config, exitCode => {
--- a/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
+++ b/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
@@ -16,7 +16,13 @@ module.exports = function runAutomatedTests( options ) {
 	return new Promise( ( resolve, reject ) => {
 		const config = getKarmaConfig( options );
 
+		if ( !config ) {
+			return reject();
+		}
+
 		const server = new KarmaServer( config, exitCode => {
+			config.removeEntryFile();
+
 			if ( exitCode === 0 ) {
 				resolve();
 			} else {
--- a/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
+++ b/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
@@ -7,85 +7,96 @@
 'use strict';
 
 const path = require( 'path' );
+const glob = require( 'glob' );
+const minimatch = require( 'minimatch' );
+const tmp = require( 'tmp' );
+const log = require( 'karma/lib/logger.js' ).create( 'config' );
 const getWebpackConfigForAutomatedTests = require( './getwebpackconfig' );
 const transformFileOptionToTestGlob = require( '../transformfileoptiontotestglob' );
-const glob = require( 'glob' );
 
-const reporters = [
+const AVAILABLE_REPORTERS = [
 	'mocha',
 	'dots'
 ];
-
-const coverageDir = path.join( process.cwd(), 'coverage' );
 
 /**
  * @param {Object} options
  * @returns {Object}
  */
 module.exports = function getKarmaConfig( options ) {
+	const basePath = process.cwd();
+	const coverageDir = path.join( basePath, 'coverage' );
+
 	if ( !Array.isArray( options.files ) || options.files.length === 0 ) {
-		throw new Error( 'Karma requires files to tests. `options.files` has to be non-empty array.' );
+		return log.error( 'Karma requires files to tests. `options.files` has to be non-empty array.' );
 	}
 
-	if ( !reporters.includes( options.reporter ) ) {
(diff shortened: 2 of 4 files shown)


packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
```diff
@@ -16,7 +16,13 @@ module.exports = function runAutomatedTests( options ) {
 	return new Promise( ( resolve, reject ) => {
 		const config = getKarmaConfig( options );
 
+		if ( !config ) {
+			return reject();
```
ma2ciek: Can we reject with an error?

packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
```diff
@@ -6,67 +6,98 @@
...
 
-		if ( options.sourceMap ) {
-			preprocessorMap[ file ].push( 'sourcemap' );
+		if ( !files.length ) {
+			log.warn( 'Pattern "%s" does not match any file.', singlePattern );
 		}
+
+		allFiles.push( ...files.filter( file => {
+			// Ignore files which are saved in `manual/` directory. There are manual tests.
+			if ( file.match( /manual/ ) ) {
```
ma2ciek: Maybe `tests[/\\]manual[/\\]` would be safer?

packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
```diff
@@ -6,67 +6,98 @@
...
+				return false;
+			}
+
+			// Ignore `_utils` directory as well because there are saved utils for tests.
+			return !minimatch( file, path.join( '**', 'tests', '**', '_utils', '**', '*.js' ) );
```
ma2ciek: The second path can be stored before the loop to not recreate it hundred of times.

packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
```diff
@@ -6,67 +6,98 @@
...
+const AVAILABLE_REPORTERS = [
 	'mocha',
 	'dots'
 ];
 
-const coverageDir = path.join( process.cwd(), 'coverage' );
-
 /**
  * @param {Object} options
  * @returns {Object}
```
ma2ciek: Or `undefined` if the error is logged in this function

packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
```diff
@@ -6,67 +6,98 @@
...
  */
 module.exports = function getKarmaConfig( options ) {
+	const basePath = process.cwd();
+	const coverageDir = path.join( basePath, 'coverage' );
+	const log = karmaLogger.create( 'config' );
+
 	if ( !Array.isArray( options.files ) || options.files.length === 0 ) {
-		throw new Error( 'Karma requires files to tests. `options.files` has to be non-empty array.' );
+		return log.error( 'Karma requires files to tests. `options.files` has to be non-empty array.' );
```
ma2ciek: Does Karma throw an error on `log.error`?

packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
pomek: No, it doesn't. Manual tests can be saved in subdirectories:

```
  '/ckeditor5-engine/tests/manual/tickets/887/1.js',
  '/ckeditor5-engine/tests/view/manual/clickobserver.js',
  '/ckeditor5-basic-styles/tests/manual/basic-styles.js'
```

But I replace the regexp with a glob pattern.

packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
pomek: I'm not sure whether we document it.

packages/ckeditor5-dev-tests/tests/utils/automated-tests/getkarmaconfig.js
```diff
@@ -9,27 +9,60 @@ const mockery = require( 'mockery' );
...
+			tmp: {
+				fileSync: sandbox.stub().callsFake( () => stubs.tmp._fileSync ),
+				_fileSync: {
+					name: '',
+					removeCallback: sandbox.stub()
+				}
+			}
+		};
 
 		mockery.enable( {
```
ma2ciek: Does it work with `useCleanCache` too? :D

pomek: You can check :P

packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
ma2ciek: We document it if a function might return something else.

packages/ckeditor5-dev-tests/tests/utils/automated-tests/getkarmaconfig.js
ma2ciek: It works, I was actually surprised :D

PR #307, opened 2017-11-13
T/306: Optimize doclet validator plugin

### LINKED ISSUE #306: JSDoc: Optimize Doclet Validator plugin
Doclet Validator plugin when run on the sources of CKEditor 5 currently takes up to 50 seconds. This is a lot when combined with other plugins and then running Umberto, so we need some optimizations.

m-turek:

Fix: Optimized doclet validator plugin. Closes #306.

### DIFF
--- a/packages/jsdoc-plugins/lib/utils/collection.js
+++ b/packages/jsdoc-plugins/lib/utils/collection.js
@@ -14,6 +14,8 @@ class Collection {
 	 */
 	constructor() {
 		this._data = {};
+		this._allData = [];
+		this._longnames = [];
 	}
 
 	/**
@@ -28,6 +30,11 @@ class Collection {
 		}
 
 		this._data[ name ].push( record );
+		this._allData.push( record );
+
+		if ( record.longname ) {
+			this._longnames.push( record.longname );
+		}
 	}
 
 	/**
@@ -46,13 +53,16 @@ class Collection {
 	 * @returns {Object[]}
 	*/
 	getAll() {
-		const result = [];
+		return this._allData;
+	}
 
-		Object.keys( this._data ).forEach( key => {
-			result.push( ...this._data[ key ] );
-		} );
-
-		return result;
+	/**
+	 * Returns all longnames.
+	 *
+	 * @returns {Object[]}
+	 */
+	getLongnames() {
+		return this._longnames;
 	}
 }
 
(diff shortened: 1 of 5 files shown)


packages/jsdoc-plugins/lib/utils/collection.js
```diff
@@ -46,13 +53,16 @@ class Collection {
...
+		return this._allData;
+	}
 
-		return result;
+	/**
+	 * Returns all longnames.
+	 *
+	 * @returns {Object[]}
+	 */
+	getLongnames() {
```
ma2ciek: Could you add some simple test(s) to that method to do not decrease CC?

packages/jsdoc-plugins/lib/utils/collection.js
```diff
@@ -46,13 +53,16 @@ class Collection {
...
-			result.push( ...this._data[ key ] );
-		} );
+		return this._allData;
+	}
 
-		return result;
+	/**
+	 * Returns all longnames.
+	 *
+	 * @returns {Object[]}
```
ma2ciek: Should be `String[]`.

packages/jsdoc-plugins/lib/utils/collection.js
ma2ciek: Maybe `getAllLongnames`?

m-turek:
Updated the PR

packages/jsdoc-plugins/lib/utils/collection.js
```diff
@@ -28,6 +30,11 @@ class Collection {
 		}
 
 		this._data[ name ].push( record );
+		this._allData.push( record );
+
+		if ( record.longname ) {
```
ma2ciek: Maybe it was a mistake that I made this class a little bit too generic. Now assuming that an object has a property `longname` is confusing. Because the class became here a `DocletCollection` with your changes. I think, that you can rename it, as it's only used for doclets and there's no other use case for that class. Eventually, you could inherit `DocletCollection` from the `Collection`, but as I mention earlier, there's no need for that.

m-turek: Yeah, that came to my mind too. I think it's best to rename it to DocletCollection as you suggest because it's the only use case of the class.

PR #313, opened 2017-11-14
T/312: Optimized Relation Fixer plugin

### LINKED ISSUE #312: JSDoc: Optimize Relation Fixer plugin
Relation Fixer plugin when run on the sources of CKEditor 5 currently takes up to 10 seconds. There is room for optimizations and I'm going to introduce them.

m-turek:

Fix: Optimized Relation Fixer plugin. Closes #312.

---

### Additional information

Used DocletCollection to reduce the number of operations on arrays of doclets.

### DIFF
--- a/packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
+++ b/packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
@@ -7,6 +7,7 @@
 
 const getMissingDocletsData = require( './getmissingdocletsdata' );
 const cloneDeep = require( 'lodash' ).cloneDeep;
+const DocletCollection = require( '../utils/doclet-collection' );
 
 module.exports = addMissingDoclets;
 
@@ -20,6 +21,13 @@ module.exports = addMissingDoclets;
  */
 function addMissingDoclets( originalDoclets ) {
 	const clonedDoclets = cloneDeep( originalDoclets );
+	const docletCollection = new DocletCollection();
+
+	for ( const doclet of clonedDoclets ) {
+		// group doclets by memberof property
+		docletCollection.add( `memberof:${ doclet.memberof }`, doclet );
+	}
+
 	const entitiesWhichNeedNewDoclets = clonedDoclets.filter( d => {
 		return d.kind === 'class' || d.kind === 'interface' || d.kind === 'mixin';
 	} );
@@ -57,7 +65,7 @@ function addMissingDoclets( originalDoclets ) {
 	for ( const childDoclet of entitiesWhichNeedNewDoclets ) {
 		for ( const setting of settings ) {
 			const missingDocletsData = getMissingDocletsData(
-				clonedDoclets,
+				docletCollection,
 				childDoclet,
 				setting
 			);
--- a/packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
+++ b/packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
@@ -24,7 +24,7 @@ function addMissingDoclets( originalDoclets ) {
 	const docletCollection = new DocletCollection();
 
 	for ( const doclet of clonedDoclets ) {
-		// group doclets by memberof property
+		// Group doclets by memberof property.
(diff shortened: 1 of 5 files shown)


packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -20,6 +21,13 @@ module.exports = addMissingDoclets;
  */
 function addMissingDoclets( originalDoclets ) {
 	const clonedDoclets = cloneDeep( originalDoclets );
+	const docletCollection = new DocletCollection();
+
+	for ( const doclet of clonedDoclets ) {
+		// group doclets by memberof property
```
pomek: A sentence should start with an uppercase letter and end with a period.

packages/jsdoc-plugins/lib/relation-fixer/buildrelations.js
```diff
@@ -27,15 +28,21 @@ module.exports = addRelationArrays;
  */
 function addRelationArrays( originalDoclets ) {
 	const clonedDoclets = cloneDeep( originalDoclets );
+	const docletCollection = new DocletCollection();
+
+	for ( const doclet of clonedDoclets ) {
+		// group doclets by longname
```
ma2ciek: > A sentence should start with an uppercase letter and end with a period.

Same here.

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
```diff
@@ -34,14 +34,14 @@ function getMissingDocletsData( originalDoclets, childDoclet, options ) {
 		clonedDoclet.memberof = childDoclet.longname;
 
 		// Add property `inherited` or `mixed`.
-		const relationProperty = getRelationProperty( originalDoclets, childDoclet, d, options.relation );
+		const relationProperty = getRelationProperty( docletCollection.getAll(), childDoclet, d, options.relation );
```
ma2ciek: Can we pass the whole collection here?

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
```diff
@@ -85,7 +85,7 @@ function getDocletsToAdd( allDoclets, childDoclet, options = {} ) {
 				}
 			}
 
-			return d.memberof === longname && isMatchingFilterOptions && !isUnwanted;
+			return isMatchingFilterOptions && !isUnwanted;
 		} );
 
 		docletsToAdd.push( ...toAdd );
```
ma2ciek: You're mixing functional programming with procedural programming. I'd either use `Array.filter()` or push elements to the original array.

E.g. return `docletsToAdd.concat( toAdd )` should be better.

Reinmar: I see that this code wasn't touched in this PR. So, unless this causes some issues, it's beyond the scope of it.

ma2ciek:
I'd change that method: https://github.com/ckeditor/ckeditor5-dev/blob/664fb2d6d4ceab57c3e814f208848dabe6f547e4/packages/jsdoc-plugins/lib/utils/doclet-collection.js#L24-L27 as the `name` has become incorrect. Maybe `category` would fit better?

ma2ciek:
Actually, you could correct more occurrences of the 'record' and 'object' after the previous PR. I didn't notice that.

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
m-turek: We can, but why? does it matter? (maybe I'm missing something)

m-turek:
> Actually, you could correct more occurrences of the 'record' and 'object' after the previous PR. I didn't notice that.

Is it ok to do it in this PR? or should I make another one for that?

ma2ciek:
> Is it ok to do it in this PR? or should I make another one for that?

I think, that we can do this in that PR as it's related to current changes.

ma2ciek:
Thanks for those changes to `DocletColelction`.

PR #326, opened 2017-11-23
T/325: MT will work on Windows

pomek:

Fix: Manual tests will work in Windows environments properly. Closes: #325.


packages/ckeditor5-dev-tests/tests/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -34,7 +35,20 @@ describe( 'compileHtmlFiles', () => {
 			},
 
 			path: {
-				join: sandbox.stub( path, 'join' ).callsFake( ( ...chunks ) => chunks.join( '/' ) )
+				join: sandbox.stub( path, 'join' ).callsFake( ( ...chunks ) => chunks.join( separator ) ),
```
ma2ciek: Don't you think, that it's easier to replace the whole `path` lib with the `path.win32` or `path.posix` depending on the env that you want to test?

PR #329, opened 2017-11-30
T/290: Restored tests as a single entry point (Karma)

### LINKED ISSUE #290: Check if tests would work better if we had one entry point in Karma

pomek:

Feature: Restored previous Karma behavior. Tests will be compiled to a single entry point. Closes #290.

---

image


szymonkups:
When running for example `npm t -- --files=core -c` there is no coverage report:
image
Also here coverage raport is empty:
image

pomek:
f2fff73 should fix that.

image

Could you confirm?

ma2ciek:
Not sure why, but Karma runs tests few times when the watch flag is turned on. E.g. try that: `npm run test -- --files utils -w`. It could be a separate issue connected with karma watcher.

pomek:
But everything works fine when call `-c -w` :|

ma2ciek:
`master` branch works as expected for me (runs tests only once).

> But everything works fine when call -c -w :|

pomek:
I changed proposed solution. 

`getKarmaConfig()` should not create the entry file itself. I moved the logic a layer higher. 

Now, the entry file is creating before call the `getKarmaConfig()` function. The entry file won't be removed after tests execution. It is saved under path `build/.automated-tests/entry-point.js`.

Unfortunately, I'm still having some troubles with double tests execution in watch mode :/

I will report a ticket to Karma's devs but I need to prepare some setup that allows reproducing the issue easy way.

Reinmar:
> Unfortunately, I'm still having some troubles with double tests execution in watch mode :/

That's an old issue. Not something introduced by these changes. And I think that you checked in the past that it's reported.

PR #347, opened 2017-12-20
T/346: Karma / Webpack watcher

### LINKED ISSUE #346: Automated tests – Karma/Webpack watcher runs tests n times
image

Fortunately, this issue was reported here – https://github.com/webpack/watchpack/issues/25.

Fix for the issue is simple. The code below should be added here:

https://github.com/ckeditor/ckeditor5-dev/blob/8feba11aa6908e69141283b2b1fd5e0ea3ff68fe/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js#L91-L92

```js
const now = Date.now() / 1000;
const then = now - 5;
fs.utimesSync( ENTRY_FILE_PATH, then, then );
```

pomek:

Fix: Karma will not execute the tests multiple times when works in watch mode. Closes #346.

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
+++ b/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
@@ -92,7 +92,7 @@ function createEntryFile( globPatterns ) {
 
 	// See: https://github.com/webpack/watchpack/issues/25.
 	const now = Date.now() / 1000;
-	const then = now - 5;
+	const then = now - 10;
 	fs.utimesSync( ENTRY_FILE_PATH, then, then );
 }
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
```diff
@@ -89,6 +89,11 @@ function createEntryFile( globPatterns ) {
 		.join( '\n' );
 
 	fs.writeFileSync( ENTRY_FILE_PATH, filesImports + '\n' );
+
+	// See: https://github.com/webpack/watchpack/issues/25.
```
ma2ciek: I'd add one comment more and specify the problem

packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
```diff
@@ -89,6 +89,11 @@ function createEntryFile( globPatterns ) {
...
+	const now = Date.now() / 1000;
+	const then = now - 5;
+	fs.utimesSync( ENTRY_FILE_PATH, then, then );
```
ma2ciek: Probably it's not a problem, but that method takes only integers according to docs - https://nodejs.org/api/fs.html#fs_fs_futimessync_fd_atime_mtime

PR #356, opened 2018-01-22
JSDoc: Support extending typedefs

### LINKED ISSUE #355: JSDoc: Support extending typedefs
It would be great if we could support situations where one typedef extends another typedef. E.g.:

```
/**
 * Heading option descriptor. Compatible with {@link module:engine/conversion/definition-based-converters~ConverterDefinition}.
 *
 * @typedef {Object} module:heading/heading~HeadingOption
 * @extends module:engine/conversion/definition-based-converters~ConverterDefinition
 * @property {String} title The user-readable title of the option.
 * @property {String} class The class which will be added to the dropdown item representing this option.
 */
```

JSDoc apparently doesn't support this: jsdoc3/jsdoc#1199.

m-turek:

Feature: JSDoc: Support extending typedefs. Closes #355.

### DIFF
--- a/packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
+++ b/packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
@@ -22,10 +22,15 @@ module.exports = addMissingDoclets;
 function addMissingDoclets( originalDoclets ) {
 	const clonedDoclets = cloneDeep( originalDoclets );
 	const docletCollection = new DocletCollection();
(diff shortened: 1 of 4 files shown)


Reinmar:
Shouldn't those properties be sorted?

image

Do we sort them in other cases?

Reinmar:
Yep, they are:

image

Reinmar:
@m-turek told me that this will be handled on Umberto's side. So we can merge this PR. Thanks!

PR #363, opened 2018-02-09
T/361

### LINKED ISSUE #361: [JSDoc] Fix relation-fixer plugin
The plugin includes interfaces in class hierarchy, which is wrong

hierarchy

m-turek:

Fix: JSDoc relation-fixer: Minor change to building inheritance hierarchy. Closes #361.


Reinmar:
Because "Minor change" tells us so much about what it does :P So, why it was needed?

m-turek:
Before the change only classes could form proper inheritance tree, so if we ever had 3 interfaces inheriting one from another, that wouldn't create a proper inheritance tree.
After the change it will work ok not only for classes, but also interfaces etc

PR #364, opened 2018-02-13
Disabled BrowserStack for community's PRs

### LINKED ISSUE #358: Switch off BrowserStack for PRs from community
The problem occurs here – https://github.com/ckeditor/ckeditor5-image/pull/168.

We have a PR created from the forked repository. Unfortunately, [Travis does not allow decrypting variables in such cases](https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions).

> A pull request sent from a fork of the upstream repository could be manipulated to expose any environment variables. The upstream repository’s maintainer would have no protection against this attack, as pull requests can be sent by anyone with a fork.
> 
> **Travis CI makes encrypted variables and data available only to pull requests coming from the same repository.** These are considered trustworthy, as only members with write access to the repository can send them.
>
> **Pull requests sent from forked repositories do not have access to encrypted variables or data.**

So, we have to change our dev script. BrowserStack cannot be used if a PR comes from the forked repositories.

We also cannot change a way how we want to decrypt the variables because a user can put in forked `.travis.yml` file something like `echo VARIABLE_NAME` and our secret will not be a secret anymore.

pomek:

Fix: BrowserStack will be disabled for PR builds that come from the community. Closes #358.

### DIFF
--- a/packages/ckeditor5-dev-tests/tests/utils/automated-tests/getkarmaconfig.js
+++ b/packages/ckeditor5-dev-tests/tests/utils/automated-tests/getkarmaconfig.js
@@ -71,4 +71,57 @@ describe( 'getKarmaConfig', () => {
 		expect( karmaConfig ).to.have.own.property( 'browsers' );
 		expect( karmaConfig ).to.have.own.property( 'singleRun', true );
 	} );
+
+	describe( 'BrowserStack integration', () => {
+		beforeEach( () => {
+			process.env.BROWSER_STACK_USERNAME = 'username';
+			process.env.BROWSER_STACK_ACCESS_KEY = 'access-key';
+		} );
+
+		// BROWSER_STACK_USERNAME=username BROWSER_STACK_ACCESS_KEY=access-key npm run test -- --files=autoformat
+		it( 'should be enabled when tests were called on a dev machine', () => {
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-dev-tests/tests/utils/automated-tests/getkarmaconfig.js
```diff
@@ -71,4 +71,57 @@ describe( 'getKarmaConfig', () => {
 		expect( karmaConfig ).to.have.own.property( 'browsers' );
 		expect( karmaConfig ).to.have.own.property( 'singleRun', true );
 	} );
+
+	describe( 'BrowserStack integration', () => {
+		beforeEach( () => {
```
Reinmar: What about tear down after overriding `process.env`? This is a global object so it's better to use stubbing and sandboxes.

pomek: It's already done https://github.com/ckeditor/ckeditor5-dev/blob/19184b1fd67f50c904cd2acc825cfbeb3a37d45c/packages/ckeditor5-dev-tests/tests/utils/automated-tests/getkarmaconfig.js#L23-L42.

PR #367, opened 2018-02-15
Fix inheritance in event doclets for observables

### LINKED ISSUE #368: Inherited events aren't documented as inherited
image

This is `DropdownButton`. It extends the `Button` interface and adds just one thing the `open` event. All the rest should be marked as inherited and invisible when the "FILTERING: INHERITED" option is checked out.

This may be a bug in the JSDoc plugins which adds events for all observable properties. Perhaps it adds them when processing `DropdownButton`, but it misses the fact that they should be marked as inherited because these observable properties are inherited too.

ma2ciek:

Fix: Fixed inheritance in event doclets for observables. Closes #368.


pomek:
Tests don't pass.

ma2ciek:
Fixed. It should be ready now.

PR #373, opened 2018-03-06
Upgrade webpack plugin and testing environment for webpack v4 

### LINKED ISSUE #371: Build fails with webpack 4
Upgraded my (working) project from webpack 3 to webpack 4, and the build now fails with the below error:

```
ERROR in ./node_modules/@ckeditor/ckeditor5-basic-styles/src/bold.js
Module build failed: TypeError: Cannot read property 'translateSource' of undefined
    at Object.translateSourceLoader (.\node_modules\@ckeditor\ckeditor5-dev-webpack-plugin\lib\translatesourceloader.js:16:22)
```

Looks like it's due to using the deprecated `this.options` value. See [webpack 4 migration guide](https://medium.com/webpack/webpack-4-migration-guide-for-plugins-loaders-20a79b927202)

ma2ciek:

Other: Updated `CKEditorWebpackPlugin` to support webpack@4. Updated webpack in testing tools. Closes #371.

---

### Additional information

Changes in other repos:
* https://github.com/ckeditor/ckeditor5/tree/t/ckeditor5-dev/371 (https://github.com/ckeditor/ckeditor5/pull/1118)
* https://github.com/ckeditor/ckeditor5-build-classic/tree/t/ckeditor5-dev/371
* https://github.com/ckeditor/ckeditor5-build-inline/tree/t/ckeditor5-dev/371
* https://github.com/ckeditor/ckeditor5-build-balloon/tree/t/ckeditor5-dev/371
* https://github.com/ckeditor/ckeditor5-build-decoupled-document/tree/t/ckeditor5-dev/371

BREAKING CHANGE: Breaks compatibility with webpack v3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "devDependencies": {
     "@ckeditor/ckeditor5-dev-env": "^10.0.0",
-    "chalk": "^2.1.0",
+    "chalk": "^2.4.1",
     "eslint": "^4.15.0",
     "eslint-config-ckeditor5": "^1.0.8",
     "husky": "^0.14.3",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
     ]
   },
   "eslintIgnore": [
-    "coverage/**"
+	"coverage/**",
+	"packages/*/node_modules/**"
   ],
   "version": "0.0.67"
 }
--- a/packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
+++ b/packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
@@ -10,10 +10,10 @@ module.exports = function getLicenseBanner() {
 
 	/* eslint-disable indent */
 	return (
-`/*--------------------------------------------------------------------------------------*
- * @license Copyright (c) 2003-${ date.getFullYear() }, CKSource - Frederico Knabben. All rights reserved. *
- * For licensing, see LICENSE.md.                                                       *
- *--------------------------------------------------------------------------------------*/`
+`/*!
+ * @license Copyright (c) 2003-${ date.getFullYear() }, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */`
 	);
 	/* eslint-enable indent */
 };
--- a/packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
+++ b/packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
@@ -10,10 +10,10 @@ module.exports = function getLicenseBanner() {
 
 	/* eslint-disable indent */
 	return (
-`/*!
- * @license Copyright (c) 2003-${ date.getFullYear() }, CKSource - Frederico Knabben. All rights reserved.
- * For licensing, see LICENSE.md.
- */`
+`/*--------------------------------------------------------------------------------------*
+ * @license Copyright (c) 2003-${ date.getFullYear() }, CKSource - Frederico Knabben. All rights reserved. *
+ * For licensing, see LICENSE.md.                                                       *
+ *--------------------------------------------------------------------------------------*/`
 	);
 	/* eslint-enable indent */
 };
--- a/packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
+++ b/packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
@@ -10,7 +10,7 @@ module.exports = function getLicenseBanner() {
 
 	/* eslint-disable indent */
 	return (
-`/**
+`/*!
  * @license Copyright (c) 2003-${ date.getFullYear() }, CKSource - Frederico Knabben. All rights reserved.
  * For licensing, see LICENSE.md.
  */`
--- a/packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
+++ b/packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
@@ -8,6 +8,10 @@
 module.exports = function getLicenseBanner() {
 	const date = new Date();
 
+	// License banner starts with `!`. That combines with uglifyjs' `comments` /^!/ option
+	// make webpack preserve that banner while cleaning code from others comments during the build task.
+	// It's because UglifyJsWebpackPlugin minification takes place after adding a banner.
+
 	/* eslint-disable indent */
 	return (
 `/*!
--- a/packages/ckeditor5-dev-webpack-plugin/lib/servetranslations.js
+++ b/packages/ckeditor5-dev-webpack-plugin/lib/servetranslations.js
@@ -54,32 +54,39 @@ module.exports = function serveTranslations( compiler, options, translationServi
 		}
 	}
 
-	// Add core translations before `translatesourceloader` starts translating.
-	compiler.plugin( 'after-resolvers', () => {
-		const resolver = compiler.resolvers.normal;
+	// Add core translations before `translateSourceLoader` starts translating.
+	compiler.hooks.normalModuleFactory.tap( 'CKEditor5Plugin', normalModuleFactory => {
+		const resolver = normalModuleFactory.getResolver( 'normal' );
+		const corePackageSampleResource = envUtils.getCorePackageSampleResource();
 
-		envUtils.getCorePackage( cwd, resolver ).then( corePackage => {
+		resolver.resolve( cwd, cwd, corePackageSampleResource, {}, ( err, pathToResource ) => {
+			if ( err ) {
+				console.error( err );
+
+				return;
+			}
+
+			const corePackage = envUtils.getCorePackagePath( pathToResource );
+
 			translationService.loadPackage( corePackage );
 		} );
 	} );
 
 	// Load translation files and add a loader if the package match requirements.
-	compiler.plugin( 'normal-module-factory', nmf => {
-		nmf.plugin( 'after-resolve', ( resolveOptions, done ) => {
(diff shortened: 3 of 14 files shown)


ma2ciek:
I just saw that one test, which uses `import * as enitreModule from 'someModule'` fails during creating a spy. https://github.com/ckeditor/ckeditor5-ui/blob/a98b63ef0b7995db9fde3f2296e5fddcadb36a5d/tests/panel/balloon/balloonpanelview.js#L175

ma2ciek:
Yep, now webpack creates only getters for module exports, so `sinon` can't wrap these functions. We can drop this test or attach `getOptimalPosition` to the BalloonPanelView, to enable wrapping it.

cc @oleq, @Reinmar

ma2ciek:
```
$ lerna bootstrap
lerna info version 2.11.0
lerna info versioning independent
lerna info Bootstrapping 8 packages
lerna info lifecycle preinstall
lerna info Installing external dependencies
lerna info hoist Installing hoisted dependencies into root

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

Not sure whether it's too low timeout set for Lerna 🤔

packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
```diff
@@ -10,7 +10,7 @@ module.exports = function getLicenseBanner() {
 
 	/* eslint-disable indent */
 	return (
-`/**
+`/*!
```
ma2ciek: I changed comment, so it differs now a little bit from the original comments used in the code. It's the result of https://github.com/webpack/webpack/issues/6630#issuecomment-400990552.

pjasiun: Can we use `/*` (with one `*`)? This `!` looks like a typo.

pjasiun: Or something like:
```
/*-------------------------*
 *                         *
 *-------------------------*/
```
Anything what does not look like a typo.

pomek:
> Not sure whether it's too low timeout set for Lerna 🤔

Travis has some troubles.

packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
ma2ciek: I'm ok with

```
/*-------------------------*
 *                         *
 *-------------------------*/
```

It just needs to differ from original license headers.

packages/ckeditor5-dev-utils/lib/bundler/getlicensebanner.js
```diff
@@ -10,10 +10,10 @@ module.exports = function getLicenseBanner() {
 
 	/* eslint-disable indent */
 	return (
-`/**
- * @license Copyright (c) 2003-${ date.getFullYear() }, CKSource - Frederico Knabben. All rights reserved.
- * For licensing, see LICENSE.md.
- */`
+`/*--------------------------------------------------------------------------------------*
```
Reinmar: It doesn't seem like nicely aligned. Is it GH's way of rendering  this string? Or isn't the box wide enough?

Reinmar: It requires a comment on why it needs to look this way.

ma2ciek: `${ date.getFullYear() }` that becomes 4-character-long as it evaluates.

ma2ciek: We can think about a shorter comment or some short custom plugin for the banner.

Reinmar: BTW can't we look for `/\/\*\n \* @license/`? And use the old license comment?

Reinmar: Because once you wrote about a shorter comment I thought that it unnecessarily increases a size of every build. Not by much, perhaps it gzips well, but it seems unnecessary if the regexp can look for `@license` too.

Reinmar: Doh, we use `@license` in every single source file :D

ma2ciek: See https://github.com/ckeditor/ckeditor5-dev/pull/373#discussion_r199095445 too 😄

Reinmar: After a couple of deep breaths, I'm for extending `CKEditorWebpackPlugin` so it adds this banner in a way which doesn't conflict with minification.

This way, we'll simplify maintaining configurations. We'll be able to get rid of this:

```js
		new webpack.BannerPlugin( {
			banner: bundler.getLicenseBanner(),
			raw: true
		} ),
```

And will be able to avoid adding this:

```js
new UglifyJsWebpackPlugin( {
		sourceMap: true,
		uglifyOptions: {
			output: {
				comments: /^\**!/
			}
		}
	} )
```

So, `CKEditorWebpackPlugin()` should have the following option:

`[options.addLicense=false] {Boolean|String}`

* `false` – don't add
* `true` – add the default one
* `String` – if any of sub-CKEditor projects will require a bit different license text

Additionally, we may need to rethink configuration because it will get crowded:

```js
		new CKEditorWebpackPlugin( {
			// Main language that will be built into the main bundle.
			language: 'en',

			// Additional languages that will be emitted to the `outputDirectory`.
			// This option can be set to an array of language codes or `'all'` to build all found languages.
			// The bundle is optimized for one language when this option is omitted.
			additionalLanguages: 'all',

			// Optional directory for emitted translations. Relative to the webpack's output.
			// Defaults to `'translations'`.
			// outputDirectory: 'ckeditor5-translations',

			// Whether the build process should fail if an error occurs.
			// Defaults to `false`.
			// strict: true,

			// Whether to log all warnings to the console.
			// Defaults to `false`.
			// verbose: true
		} ),
```

Reinmar: I'll move this to the ticket cause it'd be pity to lose this.

ma2ciek:
Ready for the review once again :)

package.json
```diff
@@ -33,7 +33,8 @@
     ]
   },
   "eslintIgnore": [
-    "coverage/**"
+	"coverage/**",
+	"packages/*/node_modules/**"
```
Reinmar: Why this?

packages/ckeditor5-dev-webpack-plugin/lib/servetranslations.js
```diff
@@ -54,33 +54,38 @@ module.exports = function serveTranslations( compiler, options, translationServi
 		}
 	}
 
-	// Add core translations before `translatesourceloader` starts translating.
-	compiler.plugin( 'after-resolvers', () => {
-		const resolver = compiler.resolvers.normal;
+	// Add core translations before `translateSourceLoader` starts translating.
+	compiler.hooks.normalModuleFactory.tap( 'CKEditor5Plugin', normalModuleFactory => {
```
Reinmar: Is this consistent with webpack plugins' naming convention?

packages/ckeditor5-dev-webpack-plugin/package.json
```diff
@@ -6,7 +6,12 @@
   "main": "lib/index.js",
   "dependencies": {
     "@ckeditor/ckeditor5-dev-utils": "^9.0.1",
-    "rimraf": "^2.6.2"
+    "chalk": "^2.4.1",
```
Reinmar: Is it imported anywhere?

packages/ckeditor5-dev-webpack-plugin/package.json
```diff
@@ -6,7 +6,12 @@
...
+    "rimraf": "^2.6.2",
+    "webpack-sources": "^1.0.0"
```
Reinmar: And this too?

Reinmar: I actually see it was imported previously. So it was missing previously, yes?

ma2ciek: https://github.com/ckeditor/ckeditor5-dev/pull/373/files/127cd77aa4d7d048b125896ef8448ce31443ad63#diff-088f1ffe6ad43d8c12490e6e14c9c73bL12

packages/ckeditor5-dev-webpack-plugin/package.json
ma2ciek: https://github.com/ckeditor/ckeditor5-dev/pull/373/files/127cd77aa4d7d048b125896ef8448ce31443ad63#diff-088f1ffe6ad43d8c12490e6e14c9c73bR8

package.json
ma2ciek: When installing two versions of dependencies in 2 different packages inside `ckeditor5-dev`'s subrepos, both aren't moved to the `ckeditor5-dev` by Lerna.

packages/ckeditor5-dev-webpack-plugin/lib/servetranslations.js
ma2ciek: Yep, see https://github.com/webpack/tapable.

packages/ckeditor5-dev-webpack-plugin/package.json
ma2ciek: Both were missing.

Issue #41, opened 2017-01-05
Warning - "The `this` keyword is equivalent..." - during `gulp build`


pomek:
Moved from: https://github.com/ckeditor/ckeditor5-dev-bundler-rollup/issues/27.

Reinmar:
This package was removed for now. We'll be reworking the Rollup support.

Issue #351, opened 2017-12-28
Build of manual tests works weird when name of project directory is other then ckeditor5


oskarwrobel:
I wanted to create second workspace with CKE5 so I cloned the project to `ckeditor5-second` directory and then I run manual tests for the entire project and booooom:
```shell
~/P/c/ckeditor5-second (master|✔) $ npm run test:manual

> ckeditor5@1.0.0-alpha.2 test:manual /Users/oskarwrobel/Projects/cksource/ckeditor5-second
> node --max_old_space_size=4096 node_modules/@ckeditor/ckeditor5-dev-tests/bin/test-manual.js

Removed directory '/Users/oskarwrobel/Projects/cksource/ckeditor5-second/build/.manual-tests'
WebpackOptionsValidationError: Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.entry should be one of these:
   object { <key>: non-empty string | [non-empty string] } | non-empty string | [non-empty string] | function
   -> The entry point(s) of the compilation.
   Details:
    * configuration.entry should not be empty.
      -> Multiple entry bundles are created. The key is the chunk name. The value can be a string or an array.
    * configuration.entry should be a string.
      -> An entry point without name. The string is resolved to a module which is loaded upon startup.
    * configuration.entry should be an array:
      [non-empty string]
    * configuration.entry should be an instance of function
      -> A Function returning an entry object, an entry string, an entry array or a promise to these things.
```

When I run tests for specified package I don't see the error anymore but still, something is wrong because log is very short (many files are missing) and tests don't work:
```shell
~/P/c/ckeditor5-second (master|✔) $ npm run test:manual -- --files upload

> ckeditor5@1.0.0-alpha.2 test:manual /Users/oskarwrobel/Projects/cksource/ckeditor5-second
> node --max_old_space_size=4096 node_modules/@ckeditor/ckeditor5-dev-tests/bin/test-manual.js "--files" "upload"

Removed directory '/Users/oskarwrobel/Projects/cksource/ckeditor5-second/build/.manual-tests'
[Webpack] Starting scripts compilation...
Processing '/Users/oskarwrobel/Projects/cksource/ckeditor5-second/packages/ckeditor5-upload/tests/manual/imageplaceholder'...
[...]

PR #382, opened 2018-03-19
Issue #381 resolution (ckeditor5-build-classic issue #41)

### LINKED ISSUE #381: themes do not load styles properly when the project name starts with ckeditor5-
This was logged as Issue 41 on the classic editor build.

https://github.com/ckeditor/ckeditor5-build-classic/issues/41

carltoncolter:
Added ig to get the last match instead of the first match.  This will allow the project to start with ckeditor5- and still work properly.

### Suggested merge commit message (patch)

Type: Message. Closes #381.

---

### Additional information

Encountered the problem originally while trying to build the ckeditor5-build-classic.  I traced it down to not importing the themes.  This was due to the repository folder on my computer matching the search params for the theme.

After tracing the issue, I was able to determine it was a bug in the themeimporter.js.  It was doing a name match on "ckeditor5-" in the getPackageName method.

The match line was:
`const match = inputFilePath.match( /ckeditor5-[^\\/]+/ );`

I changed it to:
`const match = inputFilePath.match( /ckeditor5-[^\\/]+/ig );`

This ensures I'm getting the last ckeditor5- match, which is the package name.  The way the method was written (returning `match.pop()`) makes me believe the intended result was to return the last match.  Adding ig ensures the search returns all matches and makes `match.pop()` return the last match, which is the package name.

### DIFF
--- a/packages/ckeditor5-dev-utils/lib/styles/themeimporter.js
+++ b/packages/ckeditor5-dev-utils/lib/styles/themeimporter.js
@@ -188,7 +188,7 @@ function getThemeFilePath( themePath, inputFilePath ) {
 // @param {String} inputFilePath A path to the file.
 // @returns {String} The name of the package.
 function getPackageName( inputFilePath ) {
-	const match = inputFilePath.match( /ckeditor5-[^\\/]+/ );
+	const match = inputFilePath.match( /ckeditor5-[^\\/]+/ig );
 
 	if ( match ) {
 		return match.pop();


Reinmar:
@pomek, there's also https://github.com/ckeditor/ckeditor5-dev/issues/351. I'm not sure if these issues are anyhow related, but could we somehow remove this limitation regarding the directory name?

pomek:
I checked out the tip: "_Added ig to get the last match instead of the first match._" and it was helpful. Thanks @carltoncolter.

> could we somehow remove this limitation regarding the directory name?

@Reinmar, I answer this question in #351 which is IMO more appropriate place to take this discussion. See https://github.com/ckeditor/ckeditor5-dev/issues/351#issuecomment-374529931.

pomek:
One more thing – Travis fails because of #383. I prepared fix for that (#384).

oleq:
I confirm this patch fixes the problem. For some paths containing more than one "ckeditor5-" substring like

```
/Users/oleq/Sites/cke5/ckeditor5-build-balloon-1.0.0-beta.1/node_modules/@ckeditor/ckeditor5-link/theme/linkform.css
```

the algorithm determined the package name as the first one found, i.e. `ckeditor5-build-balloon-1.0.0-beta.1` instead of `ckeditor5-link`, which broke the importing.

oleq:
What we need are tests. Can you handle that @pomek?

Reinmar:
The issue was fixed by https://github.com/ckeditor/ckeditor5-dev/pull/386. Thanks for locating the issue!

PR #385, opened 2018-03-20
Enable ckeditor5 builds to work with RequireJS 

ma2ciek:

Fix: Enabled ckeditor5 builds to work with `RequireJS`. Closes ckeditor/ckeditor5#914.

---

### Additional information

See https://github.com/ckeditor/ckeditor5/issues/914#issuecomment-374540658.

Branches:

https://github.com/ckeditor/ckeditor5-utils/tree/t/ckeditor5/914

https://github.com/ckeditor/ckeditor5-build-classic/tree/t/ckeditor5/914
https://github.com/ckeditor/ckeditor5-build-inline/tree/t/ckeditor5/914
https://github.com/ckeditor/ckeditor5-build-balloon/tree/t/ckeditor5/914
https://github.com/ckeditor/ckeditor5-build-decoupled-document/tree/t/ckeditor5/914


packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -211,7 +211,15 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 			const stringifiedTranslations = JSON.stringify( translatedStrings )
 				.replace( /"([a-z]+)":/g, '$1:' );
 
-			const outputBody = `CKEDITOR_TRANSLATIONS.add('${ language }',${ stringifiedTranslations })`;
+			const outputBody = (
+				// We need to ensure that the CKEDITOR_TRANSLATIONS variable exist and if it exists, we need to extend it.
+				'(function(d){' +
```
Reinmar: Do we add this bit after the code is minified?

ma2ciek: Hmm, I was thinking about someone that wants to use our builds and targets ES5.  Otherwise, we can change it to arrow fn.

ma2ciek: I used `Object.assign` later 🤦‍♂️

ma2ciek: I replaced it with an arrow function. But targeting ES5 might be important, and this code won't be transpiled, it should be considered.

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -211,7 +211,15 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 			const stringifiedTranslations = JSON.stringify( translatedStrings )
 				.replace( /"([a-z]+)":/g, '$1:' );
 
-			const outputBody = `CKEDITOR_TRANSLATIONS.add('${ language }',${ stringifiedTranslations })`;
+			const outputBody = (
+				// We need to ensure that the CKEDITOR_TRANSLATIONS variable exist and if it exists, we need to extend it.
+				'(d=>' +
```
Reinmar: I didn't meant why it was a `function`. I meant – why it's minified?

ma2ciek: Because it won't be transpiled later. It's added to the built asset. And we want the minified code in the production mode.

ma2ciek: Hmm. Maybe I'm a little bit wrong here. After changing listening to `emit` event to listening to `optimize-chunk-assets` it might be optimized later. I'll check it.

ma2ciek: Yes. In fact, transpilation is done later, but it breaks sourcemaps. I'll try to work-around this quickly.

ma2ciek:
OK. I've found few interesting things digging more into webpack code.

`optimize-chunk-assets` - on this hook many plugins try to do some things, that's why I felt before. E.g. Babily minifies output files, `webpack.BannerPlugin` adds a banner with the license. These plugins mainly look at the compilation chunks and for each chunk in chunks, for each fileName in chunk, they mutate the output for the corresponding file. That has some pros and cons in our case.

If we want to have nice looking code that you've proposed https://github.com/ckeditor/ckeditor5-dev/pull/385#pullrequestreview-105310694, than we will need to minify all assets, including translation assets after adding some minifier. But that will also lead to emitting map files (as have source-map enabled) and adding license banner to each translation file - and I'm not sure if we want it.

ma2ciek:
To make it clear, IMO it's best to provide already minified translations and do not touch compilation chunks as other plugins can modify related files and we actually don't want it.

ma2ciek:
I've tested it with targeting one language and multiple language and sourcemaps should work fine now. I've updated builds on the `t/ckeditor5/914` branches.

Reinmar:
R- for reasons explained in https://github.com/ckeditor/ckeditor5/issues/914#issuecomment-376099477

Reinmar:
I'll finish this PR. Reassigning.

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -211,7 +211,16 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
...
-			const outputBody = `CKEDITOR_TRANSLATIONS.add('${ language }',${ stringifiedTranslations })`;
+			const outputBody = (
+				// We need to ensure that the CKEDITOR_TRANSLATIONS variable exists and if it exists, we need to extend it.
+				// Use ES5 because this bit will not be transpiled!
+				'(function(d){' +
+					`d['${ language }']=Object.assign(` +
+						`d['${ language }']||{},` +
+						`${ stringifiedTranslations }` +
+					')' +
+				'})(window.CKEDITOR_TRANSLATIONS||(window.CKEDITOR_TRANSLATIONS={}));'
```
Reinmar: I think that this kind of access to the global object (through `window`) is going to hurt in some odd envs, but let's keep it for now.

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -211,7 +211,16 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 			const stringifiedTranslations = JSON.stringify( translatedStrings )
 				.replace( /"([a-z]+)":/g, '$1:' );
 
-			const outputBody = `CKEDITOR_TRANSLATIONS.add('${ language }',${ stringifiedTranslations })`;
+			const outputBody = (
+				// We need to ensure that the CKEDITOR_TRANSLATIONS variable exists and if it exists, we need to extend it.
+				// Use ES5 because this bit will not be transpiled!
+				'(function(d){' +
+					`d['${ language }']=Object.assign(` +
```
ma2ciek: It's still an ES6 code. We can change it to `for in` loop to be fully ES5 compatible.

Reinmar: `Object.assign()` needs to be provided by a polyfill for the rest of the code anyway. This doesn't get transpiled.

ma2ciek: This might be tricky for someone, as we add that code at the top of the bundle.

Reinmar: Right now we don't support any non-ES6 env anyway, so it's even hard to test. Let's keep it in our minds. But I don't see a problem in injecting the ES6 polyfill before our scripts.

PR #394, opened 2018-04-06
Remove babel loader rule from webpack's config when testing coverage

### LINKED ISSUE #66: No need to transpile ES6 module to CJS for automated tests?
We have now this piece of code:

```js
	if ( options.coverage ) {
		config.module.rules.push(
			{
				test: /\.js$/,
				loader: 'babel-loader',
				query: {
					cacheDirectory: true,
					plugins: [ require( 'babel-plugin-transform-es2015-modules-commonjs' ) ]
				}
			},
```

I remember that there was a reason why we transpiled modules to CJS, but I don't remember if that was because of https://github.com/ternjs/acorn/issues/493 of some other issue (with Istanbul perhaps?). Do you recall, @ma2ciek?

ma2ciek:

Internal: Removed babel loader rule from webpack's config when testing coverage. Closes #66.


Reinmar:
I wonder how much it will improve CI speed :D I guess not much, but should help a bit. Less npm installs, less transpilation.

PR #395, opened 2018-04-10
"master-revisions" will be synced with "master" correctly

### LINKED ISSUE #375: master-revisions isn't correctly synced with master
If you'd do `git diff master master-revisions --stat` you'd see that these two branches differ significantly.

They should be identical with the difference being only in `mgit.json`.

The current difference make it impossible to e.g. use `npm run docs` in `master-revisions`.

pomek:

Fix: During saving revisions, the branches `master-revisions` and `master` weren't synced correctly. Closes #375.

### DIFF
--- a/packages/ckeditor5-dev-tests/bin/save-revision.js
+++ b/packages/ckeditor5-dev-tests/bin/save-revision.js
@@ -21,7 +21,10 @@ const mainRepoUrl = 'https://github.com/ckeditor/ckeditor5';
 const revisionBranch = `${ branch }-revisions`;
 
 // Clone the repository.
-exec( `git clone -b ${ revisionBranch } ${ mainRepoUrl }.git` );
+exec( `git clone ${ mainRepoUrl }.git` );
+
+// And check out to the revision branch.
+exec( `git checkout ${ revisionBranch } ` );
 
 // Change current dir to cloned repository.
 process.chdir( path.join( process.cwd(), 'ckeditor5' ) );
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-dev-tests/bin/save-revision.js
```diff
@@ -21,24 +21,31 @@ const mainRepoUrl = 'https://github.com/ckeditor/ckeditor5';
 const revisionBranch = `${ branch }-revisions`;
 
 // Clone the repository.
-exec( `git clone -b ${ revisionBranch } ${ mainRepoUrl }.git` );
+exec( `git clone ${ mainRepoUrl }.git` );
+
+// And check out to the revision branch.
+exec( `git checkout ${ revisionBranch } ` );
```
Reinmar: Calling `checkout` outside `ckeditor5`?

packages/ckeditor5-dev-tests/bin/save-revision.js
```diff
@@ -21,24 +21,31 @@ const mainRepoUrl = 'https://github.com/ckeditor/ckeditor5';
...
 // Install Mgit.
 exec( 'npm install -g mgit2' );
 
-// Get the latest `mgit.json`.
-exec( `git checkout origin/${ branch } .` );
+// Sync the revision branch with the master.
+exec( `git checkout ${ branch } -- .` );
+
+// Remove all files from "staged".
+exec( 'git reset -q .' );
```
Reinmar: I think you may sometime need to pass additional options to also clean up old directories.

packages/ckeditor5-dev-tests/bin/save-revision.js
pomek: Yea. It should be done after L30.

packages/ckeditor5-dev-tests/bin/save-revision.js
pomek: Not sure but this command does what you wrote here - https://github.com/ckeditor/ckeditor5-dev/issues/375#issuecomment-376487447.

pomek:
Checked manually. All works fine.

PR #403, opened 2018-04-24
Filtering out unnecessary BS browsers in PRs from community

### LINKED ISSUE #402: BrowserStack's browsers are not disabled for PRs from community
This issue is strongly related to https://github.com/ckeditor/ckeditor5-dev/issues/358.

In #358 we disabled the BrowserStack integration. However, the browsers are not filtered out so even if the BS config will not be loaded, the BrowserStack browsers are still trying to use the BS.

pomek:

Fix: Browsers that require BrowserStack service will not be used in PRs that come from the community. Closes #402.


pomek:
On the occasion, I decided to increase the mocha's timeout for the last time. If it would fail, I'll try to optimize these tests.

PR #407, opened 2018-05-15
Disabled minify of CSS animations in the postCSS config.

### LINKED ISSUE #350: [postCSS] Disable merging keyframes animations
We need to add `reduceIdents: true` option to minified build, to prevent merging animations from separated stylesheets into one.

http://cssnano.co/optimisations/reduceidents/

https://github.com/ckeditor/ckeditor5-dev/blob/7f6063723d77c24110fcac591f184b4b1ab3221a/packages/ckeditor5-dev-utils/lib/styles/getpostcssconfig.js#L35-L40

cc @oleq

dkonopka:

Fix: Disabled minify of CSS Animations (`@keyframes`). Closes #350.

### DIFF
--- a/packages/ckeditor5-dev-utils/lib/styles/getpostcssconfig.js
+++ b/packages/ckeditor5-dev-utils/lib/styles/getpostcssconfig.js
@@ -35,7 +35,8 @@ module.exports = function getPostCssConfig( options = {} ) {
 	if ( options.minify ) {
 		config.plugins.push( require( 'cssnano' )( {
 			preset: 'default',
-			autoprefixer: false
+			autoprefixer: false,
+			reduceIdents: true
 		} ) );
 	}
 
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-dev-utils/lib/styles/getpostcssconfig.js
```diff
@@ -35,7 +35,8 @@ module.exports = function getPostCssConfig( options = {} ) {
 	if ( options.minify ) {
 		config.plugins.push( require( 'cssnano' )( {
 			preset: 'default',
-			autoprefixer: false
+			autoprefixer: false,
+			reduceIdents: true
```
oleq: Shouldn't it be `false`?

dkonopka: of course it should be, my bad!

PR #420, opened 2018-07-09
Updating peerDependencies and support dry-run for release tool

pomek:

Feature: The release tool supports updating the `peerDependencies` in `package.json`. Introduced a `dryRun` mode which allows testing the whole release process without publishing anything. The changelog generator for sub-repositories accepts the `newVersion` parameter. All generated changelog will use the version instead of analyzing a history of commits in a given repository. See ckeditor/ckeditor5#1061.

Dry run mode for release sub-repositories:

  - `npm version` will not create a tag (only the commit will be made),
  - `npm pack` will be called instead of `npm publish` (it packs the whole release to a ZIP archive),
  - `git push` will be replaced with a log on the screen,
  - creating a release on GitHub will be replaced with a log that will contain a URL to the release,
  - every called command will be displayed.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const path = require( 'path' );
+const chalk = require( 'chalk' );
 const { tools, logger } = require( '@ckeditor/ckeditor5-dev-utils' );
 const parseGithubUrl = require( 'parse-github-url' );
 const cli = require( '../utils/cli' );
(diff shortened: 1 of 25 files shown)


packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
```diff
@@ -196,7 +221,8 @@ module.exports = function releaseSubRepositories( options ) {
 			const releaseDetails = packagesToRelease.get( packageJson.name );
 			const errorsForPackage = validatePackageToRelease( {
 				changes: releaseDetails.changes,
-				version: releaseDetails.version
+				version: releaseDetails.version,
+				allowOtherBranch: dryRun === true
```
Reinmar: ?

PR #431, opened 2018-08-06
Split releasing packages for two steps

### LINKED ISSUE #427: Releasing ckeditor5 together with other packages
So far we've been releasing `@ckeditor/ckeditor5-*` first and then, if we had confirmation from CI that it went well, we've been publishing `ckeditor5`.

However, `ckeditor5` became a dependency of `@ckeditor/ckeditor5-utils`, which mean that we have to bump up their versions together. Or, in fact, we could even release `ckeditor5` first (it has no dependencies itself).

Anyway, we need to review our previous release process and make sure that between calling `npm release:dependencies` and `npm release` there are no manual steps so this can be done one after another to finally become a single step.

pomek:

Feature: Releasing packages has been split from single step to two. Both steps contain a `dry-run` mode which allows testing every step of the release process without publishing anything. Closes #427.

For tagging repositories use `tasks.bumpVersions()` which updates version across all packages in the project. Before starting the updating process, the task validates whether all ingredients are defined (mostly whether changelogs were generated).

For publishing changes use `tasks.releaseSubRepositories()` which cares about publishing changes on NPM and/or GitHub. It checks versions of packages published on NPM and GitHub so there is no risk to publish the same changes twice.

Both tasks contain a dry-run mode which allows testing the whole process without a pain about pushing or publishing. The dry-run mode prints every called command on the screen. Instead of publishing package, it creates an archive that contains content which will be published. All commits made by `npm version` (and its hooks like `pre` or `post`) will be removed.

BREAKING CHANGES: `tasks.releaseSubRepositories()` does not updates versions any more. Use it together with `tasks.bumpVersions()`.

BREAKING CHANGES: `cli.configureReleaseOptions()` returns `npm` and `github` keys (and opposite values) instead of `skipNpm` and `skipGithub`

BREAKNG CHANGES: `getSubRepositoriesPaths()` returns an object that contains two keys: `matched` and `skipped`. Was: `packages` and `skipped`.

NOTE: `tasks.generateChangelogForSubRepositories()` accepts `skipMainRepository` option which is passed to `getSubRepositoriesPaths()` util. If `skipMainRepository` option is set on true, the package defined in `options.cwd` will be added as `skipped`, if on false, as `matched`.

---

Usage: https://github.com/ckeditor/ckeditor5/compare/t/ckeditor5-dev/427


pomek:
I dropped tests for these release tasks because it's very hard to write them correctly and in fact, they don't check anything because the real environment is a little bit different. Instead, I focused on dry-run mode which allows testing these tools on the real environment without publishing anything.

pomek:
Using new scripts, releasing CKEditor5 process should look:

1. Generate changelogs for packages (`npm run changelog`)
2. Review changelogs and fix invalid entries.
3. Call `npm run release:bump-version`. It bumps versions of all packages (including the main repository) and their dependencies.
4. It's time for one more review, whether everything was updated correctly. If something went wrong, we can fix it manually.
5. Call `npm run release:publish` in order to publish changes. If something fails, you can repeat the step because of this task based on published versions (on npm and GitHub). There is no risk to publish the same changes twice.
6. All changes are available on GitHub and NPM so we can sync `#stable` branches. Call: `npm run release:stable-branches`.

The question is – should the last step become as the first step? I guess we should merge `#stable` to `#master` **before** releasing and `#master` to `#stable` **after**.

packages/ckeditor5-dev-env/lib/release-tools/utils/cli.js
```diff
@@ -37,6 +38,73 @@ const cli = {
 			.then( answers => answers.confirm );
 	},
 
+	/**
+	 * Asks a user for a confirmation for publishing changes.
+	 *
+	 * @param {Map} packages Packages to release.
+	 * @returns {Promise}
+	 */
+	confirmPublishing( packages ) {
```
pomek: It produces: 

image

Packages for publishing are highlighted, packages with no changes are darker.

packages/ckeditor5-dev-env/lib/release-tools/utils/getpackagestorelease.js
```diff
@@ -37,11 +45,14 @@ module.exports = function getPackagesToRelease( pathsToPackages ) {
...
-	return executeOnPackages( pathsToPackages, filterPackagesToRelease )
-		.then( () => {
-			process.chdir( cwd );
-
-			return Promise.resolve( packagesToRelease );
-		} );
 };
+
+/**
+ * @typedef {Object} ReleaseDetails
```
pomek: TODO: Add properties: `repositoryOwner`, `repositoryName`, `npmVersion`, `npmRelease`, `githubVersion`, `githubRelease`. 

Question to myself: `(npm|github)Release` should be `is(Npm|Github)Release`?

pomek:
One more screen with dry-run mode:

image

packages/ckeditor5-dev-env/lib/release-tools/utils/getpackagestorelease.js
pomek: Done. I went with `shouldReleaseOn(Npm|Github)`.

ma2ciek:
`+1,114 −1,114` ☯️

pomek:
So there is nothing to review 🤣

packages/ckeditor5-dev-env/lib/release-tools/tasks/bumpversions.js
```diff
@@ -0,0 +1,326 @@
...
+
+			return Promise.resolve();
+		} ).then( () => packages );
+	}
+
+	// Validate the repositories.
+	//
+	// @params {Map.<String, ReleaseDetails>} packages
+	// @returns {Promise.<Map.<String, ReleaseDetails>>}
+	function validateRepositories( packages ) {
```
ma2ciek: this function can be rewritten to be synchronous.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bumpversions.js
```diff
@@ -0,0 +1,326 @@
...
+
+			return Promise.resolve();
+		} ).then( () => packages );
+	}
+
+	// Gather descriptions of the release for all packages.
+	//
+	// @params {Map.<String, ReleaseDetails>} packages
+	// @returns {Promise.<Map.<String, ReleaseDetails>>}
+	function getLatestChangesForPackagesThatWillBeReleased( packages ) {
```
ma2ciek: It can be synchronous too.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bumpversions.js
```diff
@@ -0,0 +1,326 @@
...
+		return { packages, dependencies };
+	}
+
+	// Filter out packages which won't be released.
+	//
+	// @params {Map.<String, ReleaseDetails>} packages
+	// @params {Map} dependencies
+	// @returns {Object}
+	function filterPackagesThatWillNotBeReleased( packages, dependencies ) {
+		logDryRun( 'Filtering out packages that will not be updated...' );
```
ma2ciek: Not normal `log()`?

ma2ciek: I mean, you see more information what happens with the `dry run`. I think that such information can be useful in the real scenario too.

packages/ckeditor5-dev-env/lib/release-tools/utils/getpackagestorelease.js
```diff
@@ -10,15 +10,22 @@ const getPackageJson = require( './getpackagejson' );
...
 /**
- * Returns a list of packages to release.
+ * Returns a list of packages which should be releases based on changes in a changelog file and tags created in Git repository.
  *
- * @param {Set} pathsToPackages A collection of paths to packages.
- * @returns {Promise}
+ * @param {Set} pathsToPackages A collection of paths to packages that should be checked.
+ * @returns {Promise.<Map.<String, ReleaseDetails>>}
  */
 module.exports = function getPackagesToRelease( pathsToPackages ) {
```
ma2ciek: This function can be sync as well.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bumpversions.js
```diff
@@ -0,0 +1,326 @@
...
+	// @params {Map.<String, ReleaseDetails>} packages
+	// @returns {Promise.<Map.<String, ReleaseDetails>>}
+	function getLatestChangesForPackagesThatWillBeReleased( packages ) {
+		logProcess( 'Gathering changes from changelog for specified versions...' );
+
+		return executeOnPackages( pathsCollection.matched, repositoryPath => {
+			process.chdir( repositoryPath );
+
+			const packageJson = getPackageJson( repositoryPath );
+			const releaseDetails = packages.get( packageJson.name );
```
ma2ciek: I'd go with a `package` for the variable name.

pomek: There is a lot of `package*` variables. If I leave `package` without any context, it will be a little confusing. I would go with `packageDetails`. However, `releaseDetails` describes perfectly what it is.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bumpversions.js
pomek: A few functions in this code could be synchronous. I made these asynchronous because if we will add some asynchronous actions to do, a body (core) of these functions will not require a rewrite.

They are asynchronous as well because I wanted to show my intention: "execute given action on all packages". This mechanism occurs in the whole script.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bumpversions.js
pomek: From the real release point of view, it does not make any sense because we want to release packages that are ready to release. 

We know which packages will be releasing (because we know where the changelogs were generated) so the log "filtering out packages that will not be updated" does not make any sense because we don't display which packages were filtered out.

pomek:
One more thing that came to my mind. For publishing on NPM, repositories could be checked out to the tag that will be published. Thanks to that, we will be able to generate changelogs, tagging new versions and `master`s branches won't be blocked. 

- tag: `v10.0.0` (last known release)
- ...
- a lot of commits...
- ...
- generating a changelog for the new version: `10.1.0`
- preparing new release - make a tag: `v10.1.0`
- ...
- new commits for the next iteration
- ...
- release day - checkout the repository to the tag `v10.1.0` in order to publish proper code on npm
- after release - check out to `master`

Does it make sense?

PR #435, opened 2018-09-03
Disabled "save-revision" script for pull requests builds

### LINKED ISSUE #434: Saving revisions works for pull request builds
https://travis-ci.org/ckeditor/ckeditor5-media-embed/builds/421948985

image

image

https://github.com/ckeditor/ckeditor5/commit/91370c0

In fact, it does not break anything because hashes are collected from `#master`s but… should we disabled it? Now we're checking only the branch name:

https://github.com/ckeditor/ckeditor5-dev/blob/ffe3a77b9cf02caca074ffa25c4f8d515b0dcc7e/packages/ckeditor5-dev-tests/bin/save-revision.js#L13-L15

We could also check whether `TRAVIS_EVENT_TYPE` is equal to `push`.

pomek:

Fix: Disabled "save-revision" script for pull requests builds. Closes #434.

### DIFF
--- a/packages/ckeditor5-dev-tests/bin/save-revision.js
+++ b/packages/ckeditor5-dev-tests/bin/save-revision.js
@@ -8,9 +8,10 @@
 'use strict';
 
 const branch = process.env.TRAVIS_BRANCH;
+const buildType = process.env.TRAVIS_EVENT_TYPE;
 
-// Save revision only when master branches are updated.
-if ( branch !== 'master' ) {
+// Save revision only when commit has made directly on the "master" branch.
+if ( branch !== 'master' || buildType !== 'push' ) {
 	process.exit();
 }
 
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-dev-tests/bin/save-revision.js
```diff
@@ -63,6 +64,8 @@ if ( exec( 'git status -s' ).trim().length ) {
 
 	const lastCommit = exec( 'git log -1 --format="%h"' );
 	console.log( `Successfully saved the revision under ${ mainRepoUrl }/commit/${ lastCommit }` );
+} else {
+	console.log( 'Nothing to commit.' );
```
Reinmar: Something more explanatory would be better. A simple "nothing to commit" doesn't explain what are you looking at (you will see this message in travis logs). Something like this might be better:

> Nothing to commit. Revision log is up to date.

PR #455, opened 2018-11-23
Introduce custom rule for ESLint to disallow relative imports.

### LINKED ISSUE #454: Introduce custom eslint rules for CKEditor 5
I'm interested in adding a custom rule to ESLint. This rule would inspect for relative imports to other CKEditor5 packages:

```js
// incorrect:
import Position from '../../ckeditor5-engine/src/model/position';

// Fixable to: 
import Position from '@ckeditor/ckeditor5-engine/src/model/position';
```

Why? I'm using PHPStorm feature which automatically adds imports to a file - it does pretty nicely but it adds as a relative path (yay multi-repo). From time to time I commit such file and have to fix it after Travis fails. Dunno if other projects are interested but this check can be further enhanced by adding a configuration for disallowed paths and a rule to fix them. `'../../ckeditor5-' : '@ckeditor/ckeditor5-'`

I've already created a proof-of-concept. Below is a screenshoot from PHPStorm IDE displaying such error:
selection_068

And eslint fixer invoked in action:
peek 2018-11-23 10-58

This would require to create a `eslint-plugin-ckeditor5-rules` npm package and add to ckeditor5 configuration (`.eslintrc.js`):

```js
module.exports = {
	// ...
	plugins: [
		'ckeditor5-rules'
	],
	rules: {
		// ...
		'ckeditor5-rules/no-relative-imports': 'error'
	}
}
```

jodator:

Feature: Introduce Introduce custom rules for ESLint. Closes #454.

---

### Additional information

* WiP - feature preview


packages/eslint-plugin-ckeditor5-rules/lib/rules/no-relative-imports.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+module.exports = {
+	meta: {
+		type: 'problem',
+
+		docs: {
+			description: 'Disallow relative imports from CKEditor5 packages',
```
pomek: Missing dot at the end of the sentence.

packages/eslint-plugin-ckeditor5-rules/lib/rules/no-relative-imports.js
```diff
@@ -0,0 +1,35 @@
...
+			category: 'CKEditor5'
+		},
+		fixable: 'code',
+		schema: [] // no options
```
pomek:
```suggestion
		schema: []
```

This comment is so obvious and can be removed. We can see without the comment that there are no options. Or if it should stay, let's clarify why there are no options.

packages/eslint-plugin-ckeditor5-rules/lib/rules/no-relative-imports.js
```diff
@@ -0,0 +1,35 @@
...
+	},
+	create( context ) {
+		return {
+			ImportDeclaration: node => {
+				const importPath = node.source.value;
+
+				if ( importPath.startsWith( '../../ckeditor5-' ) ) {
```
pomek: What if an import will start with `../../../*`? There should be a smarter solution to check it.

ma2ciek: ☝️ This will probably happen in the CF repository where a path looks like `packages/@ckeditor/ckeditor5-engine/...`.

pomek: The second case – this replace will fail in a package that uses relative imports for importing other files from the same package (but they are nested):

```
ckeditor5-bar/src/
  |-- foo/
  |-- |-- ...
  |-- bar/
  |-- |-- foo/
  |-- |-- |-- nested/
  |-- |-- |-- |-- ...
```

Imports from `foo/*` inside `bar/foo/nested/` will require `../../foo/*` which trigger the plugin. So we need to ensure that relative imports ([even `../../../` and `../../../../`](https://github.com/ckeditor/ckeditor5-ui/blob/9aaad899d81828de6d44282f81ad2e2b4f307282/tests/manual/panel/sticky/stickypanelview.js#L6-L7)) inside single package won't be replaced.

jodator: They won't since: `/ckeditort-` is checked also not a relative path per se.

jodator:
> The second issue is a lack of the tests. I know that testing plugins to external tools is really hard.

Nah. The ESLint has nice rule tester which I've used.

@pomek I've cleaned up the PR. I've commented the `.eslintrc.js` changes due to the fact that the CI was blowing up - missing NPM package for this plugin.

I think that we should enable this in two steps:
1. merge this PR and publish NPM package
2. incorporate this plugin to the `eslint-config-ckeditor5`.

pomek:
And of course, CI will fail because I enabled this plugin.

jodator:
It look fine - this plugin is rather simple so short readme is enough :)

PR #477, opened 2019-01-21
Prepared a new testing env. on CI

### LINKED ISSUE #471: Change the way how tests are run for single packages
The [circular dependency of `ckeditor5-paragraph` and `ckeditor5-heading`](https://github.com/ckeditor/ckeditor5-dev/issues/470) showed that we shouldn't be running tests as we do now. The problem is that we run them differently then we do on development environment.

Current workflow:
1. Travis checkouts github repository
2. We bootstrap testing environment:
  - `yarn add ckeditor5-dev-tests`
  - `yarn add mgit` (second yarn add)
  - create `mgit.json` with immediate packages dependencies
  - run `mgit sync` which add all `@ckeditor` dependencies into `./packages` dir
  - run `yarn install`
3. We run test in main package

This worked well with `npm` + `lerna` but fails with `yarn` workspaces. Yarn doesn't symlink main package to the `node_modules` dir and thus we have duplicated dependencies when one of packages requires package that is being tested. (circular deps - mostly in the devDependencies).

After short F2F talk with @Reinmar we think that we should run our testing environment for single packages the same way as we run for all packages and as we do in dev environment.

In short we should create a workspace in which current package is located (as in `ckeditor5` with all it's dependencies) and then run test only for this package.

ps.: Also we should merge `yarn add` calls to a minimum as that doubled `yarn add` isn't necessary and we could merge them into one step.

pomek:

Feature: Rewrote a script that prepares the testing environment on CI. Now it creates a ckeditor5-like project that contains a proper version of all dependencies. Closes #471.

---

Testing build: https://travis-ci.org/ckeditor/ckeditor5-paragraph/builds/482415664


oleq:
The execution time increased noticeably in this PR and this kinda worries me.

screenshot 2019-01-23 at 15 11 27
screenshot 2019-01-23 at 15 12 00

pomek:
I am not surprised because we need to call `yarn install` twice.

jodator:
@pomek I'm thinking about minimizing the `yarn add/install` steps.

The job on paragraph repo has many of those :) I assume that the step with cloning & building ckeditor5-dev-test there is only for PR but we could try to lower `yarn add/install` from 3 to 2:

1. In `$PACKAGE` dir execute `yarn add ckeditor5-dev-test mgit2` - this will install all the required things for tests & linter (1st call)
2. Then use scripts from `PACKAGE_ROOT/node_modules/.bin/` in `TEST_DIR` for creating package.json & mgit.json, etc
3. Use `mgit` from `PACKAGE_ROOT/node_modules/.bin/` to bootstrap mgit packages
4. Run `yarn install` in `TEST_DIR` (2nd call) to bootstrap test environment

Will this work?

This will probably not reduce the time from ~5:30 to ~5:00 as in @oleq's comment because we need to clone & bootstrap the environment two times: for linting/dep check and for the tests.

pomek:
AFAICS you changed the directory where the `mgit` is being installed. You didn't reduce the number of calling `yarn add/install`. Am I wrong?

However, `yarn add` only adds a dependency. It does not install the rest. For _linting/dep check_ we could install missing dependencies manually (I mean `yarn add eslint ...`) but if we will introduce some improvement for the linter and it will require a new dep, CI will fail because we forgot about to modify this `install` step.

I'm also not worried about the time (4:57 / 5:30) because, after a few builds, yarn should have some cache that could improve the speed of bootstrapping the entire environment.

jodator:
> AFAICS you changed the directory where the `mgit` is being installed. You didn't reduce the number of calling `yarn add/install`. Am I wrong?

3 (add install install) instead of 2 (add install)

> However, `yarn add` only adds a dependency. It does not install the rest.

AFAICS it adds the package and install not-instlled dependencies. Tested with:
```sh
git clone git@github.com:ckeditor/ckeditor5-dev.git
cd ckeditor5-dev
yarn add jquery
```
output:
```
~/tmp/ckeditor5-dev   
13:14:19 
╰─ yarn add jquery
yarn add v1.13.0
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
[5/5] Building fresh packages...
success Saved 1 new dependency.
info Direct dependencies
└─ jquery@3.3.1
info All dependencies
└─ jquery@3.3.1
Done in 2.65s.
~/tmp/ckeditor5-dev
13:14:27 
╰─ yarn install
yarn install v1.13.0
[1/5] Validating package.json...
[2/5] Resolving packages...
success Already up-to-date.
Done in 0.42s.
```
and installed modules count:
```
╰─ ls -la node_modules | wc -l
553
```

So `yarn add` and then `yarn install` will give the same result but again small since "everything is up-to-date".

pomek:
> AFAICS it adds the package and install not-instlled dependencies. Tested with:

You're right. So should I remove this part of the installing-deps script:

```sh
# Dependencies for tested package.
cd $PACKAGE_ROOT && yarn install
```

Because they should be already installed (when Travis called `yarn add ckeditor5-dev-tests in `before_scripts` hook).

jodator:
I think not in current script setup because you have:

1. `yarn add mgit2` in `TEST`
2. `mgit sync` in `TEST` (populates `/packages` in test)
3. `yarn install` in `TEST` (will build workspace in test)
4. `yarn install` in `PACKAGE`

If we'd move `yarn add mgit` to the `PACKAGE` dir and use it from there then I think you `yarn install` in TEST package would be enough.

So then it will be:

1. `yarn add mgit` in the `PACKAGE` dir
2. `$(PACKAGE_DIR)/node_modules/.bin/mgit2 sync` (pseudo code)
3. `yarn install` in the `TEST` dir

jodator:
@pomek OMG I'm lost in the PACKAGE/TEST dirs :( probably yes it looks like safe removeal but I don't know if `PACKAGE_ROOT/package.json` was not modified somehow.

pomek:
> I'm lost in the PACKAGE/TEST dirs

It's hard to understand even for me…

pomek:
> If we'd move `yarn add mgit` to the `PACKAGE` dir and use it from there then I think you `yarn install` in TEST package would be enough.

It means that you need to modify 40x `.travis.yml`. The goal when we created those scripts was having one command for installing everything and one command for executing everything. 

https://github.com/ckeditor/ckeditor5-engine/blob/c9816b3980c1c7315a1e24d7421e70060267bdd4/.travis.yml#L22-L28

Because mgit is a part of the install script, I would like to keep it inside (in `ckeditor5-dev-tests` scripts) instead of moving it outside (to `.travis.yml`).

jodator:
> It means that you need to modify 40x `.travis.yml`. The goal when we created those scripts was having one command for installing everything and one command for executing everything.

Actually no. I've meant to: do not install it in `TEST_DIR` and install it to `PACKAGE_DIR` in the `instal-deps` scripts:

now:
```sh
# Install Mgit.
cd ${TEST_DIR} && yarn add mgit2 --ignore-workspace-root-check

# Clones repositories to `packages/` directory.
$NPM_BIN/mgit sync ...

# Dependencies for CKEditor5-like project.
cd ${TEST_DIR} && yarn install

# Dependencies for tested package.
cd $PACKAGE_ROOT && yarn install
```
proposed:
```sh
# Install Mgit.
cd ${PACKAGE_DIR} && yarn add mgit2 --ignore-workspace-root-check

# Clones repositories to `packages/` directory.
${PACKAGE_DIR/node_modules/.bin/mgit sync ...

# Dependencies for CKEditor5-like project.
cd ${TEST_DIR} && yarn install

# Dependencies for tested package.
# not neded: cd $PACKAGE_ROOT && yarn install
```

 But modifying the `travis.yml` would be even better :P:

```yml
install:
- yarn add @ckeditor/ckeditor5-dev-tests mgit2
- ckeditor5-dev-tests-install-dependencies
script:
- ckeditor5-dev-tests-travis
```
and then the install-dependencies can only:

```yml
# Install Mgit - not needed

# Clones repositories to `packages/` directory.
${PACKAGE_DIR/node_modules/.bin/mgit sync...

# Dependencies for CKEditor5-like project.
cd ${TEST_DIR} && yarn install
```

now instead of total 4 yarn install/add we could have only 2.

pomek:
```sh
# Dependencies for tested package.
# not neded: cd $PACKAGE_ROOT && yarn install
```

I guess it's not required anymore because we figured out that `yarn add` install missing packages. I will check it tomorrow.

pomek:
For the question from the last commit: [`Is the command required?`](/ckeditor/ckeditor5-dev/pull/477/commits/819bbacd7e5480393d973a81b20e8649da7fd43f)

The answer is [no, it does not](https://travis-ci.org/ckeditor/ckeditor5-paragraph/builds/489911750).

> now instead of total 4 yarn install/add we could have only 2.

As I said before:

> The goal when we created those scripts was having one command for installing everything and one command for executing everything.

If we can avoid modifying the `.travis.yml`, I really would like to do that. 

Btw, I just stumbled upon on the ticket in yarn repository that described that `yarn add --ignore-workspace-root-check (-W)`does not install missing dependencies.  It explains why I said that it does not install other dependencies. Dependencies aren't installed when you're adding a new dep to package with workspaces. https://github.com/yarnpkg/yarn/issues/4513.

Reinmar:
What's the status here?

Reinmar:
BTW, I'm leaving it for you @jodator to merge. I won't be able to look into this.

jodator:
We stopped at do want 4/2 yarn install/adds commands in repos. We could have 2 but that would require to change 20+ travis.yml files. I'll lookt into it and probably merge as current solution works and I'm not 100% sure if minimising installs will decrease build times significantly.

Reinmar:
At the current stage we don't have to really optimise CI times. If it's not a matter of adding/removing 1 minute, I'd be ok with 4 yarn installs.

jodator:
So the changes are OK IMHO. If we ever came to a conclusion that those times are too long then we have a room for an improvement.

:tada: :clap:

PR #483, opened 2019-02-08
[JSDoc] [WIP] Fixed inheritance system, filtered out internal doclets

### LINKED ISSUE #465: Mixin overrides method of class documentation
Class doclets can get `ignore: true` property. What results with wrong documentation, where mixin description is used.
Description for this method is used from mixin:
https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_controller_datacontroller-DataController.html#function-set
instead of codebase:
https://github.com/ckeditor/ckeditor5-engine/blob/7f40831bac6bdc9b583dd2e71bc5699e21a92365/src/controller/datacontroller.js#L210-L222

related to:
https://github.com/cksource/umberto/issues/702

Problematic doclets:
```json
[
{
    "comment": "/**\n\t * Sets input data parsed by the {@link module:engine/controller/datacontroller~DataController#processor data  processor} and\n\t * converted by the {@link module:engine/controller/datacontroller~DataController#upcastDispatcher view-to-model  converters}.\n\t * This method can be used any time to replace existing editor data by the new one without clearing the\n\t * {@link module:engine/model/document~Document#history document history}.\n\t *\n\t * This method also creates a batch with all the changes applied. If all you need is to parse data, use\n\t * the {@link module:engine/controller/datacontroller~DataController#parse} method.\n\t *\n\t * @param {String} data Input data.\n\t * @param {String} [rootName='main'] Root name.\n\t */",
    "meta": {
        "range": [
            8820,
            9240
        ],
        "filename": "datacontroller.js",
        "lineno": 222,
        "path": "/Users/msamsel/work/bigbang-docs.ckeditor.com/projects/ckeditor5/packages/ckeditor5-engine/src/controller",
        "code": {
            "id": "astnode100045244",
            "name": "DataController#set",
            "type": "MethodDefinition",
            "paramnames": [
                "data",
                "rootName"
            ]
        },
        "vars": {
            "": null
        }
    },
    "description": "<p>Sets input data parsed by the {@link module:engine/controller/datacontroller~DataController#processor data  processor} and\nconverted by the {@link module:engine/controller/datacontroller~DataController#upcastDispatcher view-to-model  converters}.\nThis method can be used any time to replace existing editor data by the new one without clearing the\n{@link module:engine/model/document~Document#history document history}.</p>\n<p>This method also creates a batch with all the changes applied. If all you need is to parse data, use\nthe {@link module:engine/controller/datacontroller~DataController#parse} method.</p>",
    "params": [
        {
            "type": {
                "names": [
                    "String"
                ]
            },
            "description": "<p>Input data.</p>",
            "name": "data"
        },
        {
            "type": {
                "names": [
                    "String"
                ]
            },
            "optional": true,
            "defaultvalue": "'main'",
            "description": "<p>Root name.</p>",
            "name": "rootName"
        }
    ],
    "name": "set",
    "longname": "module:engine/controller/datacontroller~DataController#set",
    "kind": "function",
    "memberof": "module:engine/controller/datacontroller~DataController",
    "scope": "instance",
    "ignore": true
},
{
    "comment": "/**\n\t * @inheritDoc\n\t */",
    "meta": {
        "range": [
            1210,
            3153
        ],
        "filename": "observablemixin.js",
        "lineno": 35,
        "path": "/Users/msamsel/work/bigbang-docs.ckeditor.com/projects/ckeditor5/packages/ckeditor5-utils/src",
        "code": {
            "id": "astnode100041074",
            "name": "set",
            "type": "FunctionExpression"
        },
        "vars": {
            "": null,
            "properties": "module:utils/observablemixin~ObservableMixin.set~properties",
            "this[undefined]": "module:utils/observablemixin~ObservableMixin.set#[undefined]"
        }
    },
    "inheritdoc": "",
    "name": "set",
    "longname": "module:engine/controller/datacontroller~DataController#set",
    "kind": "function",
    "memberof": "module:engine/controller/datacontroller~DataController",
    "scope": "instance",
    "mixes": [
        "module:utils/observablemixin~ObservableMixin.set"
    ],
    "mixed": true,
    "ignore": true
},
{
    "comment": "/**\n * Creates and sets the value of an observable property of this object. Such an property becomes a part\n * of the state and is be observable.\n *\n * It accepts also a single object literal containing key/value pairs with properties to be set.\n *\n * This method throws the `observable-set-cannot-override` error if the observable instance already\n * have a property with the given property name. This prevents from mistakenly overriding existing\n * properties and methods, but means that `foo.set( 'bar', 1 )` may be slightly slower than `foo.bar = 1`.\n *\n * @method #set\n * @param {String|Object} name The property's name or object with `name=>value` pairs.\n * @param {*} [value] The property's value (if `name` was passed in the first parameter).\n */",
    "meta": {
        "range": [
            23495,
            24249
        ],
        "filename": "observablemixin.js",
        "lineno": 727,
        "path": "/Users/msamsel/work/bigbang-docs.ckeditor.com/projects/ckeditor5/packages/ckeditor5-utils/src",
        "code": {}
    },
    "description": "<p>Creates and sets the value of an observable property of this object. Such an property becomes a part\nof the state and is be observable.</p>\n<p>It accepts also a single object literal containing key/value pairs with properties to be set.</p>\n<p>This method throws the <code>observable-set-cannot-override</code> error if the observable instance already\nhave a property with the given property name. This prevents from mistakenly overriding existing\nproperties and methods, but means that <code>foo.set( 'bar', 1 )</code> may be slightly slower than <code>foo.bar = 1</code>.</p>",
    "kind": "function",
    "name": "set",
    "params": [
        {
            "type": {
                "names": [
                    "String",
                    "Object"
                ]
            },
            "description": "<p>The property's name or object with <code>name=&gt;value</code> pairs.</p>",
            "name": "name"
        },
        {
            "type": {
                "names": [
                    "*"
                ]
            },
            "optional": true,
            "description": "<p>The property's value (if <code>name</code> was passed in the first parameter).</p>",
            "name": "value"
        }
    ],
    "longname": "module:engine/controller/datacontroller~DataController#set",
    "scope": "instance",
    "memberof": "module:engine/controller/datacontroller~DataController"
}
]
```

ma2ciek:

Fix: Fixed inheritance mechanism and filtered out internal doclets. Closes #465.

---

### Additional information

* [x] Tests covering that changes (all old tests are passing now)
* [x] Improve API docs and add some internal comments

How to test:
1. Go to `ckeditor5`, ensure that all is up to date and run `yarn run switch-to-dev-dev`.
2. In `ckeditor5-dev` check out the `t/465`.
3. In `ckeditor5` run `yarn docs --skip-snippets [--skip-validation]`.

I've added TS checked API docs, it helped me much to understand which variable is which type.


ma2ciek:
@Mgsy, I'd be happy if you could look at a few more complicated classes in docs and check whether the documentation is correct there.

ma2ciek:
Since it touches `umberto` and `ckeditor5-dev` projects, cc @pomek @msamsel. Not sure if you want to review this part of the code but the bus factor here is pretty low, so I encourage you 😄

packages/jsdoc-plugins/lib/doclet.js
```diff
@@ -0,0 +1,38 @@
+/**
```
pomek: The extension of this file should be `jsdoc`. See: https://github.com/ckeditor/ckeditor5-engine/blob/75397b2e54f2ef3e4fe153b6dae0c72417d3ef5f/src/dataprocessor/dataprocessor.jsdoc.

packages/jsdoc-plugins/lib/doclet.js
```diff
@@ -0,0 +1,38 @@
...
+ * Doclets are atomic parts of the JSDoc output that describe each symbol of parsed input.
+ *
+ * @typedef {Object} Doclet
```
pomek: In order to increase the readability, I would like to add a blank line before `@property` tag.

packages/jsdoc-plugins/lib/doclet.js
```diff
@@ -0,0 +1,38 @@
...
+ * @property {String} longname Long name of the doclet (including the `module:` part).
+ * @property {String} name Short name of the doclet (e.g. the name of the method).
+ * @property {String} memberof Where the doclet belongs to (parent of the symbol).
+ * @property {'class'|'interface'|'mixin'|'function'|'typedef'|'event'|'member'|'constant'|'module'} kind The kind of the doclet's symbol.
+ * @property {Boolean} [ignore] `true` for internal doclets which should not be published
+ * @property {Boolean} [undocumented] `true` when a doclet's symbol does not have API docs written above the declaration.
+ * @property {String} [inheritdoc] Warning: When the `@inheritdoc` is present, then in most cases the property
+ * becomes  an empty string!
```
pomek: Why does it become to be an empty string? It should be explained. I was reading it a few times and I don't get it.

packages/jsdoc-plugins/lib/doclet.js
```diff
@@ -0,0 +1,38 @@
...
+ * @property {Boolean} [overrides]
+ * @property {{inherited?: Boolean, name: String, type: Object, description: string}[]} [properties] Typedef properties.
```
pomek: We should define one more definition here.

packages/jsdoc-plugins/lib/doclet.js
```diff
@@ -0,0 +1,38 @@
...
+ * @property {Boolean} [inherited] `true` for a property / method which is inherited from the parent class.
+ * @property {String} [inherits] The longname of the parent class / interface method.
+ * @property {String} [comment] Raw API comment.
+ * @property {String} [description] API Comment wrapped in HTML tags.
+ * @property {Object[]} [params] Event and function params.
```
pomek: `Array.<Object>`

pomek: Or even another custom type.

packages/jsdoc-plugins/lib/doclet.js
```diff
@@ -0,0 +1,38 @@
...
+ * @property {'instance'|'inner'|'static'} [scope]
+ * @property {String[]} [fires] An array of events that a method or a property can fire.
```
pomek: `Array.<String>`

pomek: And few of the same below.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -2,29 +2,33 @@
  * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
  * Licensed under the terms of the MIT License (see LICENSE.md).
  */
+// @ts-check
```
pomek: I don't understand what it does here. It occurs 4 times in this PR.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -2,29 +2,33 @@
...
+function addMissingDoclets( doclets ) {
+	doclets = cloneDeep( doclets );
+
 	const docletCollection = new DocletCollection();
+
+	/** @type {Doclet[]} */
 	const typedefDoclets = [];
 
-	for ( const doclet of clonedDoclets ) {
+	for ( const doclet of doclets ) {
```
pomek: Would be good to avoid overwriting the function's parameters. I mean the change in L24 for me is invalid.

If you would like to simplify the thing, you could write something like this:

`for ( const doclet of cloneDeep( doclets ) ) {`

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -2,29 +2,33 @@
...
  * @returns {Array.<Doclet>}
  */
-function addMissingDoclets( originalDoclets ) {
-	const clonedDoclets = cloneDeep( originalDoclets );
+function addMissingDoclets( doclets ) {
+	doclets = cloneDeep( doclets );
+
 	const docletCollection = new DocletCollection();
+
+	/** @type {Doclet[]} */
```
pomek: Invalid type.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -33,11 +37,25 @@ function addMissingDoclets( originalDoclets ) {
...
-	const entitiesWhichNeedNewDoclets = clonedDoclets.filter( d => {
-		return d.kind === 'class' || d.kind === 'interface' || d.kind === 'mixin';
+	extendTypedefs( typedefDoclets );
+
+	const extensibleDoclets = doclets.filter( doclet => {
+		return (
+			doclet.kind === 'class' ||
+			doclet.kind === 'interface' ||
+			doclet.kind === 'mixin'
+		);
```
pomek: Parentheses here are useless.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -33,11 +37,25 @@ function addMissingDoclets( originalDoclets ) {
...
 	} );
+
+	/** @type {Doclet[]} */
 	const newDocletsToAdd = [];
+
+	/** @type {Doclet[]} */
 	const docletsToIgnore = [];
+
+	/**
+	 * @type {Array<{relation: 'augmentsNested'|'mixesNested'|'implementsNested';filter?:Object;onlyImplicitlyInherited?:Boolean}>}
```
pomek: What?

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -80,23 +98,28 @@ function addMissingDoclets( originalDoclets ) {
...
 	}
 
-	docletsToIgnore.forEach( d => {
-		d.ignore = true;
-	} );
-	clonedDoclets.push( ...newDocletsToAdd );
-
-	extendTypedefs( typedefDoclets );
+	// Ignore doclets that shouldn't be used anymore. They will be removed afterward.
+	for ( const docletToIgnore of docletsToIgnore ) {
```
pomek: `s/docletToIgnore/doclet`?

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -67,11 +85,11 @@ function addMissingDoclets( originalDoclets ) {
 		}
 	];
 
-	for ( const childDoclet of entitiesWhichNeedNewDoclets ) {
+	for ( const extensibleDoclet of extensibleDoclets ) {
```
pomek: `s/extensibleDoclet/doclet`?

packages/jsdoc-plugins/lib/relation-fixer/buildrelations.js
```diff
@@ -2,132 +2,163 @@
...
 
 	return resultRelations;
 }
 
-function isNonEmptyArray( obj ) {
-	return Array.isArray( obj ) && obj.length > 0;
+/**
+ * Returns `true` when the input in equal to `undefined` or is an empty array.
+ *
+ * @param {Array.<any>|undefined} arr
```
pomek: `s/Array.<any>/Array`

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
```diff
@@ -62,57 +72,62 @@ function getMissingDocletsData( docletCollection, childDoclet, options ) {
...
+
+	return docletToAdd;
 }
 
 function isNonEmptyArray( obj ) {
 	return Array.isArray( obj ) && obj.length > 0;
 }
 
+/**
+ * @param {Readonly<Doclet>} parentDoclet
```
pomek: What is the `Readonly<Doclet>`?

ma2ciek:
Thanks for the detailed review @pomek 😂 

Please, install the TS extension and see how it works, even if you find these types of annotation unfriendly at first glance.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: But they increase the readability IMO.

packages/jsdoc-plugins/lib/doclet.js
ma2ciek: Because when JSDoc sees the `@inheritDoc` tag, then this property becomes `""`. Otherwise, it becomes undefined. Most probably due to the fact that after the `@inheritDoc` tag you should be able to type more docs for that specific symbol.

pomek: When I wrote: "It should be explained.", I meant: "an answer for the asked question could expand the description ".

pomek:
> Please, install the TS extension and see how it works

Why should I do it? Do we support it? I guess that we don't so I was writing everything that is invalid to our rules and terms how we prepare the code.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: IMO overriding parameters is totally fine if you want to be sure that you won't make changes to the original parameters.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: It's more precise.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: It's more precise.

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
ma2ciek: Readonly<T> is an object of type T that can't be modified. But if you wish I can remove that.

ma2ciek:
Unfortunately, integration tests won't be checked by the CC.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
pomek: I don't agree. Overriding parameters always leads to unexpected results. It may not happen in this situation but we should avoid such danger.

packages/jsdoc-plugins/lib/doclet.jsdoc
```diff
@@ -0,0 +1,52 @@
...
+ * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
+ * Licensed under the terms of the MIT License (see LICENSE.md).
+ */
+
+/**
+ * Doclets are atomic parts of the JSDoc output that describe each symbol of parsed input.
+ *
+ * @typedef {Object} Doclet
+ *
+ * @property {String} longname Long name of the doclet (including the `module:` part).
```
pomek: Please, add a blank line between each `@property` tag.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: why?

ma2ciek: what bad can happen?

packages/jsdoc-plugins/lib/doclet.jsdoc
```diff
@@ -0,0 +1,52 @@
...
+ * @property {'instance'|'inner'|'static'} [scope]
+ * @property {Array.<String>} [fires] An array of events that a method or a property can fire.
+ * @property {Object} meta Doclet's metadata - filename, line number, etc.
+ * @property {Array.<String>} [augments] An array of classes that the doclet's symbol extends.
+ * Applies for `@class`, `@mixin`, `@interface`.
+ * @property {Array.<String>} [mixes] An array of mixins that the doclet's symbol mixes.
+ * Applies for `@class` and `@mixin`.
+ * @property {Array.<String>} [implements] An array of interfaces that the doclet's symbol implements.
+ * Applies for `@class` and `@mixin`.
+ * @property {Array.<String>} [augmentsNested] [A custom property used by the relation fixer] -
```
pomek: "Square brackets" for the description are useless. See properties below as well.

ma2ciek: I used them to group these props with additional information.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -33,12 +37,26 @@ function addMissingDoclets( originalDoclets ) {
 		}
 	}
 
-	const entitiesWhichNeedNewDoclets = clonedDoclets.filter( d => {
-		return d.kind === 'class' || d.kind === 'interface' || d.kind === 'mixin';
+	extendTypedefs( typedefDoclets );
```
pomek: Side effect. You should pass the `typedefDoclets` object and expect that this function will return a new object.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -33,12 +37,26 @@ function addMissingDoclets( originalDoclets ) {
...
+
+	/** @type {Array.<Doclet>} */
 	const newDocletsToAdd = [];
+
+	/** @type {Array.<Doclet>} */
 	const docletsToIgnore = [];
-	const settings = [
+
+	/**
+	 * @type {Array.<import('./getmissingdocletsdata').Options>}
```
pomek: [Invalid type.](http://usejsdoc.org/tags-type.html#syntax)

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
pomek: And it's really hard to see difference between `extensibleDoclet` and `extensibleDoclets`.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
```diff
@@ -67,36 +85,41 @@ function addMissingDoclets( originalDoclets ) {
...
+	];
 }
 
-// Copy properties from parent typedefs to typedefs which extend them.
+/**
+ * Copy properties from parent typedefs to typedefs which extend them.
+ *
+ * @param {Array.<Doclet>} typedefDoclets
+ */
 function extendTypedefs( typedefDoclets ) {
```
pomek: As I said https://github.com/ckeditor/ckeditor5-dev/pull/483/files#r256765868 – this function should return a new object instead of modifying passed parameters. Even if it was done earlier, now we are changing this part of code and this bug should be fixed if it isn't too hard.

packages/jsdoc-plugins/lib/relation-fixer/buildrelations.js
```diff
@@ -2,132 +2,163 @@
...
-// @param {Array} options.relations Array of relation names which should be used
-// @returns {Object} An object containing arrays of ancestors' longnames
+/**
+ * Gets long names of the current doclet ancestors (classes it extends, interfaces it implements and so on).
+ *
+ * @param {DocletCollection} docletCollection Doclets grouped by their longnames
+ * @param {Doclet} currentDoclet
+ * @param {Object} options
+ * @param {Array.<'augments'|'implements'|'mixes'>} options.relations An array of relation names which should be used.
+ * @returns {{ [relation: string ]: Array.<String>}} An object containing arrays of ancestors' longnames.
```
pomek: Just `@returns {Object} ...`

packages/jsdoc-plugins/lib/relation-fixer/filteroutintenraldoclets.js
```diff
@@ -0,0 +1,36 @@
...
+		.filter( doclet => filterOutReExportedSymbols( doclet, doclets ) );
+};
+
+/**
+ * @param {Doclet} doclet
+ * @param {Array.<Doclet>} doclets
+ */
+function filterOutReExportedSymbols( doclet, doclets ) {
+	// Filter out exported symbols that are defined previously (JSDoc creates doclets for both).
+	const isUnwanted = ( doclet.scope == 'inner' && doclet.kind == 'constant' && doclet.undocumented );
```
pomek: Useless `(` and `)`.

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
```diff
@@ -62,57 +68,60 @@ function getMissingDocletsData( docletCollection, childDoclet, options ) {
...
 
 	// Longnames of doclets which are related ( extended, mixed, implemented ) to childDoclet.
 	const ancestors = childDoclet[ options.relation ];
 
-	return ancestors.reduce( ( docletsToAdd, longname ) => {
-		const toAdd = docletCollection.get( `memberof:${ longname }` ).filter( d => {
-			let isMatchingFilterOptions = true;
+	const docletToAdd = ancestors.reduce( ( /** @type {Array.<Doclet>} */ docletsToAdd, longname ) => {
+		const toAdd = docletCollection.get( `memberof:${ longname }` ).filter( doclet => {
+			let matchingFilterOptions = true;
```
pomek: This variable can be removed. Following codes could be replaced:

* `matchingFilterOptions = false` => `return false`
* `return matchingFilterOptions`=> `return true`

packages/jsdoc-plugins/lib/relation-fixer/buildrelations.js
ma2ciek: It's more precise

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
```diff
@@ -62,57 +68,60 @@ function getMissingDocletsData( docletCollection, childDoclet, options ) {
...
 		return [];
 	}
 
 	// Longnames of doclets which are related ( extended, mixed, implemented ) to childDoclet.
 	const ancestors = childDoclet[ options.relation ];
 
-	return ancestors.reduce( ( docletsToAdd, longname ) => {
-		const toAdd = docletCollection.get( `memberof:${ longname }` ).filter( d => {
-			let isMatchingFilterOptions = true;
+	const docletToAdd = ancestors.reduce( ( /** @type {Array.<Doclet>} */ docletsToAdd, longname ) => {
```
pomek: I've never seen this syntax in our code.

packages/jsdoc-plugins/lib/relation-fixer/filteroutintenraldoclets.js
ma2ciek: It's a common practice to indicate that the value is a boolean AFAIK.

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
```diff
@@ -62,57 +68,60 @@ function getMissingDocletsData( docletCollection, childDoclet, options ) {
...
 }
 
 function isNonEmptyArray( obj ) {
 	return Array.isArray( obj ) && obj.length > 0;
 }
 
+/**
+ * @param {Doclet} parentDoclet
+ * @param {Doclet} childDoclet
+ */
```
pomek: Missing `@returns`.

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
```diff
@@ -164,12 +181,47 @@ function getRelationProperty( docletMap, childDoclet, memberDoclet, relation ) {
 	}
 }
 
-function checkIfExplicitlyInherits( doclets ) {
+/**
+ * @param {Array.<Doclet>} doclets
+ */
+function doAllParentsExplicitlyInherit( doclets ) {
```
pomek: Body of this function could be simplified:

```js
function doAllParentsExplicitlyInherit( doclets ) {
	return doclets.every( doclet => doclet.inheritdoc !== undefined || doclet.overrides !== undefined );
}
```

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
```diff
@@ -164,12 +181,47 @@ function getRelationProperty( docletMap, childDoclet, memberDoclet, relation ) {
...
+/**
+ * @typedef {Object} Options
+ *
+ * @property {'augmentsNested'|'mixesNested'|'implementsNested'} relation Name of relation between child entity
+ * and its ancestors.
+ * @property {Partial<Doclet>} [filter] Object used to filter missing doclets (e.g. { scope: 'static' }).
+ * @property {Boolean} [onlyImplicitlyInherited]
+ */
+
+/** @typedef {{ [longname: string]: Doclet}} DocletMap */
```
pomek: Invalid type definition.

packages/jsdoc-plugins/lib/relation-fixer/getmissingdocletsdata.js
```diff
@@ -164,12 +181,47 @@ function getRelationProperty( docletMap, childDoclet, memberDoclet, relation ) {
...
+
+/** @typedef {import('../utils/doclet-collection')} DocletCollection */
```
pomek: I guess it isn't necessary.

packages/jsdoc-plugins/lib/relation-fixer/buildrelations.js
pomek: "more precise" !== proper definition in terms of our code style. We should use `Object` or create another custom type.

packages/jsdoc-plugins/lib/relation-fixer/filteroutintenraldoclets.js
pomek: Not in JS or at least not in our code style:

- https://github.com/ckeditor/ckeditor5-list/blob/a3122af2cfa493b0bb89d8737e24bc504fcf1071/src/listediting.js#L148
- https://github.com/ckeditor/ckeditor5-typing/blob/e553afb1592c66a45f0964ccb41d6f1672ad3702/src/utils/injectunsafekeystrokeshandling.js#L48
- https://github.com/ckeditor/ckeditor5-typing/blob/e553afb1592c66a45f0964ccb41d6f1672ad3702/src/deletecommand.js#L151
- https://github.com/ckeditor/ckeditor5-list/blob/e3b3ff3856077ee07cbef2ec14df9e3a8150fd2b/src/converters.js#L352

packages/jsdoc-plugins/lib/doclet.jsdoc
```diff
@@ -0,0 +1,78 @@
...
+ * an array of all interfaces in the inheritance chain, which the current doclet's symbol implements.
+ *
+ * @property {Array.<String>} [descendants] [A custom  property used by the relation fixer] -
+ * an array of doclets which inherits / implements / mixes the doclet's symbol.
+ */
+
+/**
+ * @typedef {Object} Property
+ *
+ * @property {String } name
```
pomek:
```suggestion
 * @property {String} name
```

ma2ciek:
@pomek, have you checked at all whether this PR fixes the mentioned bug?

pomek:
Yes, it does. Unfortunately, after switching to dev versions of our packages, my console shows a lot of warnings:

```
Incorrect link: module:utils/observablemixin~ObservableMixin.decorate
        at /Users/pomek/Projects/ckeditor/ckeditor5/packages/ckeditor5-core/src/command.js (line 157)

Incorrect link: module:utils/observablemixin~ObservableMixin.decorate
        at /Users/pomek/Projects/ckeditor/ckeditor5/packages/ckeditor5-ui/src/view.js (line 503)

Incorrect link: module:utils/observablemixin~ObservableMixin.decorate
        at /Users/pomek/Projects/ckeditor/ckeditor5/packages/ckeditor5-engine/src/controller/datacontroller.js (line 387)
```

ma2ciek:
> Yes, it does. Unfortunately, after switching to dev versions of our packages, my console shows a lot of warnings:

Because of two incorrect links that we have in docs. It should be: `module:utils/observablemixin~ObservableMixin#decorate`. That solves this issue.

packages/jsdoc-plugins/lib/relation-fixer/addmissingdoclets.js
ma2ciek: This part of the code is full of side effect. But I don't want to change it within that PR.

PR #500, opened 2019-03-14
Release tool will not crash when publishing a package for the first time

### LINKED ISSUE #491: The new package has not been pushed to GH during the release
That's the state after runnign the release script:

image

Also: 

image

pomek:

Fix: Release tool will not crash when publishing a package for the first time. Closes #491.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
@@ -143,7 +143,7 @@ module.exports = function releaseSubRepositories( options ) {
 		.then( () => {
 			process.chdir( cwd );
 
-			logProcess( `Finished releasing ${ chalk.underline( releasedPackages.size ) } packages.` );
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-dev-env/lib/release-tools/tasks/releasesubrepositories.js
```diff
@@ -256,6 +256,26 @@ module.exports = function releaseSubRepositories( options ) {
...
+				const errorAsArray = err.message.split( '\n' ).slice( 0, 2 );
+
+				if (
+					errorAsArray[ 0 ].endsWith( 'npm ERR! code E404' ) &&
+					errorAsArray[ 1 ] === `npm ERR! 404 '${ packageName }' is not in the npm registry.`
+				) {
+					return null;
+				}
+
+				throw new Error( err );
```
ma2ciek: or `throw err` since the error is already created?

pomek:
```suggestion
				throw err;
```

PR #513, opened 2019-04-10
Fixes errors on Travis

pomek:

Fix: Fixed issues related to a new dependency checker on Travis. Closes ckeditor/ckeditor5#1662.

### DIFF
--- a/packages/ckeditor5-dev-tests/bin/test-travis.sh
+++ b/packages/ckeditor5-dev-tests/bin/test-travis.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+# Breaks the entire script if any command returned exit code other than 0.
 set -e
 
 # Current work directory.
--- a/packages/ckeditor5-dev-tests/bin/test-travis.sh
+++ b/packages/ckeditor5-dev-tests/bin/test-travis.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 # Current work directory.
 PACKAGE_ROOT=$(pwd)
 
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-dev-tests/bin/test-travis.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
```
jodator: Please add info why `set -e` - I had to google it so others might as well as we do not code in Linux shell that much :)

packages/ckeditor5-dev-tests/bin/test-travis.sh
```diff
@@ -11,10 +13,9 @@ PACKAGE_NAME=$(node -e "console.log( require( process.cwd() + '/package.json' ).
 # The `.ckeditor5_test_environment` file is created by the "install-dependencies" script.
 CKEDITOR5_TEST_ENVIRONMENT=$(cat ${PACKAGE_ROOT}/.ckeditor5_test_environment)
 
-# Those tasks must be executed from the original package.
+# Linter and dependency checker tasks must be executed from the original package.
 yarn run lint && \
-${ROOT_BIN}/ckeditor5-dev-tests-check-dependencies
-
+${ROOT_BIN}/ckeditor5-dev-tests-check-dependencies && \
```
jodator: Taking that we have `set -e` it the `&&` still necessary?

Looks safe but might be superfluous...

packages/ckeditor5-dev-tests/bin/test-travis.sh
pomek: Done.

packages/ckeditor5-dev-tests/bin/test-travis.sh
pomek: I added two solutions which fix one problem. Most probably we don't need this change but I am safer if we would keep it.

pomek:
https://travis-ci.org/ckeditor/ckeditor5-autoformat/builds/518178545 – this build fails because our dep-check returns some errors.
https://travis-ci.org/ckeditor/ckeditor5-autoformat/builds/518180711 – after fixing dep-check errors, build works. All tests finish their work correctly.

PR #522, opened 2019-05-14
Change the `space-before-function-paren` rule to preserve spaces before async arrow functions

### LINKED ISSUE #521: The async keyword should be followed by a single space
The correct code for the async arrow function with the current `eslint-config-ckeditor5` rules looks like this:

```js
it( 'should display header message', async() => {} );
```

While the proper code style IMO should look like this:

```js
it( 'should display header message', async () => {} );
```

Unfortunately, it's a BC for anybody using our preset and the `async/await` syntax.

cc @jodator @Reinmar 

Ref https://eslint.org/docs/rules/space-before-function-paren.

ma2ciek:

Other: Updated `ecmaVersion` from `es6` to `es2018` (`es9`). Changed the `space-before-function-paren` rule to preserve spaces before async arrow functions. Closes #521.

---

### Additional information

BREAKING CHANGE: During async arrow function linting, the space before the parenthesis will be required, unlike in the past.

The correct code style after this change:

```
const fn = function x() {};
const anonymousFn = function() {};

const arrowFn = () => {};

const asyncArrowFn = async () => {};
const asyncAnonymousFn = async function() {};
const asyncFn = async function y() {};
```

### DIFF
--- a/packages/eslint-config-ckeditor5/.eslintrc.js
+++ b/packages/eslint-config-ckeditor5/.eslintrc.js
@@ -8,7 +8,7 @@
 module.exports = {
 	extends: 'eslint:recommended',
 	parserOptions: {
-		ecmaVersion: 6,
+		ecmaVersion: 2018,
 		sourceType: 'module'
 	},
 	env: {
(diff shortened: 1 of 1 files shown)


ma2ciek:
Can we merge it?

PR #528, opened 2019-07-05
Upgraded dependencies and merged Lerna+Yarn

### LINKED ISSUE #466: Switch to lerna+yarn
We use lerna+npm, but yarn better supports lerna (see e.g. https://stackoverflow.com/questions/52181762/are-there-any-advantages-to-using-lerna-with-yarn-workspaces). The only problem I can see is with releasing packages – I'm not sure how well yarn handles this.

### LINKED ISSUE #527: Upgrade dependencies
It does not show all the issues:

image

but we should upgrade those deps.

pomek:

Other: Upgraded dependencies for most of the packages. Merged Lerna + Yarn and they can work together now. Closes #527. Closes #466.


.travis.yml
```diff
@@ -1,14 +1,17 @@
...
 cache:
   - node_modules
+before_install:
+  - npm i -g yarn lerna
 install:
-  - npm install
-  - npm install coveralls
-  - lerna bootstrap --no-ci
+  - lerna bootstrap
+  - yarn add coveralls --ignore-workspace-root-check
```
ma2ciek: Shouldn't it be added to root deps?

pomek: ? 

It is. And it is required ONLY for CI so we are installing in on CI.

ma2ciek: Let's add it this way. It changes the `yarn.lock` file, so the CI tests won't be predictable. That what I meant. Having another one small dep installed locally isn't such a big problem IMO.

pomek: What a problem are you trying to solve?

ma2ciek: Using our `yarn.lock` file. I don't see much sense tracking it by Git and not using it on CI.

pomek: By removing `--no-ci` flag from Lerna, now it will use the lock file. Still, not related to the issue.

ma2ciek: `yarn add` will reset the lock file.

ma2ciek: It's connected to the issue because you're adding the `yarn.lock` file.

pomek: Ehh. First is being executed `lerna bootstrap` which bootstrap everything and it uses the lock file. Then, when everything is done, we don't care about the deps because they are already installed and we can safely add `coveralls`. If we want to be more precise, we can move the entire `yarn add coveralls --ignore-workspace-root-check` to `after_success` hook. In the future, would be good to be more precise because after 5 hours I finally was able to understand what's the problem (which in fact wasn't a problem).

pomek: PS: [`yarn.lock`](https://github.com/ckeditor/ckeditor5-dev/blob/master/yarn.lock) was added in https://github.com/ckeditor/ckeditor5-dev/commit/3ba004febaf806ebfae47518c0940a8c22184f11.

ma2ciek: I meant that installing one package means that you install its dependencies as well, so you can't assume that the `node_modules looks like before as `yarn add` changes `package.json`, `yarn.lock` and `node_modules`. Every installation will affect other packages too. Also, we can't track this way the version of `coveralls` we use. I'm not saying it's a crucial or vulnerable package, but I still don't know why this package can't be simply added to `node_modules`, especially since we always want to use packages at their specific versions.

> PS: yarn.lock  was added in 3ba004f.

Sorry, I overlooked it, GH didn't show removed lines where I scrolled down

Screen Shot 2019-07-05 at 16 04 03

pomek: So, can we agree that moving the package to proper Travis hook instead of adding a package that never will be used on our local machines would be a better solution?

ma2ciek: Yes! 🎉

pomek: Done.

PR #530, opened 2019-07-08
t/529: Feature: Enabled `CKEditorWebpackPlugin` in `getWebpackConfigForManualTes...

### LINKED ISSUE #529: Allow running manual tests with different languages
A possibility to use `CKEditorWebpackPlugin` with different `language` and `additionalLanguages` setups is required by https://github.com/ckeditor/ckeditor5/issues/1151.

oleq:

Feature: Enabled `CKEditorWebpackPlugin` in `getWebpackConfigForManualTests()`. Allowed configuring `language` and `additionalLanguages`. Closes #529.

----

I refactored some code in this PR switching from separate arguments to `options` object because some helpers got too many arguments. Also added some missing docs.

Sample usage:

```
yarn manual --files=utils --language=en --additionalLanguages=ar,pl
```

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
+++ b/packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
@@ -19,19 +19,34 @@ const transformFileOptionToTestGlob = require( '../utils/transformfileoptiontote
  * Main function that runs manual tests.
  *
  * @param {Object} options
- * @param {Array.<String>} options.files
+ * @param {Array.<String>} options.files Glob patterns specifying which tests to run.
+ * @param {String} options.themePath A path to the theme the PostCSS theme-importer plugin is supposed to load.
+ * @param {String} [options.language] A language passed to `CKEditorWebpackPlugin`.
+ * @param {Array.<String>} [options.additionalLanguages] Additional languages passed to `CKEditorWebpackPlugin`.
  * @returns {Promise}
  */
 module.exports = function runManualTests( options ) {
 	const buildDir = path.join( process.cwd(), 'build', '.manual-tests' );
 	const files = ( options.files && options.files.length ) ? options.files : [ '*' ];
-	const manualTestFilesPattern = files.map( file => transformFileOptionToTestGlob( file, true ) );
+	const patterns = files.map( file => transformFileOptionToTestGlob( file, true ) );
+	const additionalLanguages = options.additionalLanguages ? options.additionalLanguages.split( ',' ) : null;
 
 	return Promise.resolve()
 		.then( () => removeDir( buildDir ) )
 		.then( () => Promise.all( [
-			compileManualTestScripts( buildDir, manualTestFilesPattern, options.themePath ),
-			compileManualTestHtmlFiles( buildDir, manualTestFilesPattern ),
+			compileManualTestScripts( {
+				buildDir,
+				patterns,
+				themePath: options.themePath || null,
+				language: options.language || 'en',
+				additionalLanguages
+			} ),
+			compileManualTestHtmlFiles( {
+				buildDir,
+				patterns,
+				language: options.language || 'en',
+				additionalLanguages
+			} ),
 			copyAssets( buildDir )
 		] ) )
 		.then( () => createManualTestServer( buildDir ) );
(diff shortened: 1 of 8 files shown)


packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -44,6 +48,11 @@ module.exports = function compileHtmlFiles( buildDir, manualTestScriptsPatterns
 	const staticFiles = _.flatten( sourceDirs.map( sourceDir => {
 		return globSync( path.join( sourceDir, '**', '*.!(js|html|md)' ) );
 	} ) ).filter( file => !file.match( /\.(js|html|md)$/ ) );
+	const languagesToLoad = [];
+
+	if ( options.additionalLanguages ) {
+		languagesToLoad.push( options.language, ...options.additionalLanguages );
```
ma2ciek: Does it work? There're two options:

1. Additional languages are defined, then:
   - The main language is added to the bundle
   - Other languages are added to separate files.
2. Additional languages aren't defined, then:
   - The main language is built-in to the bundle. No other files are generated.

I mean, these two lines aren't necessary IMO. I'd just set `const languagesToLoad = options.additionalLanguages || [];`

packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
```diff
@@ -19,19 +19,35 @@ const transformFileOptionToTestGlob = require( '../utils/transformfileoptiontote
...
+				buildDir,
+				patterns,
+				themePath: options.themePath || null,
+				language,
+				additionalLanguages
+			} ),
+			compileManualTestHtmlFiles( {
+				buildDir,
+				patterns,
+				language,
```
ma2ciek: This prop shouldn't be needed.

packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
```diff
@@ -19,19 +19,35 @@ const transformFileOptionToTestGlob = require( '../utils/transformfileoptiontote
  * Main function that runs manual tests.
  *
  * @param {Object} options
- * @param {Array.<String>} options.files
+ * @param {Array.<String>} options.files Glob patterns specifying which tests to run.
+ * @param {String} options.themePath A path to the theme the PostCSS theme-importer plugin is supposed to load.
+ * @param {String} [options.language] A language passed to `CKEditorWebpackPlugin`.
+ * @param {String} [options.additionalLanguages] Additional languages passed to `CKEditorWebpackPlugin`.
```
ma2ciek: I'd expect `Array.<String>` here. Maybe you can parse it in the `parseArgument` fn?

packages/ckeditor5-dev-tests/lib/utils/manual-tests/compilehtmlfiles.js
ma2ciek: There's also the `additionalLanguages: all` thing, that the `CKEditroWebpackPlugin` accepts, but IDK if we want to support it as the languages are determined during the compilation time based on the existing PO files.

oleq: It is necessary. If both `language` and `additionalLanguages` are specified, you must load the former and the later. Otherwise

```
yarn manual --files=core,utils --language=en --additionalLanguages=ar,pl
```

image

because `translations/en.js` is missing

image

oleq: Not sure why it is so. Can you investigate that?

ma2ciek: With the following config:

```js
new CKEditorWebpackPlugin( {
	// See https://ckeditor.com/docs/ckeditor5/latest/features/ui-language.html
	language: options.language,
	additionalLanguages: options.additionalLanguages,
	verbose: true
} )
```

I could remind myself the problem 😄

```
Warning: Because of the many found bundles, none of the bundles will contain the main language.
You should add it directly to the application from the 'translations/en.js'.
```

So... Actually, everything should be fine as it is now...

packages/ckeditor5-dev-tests/lib/utils/automated-tests/parsearguments.js
```diff
@@ -56,6 +56,10 @@ module.exports = function parseArguments( args ) {
 		options.files = options.files.split( ',' );
 	}
 
+	options.language = options.language || 'en';
```
ma2ciek: I linked to the incorrect file 😞

This is for automated tests, for manual tests we do parsing in the `test-manual.js` file.

oleq: Manual tests use the same util. See https://github.com/ckeditor/ckeditor5-dev/blob/t%2F529/packages/ckeditor5-dev-tests/bin/test-manual.js#L15 and https://github.com/ckeditor/ckeditor5-dev/blob/t%2F529/packages/ckeditor5-dev-tests/lib/index.js#L11.

PR #537, opened 2019-07-25
Add support for debugging flags in automated and manual tests

ma2ciek:

Feature: Added support for debugging flags in automated and manual tests available via the `--debug` (`-d`) flag. Closes ckeditor/ckeditor5#383.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5/pull/1912.

This PR adds support for the `--debug` (`-d`) flag in `yarn test` and `yarn manual` scripts. By default, it uncomments all lines starting with `// @if CK_DEBUG //`, but it also supports custom debugging flags. For example, the `yarn test --debug engine` will uncomment both `// @if CK_DEBUG //` and `// @if CK_DEBUG_ENGINE //`. The special `--debug false` turns off the default debugging and makes the `// @if CK_DEBUG //` statements uncommented.


packages/ckeditor5-dev-tests/lib/utils/ck-debug-loader.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * The loader matches sentences like: `// @if CK_DEBUG // someDebugCode();` and uncomment them.
+ * If also uncomments code after specific flags if they are provided to the webpack configuration.
```
Reinmar:
```suggestion
 * It also uncomments code after specific flags if they are provided to the webpack configuration.
```

Reinmar:
> Support for `--no-debug` should be nice. In such a case, `options.debug` is a boolean (`false`).

What would this option do?

pomek:
Same what `"--debug false"` does. For blocking options via CLI, people are using `--no-*` syntax.

pomek:
I'm also not sure about unlocking `--debug` options by default in automated tests. In the nearby future, we will have a solid mess in our build logs.

☝️ I'm assuming that we will have a lot of `console.log()` calls behind those debug flags.

Reinmar:
I'm sorry, I don't know what this change does precisely. My guess is that we nee `--debug false` because `--debug` is on by default in our tools. Am I correct?

pomek:
`--debug false` === `--no-debug`.

Reinmar:
How does it answer my question? My question was about why do we need `--debug false` or `--no-debug`.

ma2ciek:
@Reinmar, Yep, `debug` is turned on by default, like we've been discussing it in the https://github.com/ckeditor/ckeditor5/issues/383#issuecomment-505350323

@pomek I can add the support with:

pomek:
Yes, it will.

PR #546, opened 2019-08-09
Update acorn dependency.

pjasiun:

Internal: Update acorn dependency.

---

### Additional information


pjasiun:
In fact, we could remove `ecmaVersion` options since `9` is the default value.

ma2ciek:
I've checked it and it works perfectly. 🍺

PR #551, opened 2019-08-14
JSDoc validator found errors, mark the entire process with proper exit code

### LINKED ISSUE #550: Errors raised in JSDoc does not stop building docs process
I've noticed that errors in documentation which appears in JSDoc don't stop building process and it is finished as complete without error. I noticed it when I saw errors on the master branch:
Screenshot 2019-08-13 at 12 59 00
What's more, there is a flag which seems to intentionally disable those errors with its default state: `If not passed, the errors will be printed to the console but the task will finish with '0'.`
https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-docs/lib/index.js

It seems like the wrong approach as allows to merge branches with errors in the documentation.

pomek:

Fix: JSDoc validator will end with proper exit code if, during the validation, any error occurred. Closes #550.

### DIFF
--- a/packages/jsdoc-plugins/lib/validator/validator.js
+++ b/packages/jsdoc-plugins/lib/validator/validator.js
@@ -21,6 +21,11 @@ exports.handlers = {
 		// Prints only first 50 errors to stdout.
 		printErrors( errors, 50 );
 
+		// Mark the process as ended with error.
+		if ( errors.length ) {
+			process.exitCode = 1;
+		}
+
 		if ( process.env.JSDOC_VALIDATE_ONLY ) {
 			process.exit();
 		}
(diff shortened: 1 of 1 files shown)


pomek:
Should be merged together with https://github.com/ckeditor/ckeditor5/pull/1968.

ma2ciek:
Taking the occasion, could we move https://github.com/ckeditor/ckeditor5-dev/pull/551/files#diff-d9ba2abf4483c7964be2611fb7a2fb73R58-R60 outside of the `printErrors()` function.

pomek:
@ma2ciek, could you check once again?

ma2ciek:
LGTM

PR #552, opened 2019-08-23
Adjusted the changelog generator to match to planned changes related to merging issue trackers

pomek:

Other: Adjusted the changelog generator to match to planned changes related to merging issue trackers. Closes ckeditor/ckeditor5#1988.

BREAKING CHANGE: Due to merging our issue trackers, `pkgJson.bugs` will point to the same place for every package. We cannot rely on this value anymore. See ckeditor/ckeditor5#1988. Instead of we can take a value from `pkgJson.repository` and adjust it to match to our requirements.

---

### Additional information

This PR won't change anything in our project because we already have defined `pkgJson.repository` for each package but if somebody uses the tool and he hasn't defined the value, the tool will throw an ugly error and that's why I marked it as BC.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
@@ -79,7 +79,7 @@ const transformCommitUtils = {
 			const packageJson = getPackageJson();
 
 			// Due to merging our issue trackers, `packageJson.bugs` will point to the same place for every package.
-			// We cannot relay on this value anymore. See: https://github.com/ckeditor/ckeditor5/issues/1988.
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-dev-env/lib/release-tools/utils/transform-commit/transform-commit-utils.js
```diff
@@ -77,13 +77,26 @@ const transformCommitUtils = {
 			}
 
 			const packageJson = getPackageJson();
-			const issuesUrl = ( typeof packageJson.bugs === 'object' ) ? packageJson.bugs.url : packageJson.bugs;
+
+			// Due to merging our issue trackers, `packageJson.bugs` will point to the same place for every package.
+			// We cannot relay on this value anymore. See: https://github.com/ckeditor/ckeditor5/issues/1988.
```
pomek:
```suggestion
			// We cannot rely on this value anymore. See: https://github.com/ckeditor/ckeditor5/issues/1988.
```

PR #559, opened 2019-09-19
Introduce the `--disallow-console-use` flag in automated tests

ma2ciek:

Feature: Introduced the `--disallow-console-use` flag in automated tests. Disallowed console use in CI tests. Part of ckeditor/ckeditor5#1996.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-tests/README.md
+++ b/packages/ckeditor5-dev-tests/README.md
@@ -53,7 +53,7 @@ You can also use the bin script for testing a package:
 * `files` - Specify file(s) to test. Also available as an alias: `-f`.
 * `browsers` - Browsers which will be used to run the tests. Also available as an alias: `-b`.
 * `reporter` - Mocha reporter – either `mocha` (default) or `dots` (less verbose one).
-* `disallow-console-use` - Makes `console.log()/warn()/error()` throw an error when set to `true`
+* `disallow-console-use` - Whether to throw an error when the `console.log()/warn()/error()/info()/debug()` is executed.
 
 #### Examples
 
--- a/packages/ckeditor5-dev-tests/README.md
+++ b/packages/ckeditor5-dev-tests/README.md
@@ -53,6 +53,7 @@ You can also use the bin script for testing a package:
 * `files` - Specify file(s) to test. Also available as an alias: `-f`.
 * `browsers` - Browsers which will be used to run the tests. Also available as an alias: `-b`.
 * `reporter` - Mocha reporter – either `mocha` (default) or `dots` (less verbose one).
+* `disallow-console-use` - Makes `console.log()/warn()/error()` throw an error when set to `true`
 
 #### Examples
 
--- a/packages/ckeditor5-dev-tests/README.md
+++ b/packages/ckeditor5-dev-tests/README.md
@@ -53,7 +53,7 @@ You can also use the bin script for testing a package:
 * `files` - Specify file(s) to test. Also available as an alias: `-f`.
 * `browsers` - Browsers which will be used to run the tests. Also available as an alias: `-b`.
 * `reporter` - Mocha reporter – either `mocha` (default) or `dots` (less verbose one).
-* `disallow-console-use` - Whether to throw an error when the `console.log()/warn()/error()/info()/debug()` is executed.
+* `disallow-console-use` - Whether to throw an error when one of the console methods (e.g. `console.log()`) is executed.
 
 #### Examples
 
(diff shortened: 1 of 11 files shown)


packages/ckeditor5-dev-tests/README.md
```diff
@@ -53,6 +53,7 @@ You can also use the bin script for testing a package:
 * `files` - Specify file(s) to test. Also available as an alias: `-f`.
 * `browsers` - Browsers which will be used to run the tests. Also available as an alias: `-b`.
 * `reporter` - Mocha reporter – either `mocha` (default) or `dots` (less verbose one).
+* `disallow-console-use` - Makes `console.log()/warn()/error()` throw an error when set to `true`
```
mlewand: We're describing CLI flag/option here, so instead of treating it as a bool-ish thing (though in the end it is ofc converted to bool). Actually it would be nice to have this consistent with previous flag, which goes like: "- Whether to (...)".

packages/ckeditor5-dev-tests/bin/test-travis.sh
```diff
@@ -18,5 +18,5 @@ CKEDITOR5_TEST_ENVIRONMENT=$(cat ${PACKAGE_ROOT}/.ckeditor5_test_environment)
 yarn run lint && \
 ${ROOT_BIN}/ckeditor5-dev-tests-check-dependencies && \
 cd ${CKEDITOR5_TEST_ENVIRONMENT} && \
-node --max_old_space_size=4096 $ROOT_BIN/ckeditor5-dev-tests --files=$PACKAGE_NAME --coverage --reporter=dots --browsers=Chrome && \
-node --max_old_space_size=4096 $ROOT_BIN/ckeditor5-dev-tests --files=$PACKAGE_NAME --reporter=dots --browsers=Firefox
+node --max_old_space_size=4096 $ROOT_BIN/ckeditor5-dev-tests --files=$PACKAGE_NAME --coverage --reporter=dots --browsers=Chrome --disallow-console-use && \
+node --max_old_space_size=4096 $ROOT_BIN/ckeditor5-dev-tests --files=$PACKAGE_NAME --reporter=dots --disallow-console-use --browsers=Firefox
```
mlewand:
```suggestion
node --max_old_space_size=4096 $ROOT_BIN/ckeditor5-dev-tests --files=$PACKAGE_NAME --reporter=dots --browsers=Firefox --disallow-console-use
```
Nit: it would be nice to have both in the same order.

packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
```diff
@@ -91,11 +96,31 @@ function createEntryFile( globPatterns ) {
...
+beforeEach( () => {
+	console.log = ( ...data ) => {
+		originalWarn( 'Detected \`console.log()\`:', ...data );
+		throw new Error( 'Detected \`console.log()\`' );
+	};
+	console.warn = ( ...data ) => {
+		originalWarn( 'Detected \`console.warn()\`:', ...data );
+		throw new Error( 'Detected \`console.warn()\`' );
+	};
+	console.error = ( ...data ) => {
```
mlewand: Since we are covering log, warn and error it make sense to also cover `console.info` which is something that can be used similarly. I checked the reference and there's also `console.debug` 🤔  I think we can cover all five, in which case you'd probably want to do it in loop to simplify it.

packages/ckeditor5-dev-tests/tests/tasks/runautomatedtests.js
```diff
@@ -122,14 +124,14 @@ describe( 'runAutomatedTests', () => {
 
 				expect( stubs.fs.writeFileSync.calledOnce ).to.equal( true );
 				expect( stubs.fs.writeFileSync.firstCall.args[ 0 ] ).to.equal( '/workspace/build/.automated-tests/entry-point.js' );
-				expect( stubs.fs.writeFileSync.firstCall.args[ 1 ] ).to.equal( expectedEntryPointContent );
+				expect( stubs.fs.writeFileSync.firstCall.args[ 1 ] ).to.include( expectedEntryPointContent );
```
mlewand: Let's keep these tests strict, making sure that `expectedEntryPointContent` contains also the code generated for handling `console.*` members, will prevent us from accidentally dropping this feature.

packages/ckeditor5-dev-tests/tests/tasks/runautomatedtests.js
```diff
@@ -176,4 +179,41 @@ describe( 'runAutomatedTests', () => {
 				}
 			);
 	} );
+
+	it( 'should warn when the `disallowConsoleUse` option is set to `false`', () => {
```
mlewand: We should also add another unit case, which ensures that `disallowConsoleUse=false` does not add our custom code, so it will be doing a similar assertion to the "should create an entry file before tests execution" case.

ma2ciek:
@mlewand it's ready for another round.

PR #567, opened 2019-09-26
Fix console patching

### LINKED ISSUE #566: Fix overriding console methods
Right now the console patching mistakenly overrides FF console properties. Thus a property/function filtering is necessary to fix this situation.

ma2ciek:

Internal: Fixed console patching. Closes #566.

---

### Additional information


mlewand:
> One thing that came to my mind – should we consider adding a test for the code snippet? I believe it shouldn't break in the future but who knows. OTOH, I have no idea how to test it now.

You mean the fragment generated by this console use prevention feature? It's already tested, tests were added in previous iteration so we're safe from removing it unintentionally.

PR #573, opened 2019-10-25
Improvements for changelog generators

### LINKED ISSUE #571: Improvements for changelog generator
1. Change the color of the title (that describes a process). Some terminals can display it too dark:
image
2. For all commits printed out during "Looking for "MAJOR BREAKING CHANGES" commits...", add a link to that commit.
3. Add indentations in several places. E.g.:
image
    - Generating changelog for "..." could be a list
    - All commits from that package are another list
4. After confirming that a release will be a major release, the tool could predict a new version. Find the highest one in all packages and bump its major. After printing out the data, a user must confirm the new version (it can be also modified). This version is suggested in all packages by default.
5. In all cases, a list of commit for parsing packages should be displayed. Now it is disabled if the next version is treated as the major bump.
6. When a guide that explains our versioning will be published, would be good to add a link to the guide in the changelog (as an icon next to the header, e.g.: `### MAJOR BREAKING CHANGES [icon](url)`
7. Summary changelog generator – move a part generated from commits before changes in dependencies.
8. Single package or ckeditor5-dev repository changelog generators: use `BREAKING CHANGES` instead of `MAJOR|MINOR BREAKING CHANGES`.
9. BUG: if the user typed "internal" as the new version, it does not bump the patch. It uses "internal" as the version.

pomek:

Feature: Introduced several improvements for `generateSummaryChangelog()` and `generateChangelogForSubRepositories()` tasks. Closes #571.

- All logs are indented. It increased readability.
- Changed colors in the titles sections because on some terminals they could be too dark.
- `MAJOR|MINOR BREAKING CHANGES` notes are displayed on a list where all commits are printed out.
- `generateChangelogForSubRepositories()` – `MAJOR BREAKING CHANGES` commits contain URLs to GitHub. It allows opening specified commit in the web-browser faster.
- `generateChangelogForSubRepositories()` - after confirming a major release, you can type a version that will be proposed for all packages. The mechanism looks for the highest one and bumps its major.
- `generateSummaryChangelog()` - introduced a new group of packages when the release was a major bump - "Major releases (contain minor breaking changes)".

---

### Additional information

image

image

image


Reinmar:
I've run the tool and found that the readability is still not perfect. 

1. It seems that it'll be better to have this question in one line:

   image

2. Unfortunately, completely outdented questions break the readability. It completely distorts the left edge od the text. 

   image

   So, we need to figure out how to indent the question.

3. There's a missing newline after the "Looking for..." heading. Please make sure they are consistently surrounded by newlines (all other ones I saw were surrounded):

   image

The three most important aspects that improve readability are:

* consistency
* use of white space
* consistency

pomek:
Fixed with the latest commits:

image

PR #583, opened 2019-12-16
Integration with Intellij IDEs

### LINKED ISSUE #582: Tests: Integrate with IntelliJ IDEs
It should be possible to run tests directly from any IntelliJ IDE, like WebStorm. This would enable access to a much more powerful visualization of the test results, highly enhancing productivity.

fredck:
Feature: Added integration with Intellij IDEs. Closes #582.

### DIFF
--- a/packages/ckeditor5-dev-tests/README.md
+++ b/packages/ckeditor5-dev-tests/README.md
@@ -99,6 +99,12 @@ $ npm t -- --files='!(engine|ui)'
 | `*` | `node_modules/ckeditor5-*/tests/**/*.js` | all installed package's tests |
 | `ckeditor5` | `tests/**/*.js` | tests from the main repository |
 
+## IDE integrations
+
+The CKEditor 5 can be integrated with IDEs via integrations.
+
+Currently only the IntelliJ based IDEs are supported (WebStorm, PHPStorm, etc). Detailed information are provided in [`bin/intellijkarmarunner/README.md`](./bin/intellijkarmarunner/README.md).
+
 ## Changelog
 
 See the [`CHANGELOG.md`](https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-tests/CHANGELOG.md) file.
--- a/packages/ckeditor5-dev-tests/bin/intellijkarmarunner/README.md
+++ b/packages/ckeditor5-dev-tests/bin/intellijkarmarunner/README.md
@@ -1,13 +1,59 @@
-## Test runner adapter for IntelliJ
+# Test runner adapter for IntelliJ
 
-To run CKEditor 5 tests integrated into an IntelliJ IDE of your preference, the IDE test runner must be configured to use the adapter provided within the package.
+This runner allows to benefit from integrated testing environment inside your IDE. With proper setup you can
+- run pre-defined set of tests (group of packages)
+- run any tests right from the editor
+- run tests in debug mode inside IDE
+- run tests with coverage
 
-1.  In the IDE, go to _Run_ > _Edit configurations..._:
-    1.  Add a new configuration of type "**Karma**".
-    2.  Give a name to your configuration, at your preference.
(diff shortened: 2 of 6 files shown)


fredck:
This implementation is the very minimum necessary to run our test runner as a karma runner in the IDE.

A possible future evolution for this feature could be the introduction of a dedicated plugin to IntelliJ, which would add the "CKEditor 5" test runner in the run configurations. This must be done in Java though.

jodator:
Selection_099_good

It is a minimal but working great plugin. After a few test runs I can see it's potential. Current drawbacks are:
1. Cannot run coverage.
2. Cannot run a single test (ie re-run failed tests).
3. Probably you can't run it from IDE from a test file.
4. You have to have different configurations for different setups if you wish to run specific tests. You must create a test run configuration for every `--files=""` variation that you use.

... but ...

it works and is my Christmas gift from @fredck :D :gift_heart: :christmas_tree: :santa:.

I'll continue on testing this and provide a possible next-steps for further enhancing this feature (coverage, single test re-run?).

jodator:
To make coverage work you need to pass `-c` option alongside the `--files` directive. Then the karma plugin pickups the created coverage:

Selection_102

Selection_103

jodator:
More "feels good man": The debug works out of the box:

Selection_104

jodator:
Aaaand you can run single test from the test run window, it will however create temporary configs:
Selection_105

jodator:
Aaaand to make running tests from the test file you need to modify your project's template for karma like below:
Selection_106

The drawback is that the first run is slow as it load **all** tests (like running withouth `--files` directive_ but subsequent runs are fast.

The other is that when you run the test from a test file with a coverage it does not collect it but that's a minor issue as you can always create a run configuration with `--files=` and it will cover the whole package.

jodator:
Running tests from a file through IDE fancy buttons:

Peek 2019-12-17 19-56

jodator:
@pomek do you know if the problem of not collecting a coverage might be related to not providing a `--files` directive? It looks like the `-c` switch is ignored even if you run the tests from a console.

jodator:
This PR looks better then @fredck described it to me before. It turned out that the things that your said do not work actually work pretty well with only some configuration hacks. Probably knowing the IDE and our `yarn run` switches helped a lot. But this integration has it all:
- single test run
- coverage
- debug tests

It is a tool that we will benefit if used (debug from the IDE is sweet).

I hope that we could make coverage working also as I'm pretty sure that it is disabled for some reason.

pomek:
> @pomek do you know if the problem of not collecting a coverage might be related to not providing a `--files` directive? It looks like the `-c` switch is ignored even if you run the tests from a console.

[`options.coverage` bases on value from `options.files`](https://github.com/ckeditor/ckeditor5-dev/blob/d86595eddd4b50f9fb1863a103916f7aec1b0374/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getwebpackconfig.js#L74-L88). The solution we have right is not perfect and I would say, it's hard to understand what's going on. We could improve that and maybe the entire plugin will understand what you want to achieve.

fredck:
> This PR looks better then @fredck described

That's a good surprise. Thanks for the in-depth details, @jodator.

Btw, it may be true that we can always setup the IDE configurations with `-c`... by doing so, I've noticed the following message being output when the IDE runs when using "Run xxx" without using the "Run xxx with coverage" option:

```plaintext
IntelliJ integration disabled coverage for faster run and debug capabilities
```

fredck:
Aaaaaand... I've notice that you can "Rerun Failed Tests" without having to created a new configuration, which is great.

fredck:
Aaaaaaaaaaand.... coverage works for me with `-s`.

jodator:
> I've noticed the following message being output when the IDE runs when using "Run xxx"

Hmm I didn't noticed that but that's a sensible option to set and a warning is kinda what we end up with.

jodator:
> Aaaaaaaaaaand.... coverage works for me without `-s`.

Ah that may be right as it is done by the isntrumentation. Anayway the `-s` is useful for navigation :). 

We can always work on that readme more :)

PR #587, opened 2020-01-07
Automatic DOM leak detection for unit tests

### LINKED ISSUE #586: Automatic DOM leak detection for unit tests
In the past we had multiple cases when tests would give different results when run alone and with all other test suites (e.g. ckeditor/ckeditor5#6002).

With ckeditor/ckeditor5#6002 we were able to fix all the cases that were leaking. Now having a clear situation it's a perfect time to introduce automated detection (given that it's not difficult to make one) that will fail tests if there's a new DOM leak introduced.

mlewand:

Feature: Introduced automatic DOM leak detection for unit tests. Closes #586.

MAJOR BREAKING CHANGE: `--disallow-console-use` flag was removed, use `--production` instead.

---

### Additional information

See https://github.com/ckeditor/ckeditor5-dev/issues/586#issuecomment-571573488 for some notes.

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/utils/automated-tests/leaksdetector.js
+++ b/packages/ckeditor5-dev-tests/lib/utils/automated-tests/leaksdetector.js
@@ -1,0 +1,29 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* global window, document */
+
+( function() {
+	let lastDomElementsCount;
+
+	beforeEach( () => {
+		lastDomElementsCount = document.body.childElementCount;
+	} );
+
+	afterEach( () => {
+		if ( document.body.childElementCount !== lastDomElementsCount ) {
+			const leaksCount = document.body.childElementCount - lastDomElementsCount;
+			const errorMessage = `${ leaksCount } elements were leaked. Be a good citizen and clean your DOM once you're done with it.`;
+
+			lastDomElementsCount = document.body.childElementCount;
+
+			if ( window.production ) {
+				throw new Error( errorMessage );
+			} else {
+				console.error( errorMessage );
+			}
+		}
+	} );
+}() );
--- a/packages/ckeditor5-dev-tests/lib/utils/automated-tests/leaksdetector.js
+++ b/packages/ckeditor5-dev-tests/lib/utils/automated-tests/leaksdetector.js
@@ -19,6 +19,7 @@
 
 			lastDomElementsCount = document.body.childElementCount;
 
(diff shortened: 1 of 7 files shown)


packages/ckeditor5-dev-tests/lib/utils/automated-tests/leaksdetector.js
```diff
@@ -0,0 +1,29 @@
...
+	let lastDomElementsCount;
+
+	beforeEach( () => {
+		lastDomElementsCount = document.body.childElementCount;
+	} );
+
+	afterEach( () => {
+		if ( document.body.childElementCount !== lastDomElementsCount ) {
+			const leaksCount = document.body.childElementCount - lastDomElementsCount;
+			const errorMessage = `${ leaksCount } elements were leaked. Be a good citizen and clean your DOM once you're done with it.`;
```
pomek: 😄

pomek:
After introducing a small change, you can display a test case where a leak was detected:

```js
afterEach( function() {
	if ( document.body.childElementCount !== lastDomElementsCount ) {
		const leaksCount = document.body.childElementCount - lastDomElementsCount;
		const errorMessage = `${ leaksCount } elements were leaked. Be a good citizen and clean your DOM once you're done with it. Detected after test case: ${ this.currentTest.title }`;

		lastDomElementsCount = document.body.childElementCount;

		// See https://github.com/ckeditor/ckeditor5-dev/issues/586#issuecomment-571573488.
		if ( window.production ) {
			throw new Error( errorMessage );
		} else {
			console.error( errorMessage );
		}
	}
} );
```

Since a log (`console.error()`) is printed under a test case, there is no sense to duplicate such information:

image

But after adding one more hook, you can display after executing all tests which caused DOM-leaks:

```js
( function() {
	let lastDomElementsCount;
	const testsToCheck = [];

	beforeEach( () => {
		lastDomElementsCount = document.body.childElementCount;
	} );

	afterEach( function() {
		if ( document.body.childElementCount !== lastDomElementsCount ) {
			testsToCheck.push( this.currentTest.fullTitle() );

			const leaksCount = document.body.childElementCount - lastDomElementsCount;
			const errorMessage = `${ leaksCount } elements were leaked. Be a good citizen and clean your DOM once you're done with it.`;

			lastDomElementsCount = document.body.childElementCount;

			// See https://github.com/ckeditor/ckeditor5-dev/issues/586#issuecomment-571573488.
			if ( window.production ) {
				throw new Error( errorMessage );
			} else {
				console.error( errorMessage );
			}
		}
	} );

	after( () => {
		if ( !testsToCheck.length ) {
			return;
		}

		const message = 'Test cases listed below may leaking DOM elements: \n' +
			testsToCheck.map( testName => `  - ${ testName }` ).join( '\n' );

		console.warn( message );
	} );
}() );
```

image

But for now, we will go with the proposed solution without displaying additional notes.

PR #594, opened 2020-02-07
Made ckeditor5-dev-tests a general use application

### LINKED ISSUE #593: Tests: make it available for generic use
The ckeditor5-dev-tests application is very specific to the ckeditor/ckeditor5 repo only. It is very powerful though and it could be available for general use.

The main problem seems to be related to the `--file` argument, which transforms whatever is given to it to CKEditor specific stuff.

fredck:

Feature: Made ckeditor5-dev-tests a general use application. Closes ckeditor/ckeditor5#7196.

---

### Additional information

This implementation is pretty simple, doing mainly the following:

*   Allow `--files` to point to resolved globs, by beginning the with `./`.
*   Introduced `--coverage-paths` to point to additional paths to be included in coverage.

It basically doesn't change the current features, just adding to top of them.

I've been able to successfully use this with arguments like `--files=./tests/unit/**/*.js --coverage-paths=./src`.

This should even allow to enable tests when cloning a single editor plugin, instead of having to deal with the big and fancy `ckeditor5` repo.

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
+++ b/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
@@ -77,7 +77,9 @@ function createEntryFile( globPatterns, production ) {
 		let hasFiles = false;
 
 		for ( const resolvedPattern of globPatterns[ singlePattern ] ) {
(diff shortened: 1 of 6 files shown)


fredck:
Outdated

PR #595, opened 2020-02-13
Added support for the `"` character in translation messages and contexts

### LINKED ISSUE #523: Translation upload fail with no error if context contains `"` character
My context:

```js
"%0 of %1": "Label for a \"page X of Y\" status of a typical next/previous page navigation."
```

During upload it hangs on this:

image

I had to figure out which resource causes a problem (the error is logged in a random place) and find the pot file that causes it:

```
msgctxt "Label for a "page X of Y" status of a typical next/previous page navigation."
msgid "%0 of %1"
msgstr "%0 of %1"
```

ma2ciek:

Fix: Added support for the `"` character in translation messages and contexts. Closes #523.

---

### Additional information

I need to ensure that the `"` characters won't break during building the editor with the CKEditor5 webpack plugin.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/translations/collect-utils.js
+++ b/packages/ckeditor5-dev-env/lib/translations/collect-utils.js
@@ -186,9 +186,9 @@ const utils = {
 		return translationObjects.map( translationObject => {
 			// Note that order is important.
 			return [
-				`msgctxt "${ translationObject.ctxt }"`,
-				`msgid "${ translationObject.id }"`,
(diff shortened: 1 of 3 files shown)


ma2ciek:
It seems to work, Though there might be a problem with the `'` character as well, but during the webpack building process:

`e('The value is invalid. Try "10px" or "2em" or simply "2".')` 

I'll add some tests to check the correctness of `"` character escaping in the vulnerable places.

PR #613, opened 2020-04-09
Release tools - monorepo

### LINKED ISSUE #606: Adjust releasing tools to work with monorepo
Subissue of ckeditor/ckeditor5#6466.

Requires: https://github.com/ckeditor/ckeditor5-dev/issues/608.

pomek:

Other: Adjusted release tools to handle single mono-repository architecture. Closes #606.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/utils/getpackagestorelease.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/utils/getpackagestorelease.js
@@ -7,15 +7,18 @@
 
 const executeOnPackages = require( './executeonpackages' );
 const getPackageJson = require( './getpackagejson' );
-const versionUtils = require( './versions' );
 
 /**
  * Returns a list of packages which should be releases based on changes in a changelog file and tags created in Git repository.
  *
  * @param {Set} pathsToPackages A collection of paths to packages that should be checked.
+ * @param {Object} options
+ * @param {String} options.changes A description of changes for the main repository. It should contain a script-friendly list that provides
+ * new versions for parsed packages. If didn't find, `options.version` will be used instead.
+ * @param {String} options.version New version for the main repository.
  * @returns {Promise.<Map.<String, ReleaseDetails>>}
  */
-module.exports = function getPackagesToRelease( pathsToPackages ) {
+module.exports = function getPackagesToRelease( pathsToPackages, options ) {
 	const cwd = process.cwd();
 	const packagesToRelease = new Map();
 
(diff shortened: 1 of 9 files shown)


mlewand:
Also small UX thing: during the first use of `npm run release:bump-version -- --dry-run` I started to worry when the prompt was stuck for more than ~20s here:



Compared to previous packages build seemed to be stuck. After inspecting the changes I know now that it's because for builds we are are generating build package.

I was wondering whether we can quickly/cheaply append something like " (builds will take a longer time… relax)"?

pomek:
> there's one issue that I found in release script though

https://github.com/ckeditor/ckeditor5/blob/master/scripts/release/update-utils-version.js#L12-L13

If we skip some steps of the release process, things must be done manually.

packages/ckeditor5-dev-env/lib/release-tools/utils/getpackagestorelease.js
```diff
@@ -29,24 +32,43 @@ module.exports = function getPackagesToRelease( pathsToPackages ) {
...
 		}
 
 		return Promise.resolve();
 	}
 };
 
+// Search for a new version for specified package in the main changelog description.
+// It must be defined as following:
+//
+//      [packageName](https://www.npmjs.com/package/packageName): v0.0.1 => [v1.0.0](https://github.com/.../releases/tag/v1.0.0)
```
pomek, line 56: TODO: Check whether it works with the following entries:

```md
* [@ckeditor/ckeditor5-word-count](https://www.npmjs.com/package/@ckeditor/ckeditor5-word-count): v0.0.2
* [@ckeditor/ckeditor5-build-balloon](https://www.npmjs.com/package/@ckeditor/ckeditor5-build-balloon): v19.0.0 => v19.0.1
```

pomek: It won't Require changes. For upgrading existing repo:

```js
\[@ckeditor\/ckeditor5-word-count\].*: v[\d.]+ \=\> v([\d.]+)
```

For new release:

```js
\[@ckeditor\/ckeditor5-word-count\].*: v([\d.]+)
```

The order of execution is important. First must be checked an option with already existing package.

pomek:
> I was wondering whether we can quickly/cheaply append something like " (builds will take a longer time… relax)"?

The tool does not know which scripts are being executed. We could add something like "echo Building..." to `preversion` hook in our builds.

packages/ckeditor5-dev-env/lib/release-tools/utils/getpackagestorelease.js
pomek: Done.

mlewand:
> The tool does not know which scripts are being executed. We could add something like "echo Building..." to `preversion` hook in our builds.

Ok, so no cheap solution here. That being said, we can simply document this in our release instructions and this will be fine.

PR #640, opened 2020-05-19
Rebuild the changelog generator and the repository itself

pomek:
Feature (env): Support for multi-entries messages in the single commit and scoped changes. Closes ckeditor/ckeditor5#7207. Closes ckeditor/ckeditor5#7171. See https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html.

Feature (env): Added new utils that help to collect commits, parsing them, and generating the changelog.

* The util for generating changelog from commits (those must be specified as an argument). See `/lib/release-tools/utils/generatechangelog.js`
* The util for collecting commits. See `/lib/release-tools/utils/getcommits.js`
* The util for suggesting new version based on commits. See `/lib/release-tools/utils/getnewversiontype.js`

Feature (env): Task `generateChangelogForSinglePackage()` supports new options: `from` - a commit or tag for collecting commits since the last release, `highlightsPlaceholder` - whether to add "Release highlights" placeholder in the changelog, `collaborationFeatures` - whether to add a URL to collaboration features changelog.

Fix (env): The `getChangedFilesForCommit()` util filters files returned by the Git command. It won't return an empty string anymore.

Other (env): Commits in the changelog will display the word `commit` instead of the first 7 characters from the commit's hash. In big repositories (the number of commits is huge), 7 characters are not unique anymore.

Other (env): `Closes` references will be merged into a single entry. Github does not support such references (`Closes x, y`) but it can be simplified during the commit's transformation.

Other (env): The `provideVersion()` util from `lib/release-tools/utils/cli.js` allows disabling returning `skip` version by setting its option `disableSkipVersion` to `true`.

Other: Removed `lerna` and all its files from the project. Now the release process is handled by our tools. The entire repository will follow the same rules as `ckeditor5.` Read more: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/versioning-policy.html.

MAJOR BREAKING CHANGES (env): Removed `generateChangelogForSubPackages()` task. Use `generateChangelogForMonoRepository()` instead.

MAJOR BREAKING CHANGES (env): Removed `generateChangelogForSubRepositories()` task. Use `generateChangelogForMonoRepository()` instead if your repository is a monorepository.

MINOR BREAKING CHANGES (env): The util `getPackagesPaths()` does not check whether packages are defined as `dependencies` in `package.json` in the main repository.

MINOR BREAKING CHANGES (env): Task `generateChangelogForSinglePackage()` does not accept options: `newVersion`, `disableMajorBump`, `isInternalRelease`, `indentLevel`, `useExplicitBreakingChangeGroups` anymore. The task should be used for generating the changelog for the single repository.

MAJOR BREAKING CHANGES (env): Removed `generateSummaryChangelog()` task.

MINOR BREAKING CHANGES (env): Removed `hasMajorBreakingChanges()` and `hasMinorBreakingChanges()` utils from `/lib/release-tools/utils/changelog.js` helper.

MINOR BREAKING CHANGES (env): Removed `getSubPackagesPaths()` util.

MINOR BREAKING CHANGES (env): Removed the `getNewReleaseType()` util. Use `getCommits()` and `getNewVersionType()` instead.

MINOR BREAKING CHANGES (env): Removed support for the `NOTE` type of commit's notes.

MINOR BREAKING CHANGES (env): Moved all utils from `/lib/release-tools/utils/transform-commit` to `/lib/release-tools/utils`.

MINOR BREAKING CHANGES (env): Renamed `getSubRepositoriesPaths()` util to `getPackagesPaths()`.


pomek:
The [CHANGELOG.txt](https://github.com/ckeditor/ckeditor5-dev/files/4650288/CHANGELOG.txt) generated from the commit message (and commits since the last release of `ckeditor5-dev`).

pomek:
I replaced examples of multi-entries commits with a URL to our documentation. The attached changelog above shows why it can't be here (core and engine commits).

pomek:
>I got such message when running changelog command on ckeditor5 repo.
>
>I assume the intention here is to allow aborting major bump in case there is any commit marked as major breaking change? In that case I think it should be skipped if there are no commits like that.

We could skip that but somebody could make a commit with "MINOR BREAKING CHANGE" instead of "MAJOR...". The tool does not know about it but allows to bump the major version anyway.

It does not generate too much mess. If it's really a problem, let's remove it. If not, I suggest leaving it as it is.

---

> What is the difference between internal and skip?

The `internal` version generates the changelog like below:

> Internal changes only (updated dependencies, documentation, etc.).

While `skip` – does not generate anything.

---

> None of the changes affect horizontal rule, yet it has been proposed to be bumped by a major version.

Minor*.

Why? The merge commit (https://github.com/ckeditor/ckeditor5/commit/dbf24a29ac64f52bceb2efc106b50c736c16f1c3) touches any file located under the `/packages/ckeditor5-horizontal-rule` directory. If any file was touched, the package assumes the change described itself. There are a few ways to improve it:

- group changes by the scope (but you can easily miss something important)
- group changes by the scope and the package's dependencies (more information but  also more complication in the code)
- merge multi-commits into single commit on the list and show the single entry
- group changes by commit hash and separate them on the list

While grouping changes by the scope is the best solution, it requires strict commit messages: a number of changed packages must be equal to commit messages, we decided to go with a little simpler solution.

> Actually for some reason I don't see Internal (horizontal-line) entry from

It happens because the tool assumes that only public commits (Fix, Feature, Other) will be isolated. I'll fix that.

mlewand:
> We could skip that but somebody could make a commit with "MINOR BREAKING CHANGE" instead of "MAJOR...". The tool does not know about it but allows to bump the major version anyway.
> 
> It does not generate too much mess. If it's really a problem, let's remove it. If not, I suggest leaving it as it is.

The thing is that you don't even get these minor BC listed beforehand so it's not relevant.

Either minor BCs should be listed (which make sense too) or this step should be skipped in case there's 0 major BCs.

---

> Why? The merge commit ([ckeditor/ckeditor5@dbf24a2](https://github.com/ckeditor/ckeditor5/commit/dbf24a29ac64f52bceb2efc106b50c736c16f1c3)) touches any file located under the `/packages/ckeditor5-horizontal-rule` directory. If any file was touched, the package assumes the change described itself.

That wasn't perceivable when using the tool. "group changes by commit hash and separate them on the list" sounds on paper like a good balance between usability and time/effort required :+1:

pomek:
> When ran on ckeditor5 one of scoped widget changelog entries leaked some internal entries. I believe that this should be fixed with this:
> 
> > Actually for some reason I don't see Internal (horizontal-line) entry from
> 
> It happens because the tool assumes that only public commits (Fix, Feature, Other) will be isolated. I'll fix that.
> 
> Am I right?

Yeah, it should be fixed with https://github.com/ckeditor/ckeditor5-dev/pull/640/commits/2a5eeb118409df871a5ff429b663abb23f92af30.

It happens because we ignored non-public changes.

mlewand:
> It looks to me that we're missing invalid changelog entries with wrong package name.
> 
> An example is: [ckeditor/ckeditor5@dbf24a2](https://github.com/ckeditor/ckeditor5/commit/dbf24a2)
> 
> See that it lists "Fix (media-**mebed**): The media", which is not caught by the tool.

Actually it's my bad, it is listed in the screen below so that's fine :+1:

pomek:
> More of a nitpick: but can we remove this extra empty line after the summary? It's the only one like that and stands out a lot 🙂

Fixed with https://github.com/ckeditor/ckeditor5-dev/pull/640/commits/715e684e011d5577780dafb6eced99cd27a590c1.

pomek:
> That wasn't perceivable when using the tool. "group changes by commit hash and separate them on the list" sounds on paper like a good balance between usability and time/effort required 👍

image

Implemented in https://github.com/ckeditor/ckeditor5-dev/pull/640/commits/89010160ebd04911483dfaef9c25bef2c985caa1

pomek:
> The thing is that you don't even get these minor BC listed beforehand so it's not relevant.
>
> Either minor BCs should be listed (which make sense too) or this step should be skipped in case there's 0 major BCs.

In fact, pressing [Enter] when you see the message does not cost too much. 

On the other hand, if you still would like to make the release as a MAJOR bump (for some reason), you can answer the question. The tool will suggest the next major version. You don't need to go through all packages and search for the version (but it isn't a problem when we have all the same majors). It will also suggest the bump in all packages so instead of typing "x.y.z" all the time, you will hit Enter x times.

Let me know should it stays or I should improve the flow.

For the rest comments - I checked the entire topic and it seems like we are ready for review once again.

PR #648, opened 2020-06-08
Changelog generator - "skip" breaks the process

pomek:

Fix: When typed "skip" as a new version, the changelog generator should abort the process instead of writing invalid entries to the changelog file. Closes ckeditor/ckeditor5#7402.


ma2ciek:
IMO `throw`s shouldn't be used for making `if`s in the code :P But as this is a small change and the whole code needs to be rewrite into async/await in the future I can approve this change.

PR #656, opened 2020-06-29
Changelog for monorepo - use the same version for all packages

pomek:
The changelog generator will use the same version for all packages in the mono repository project.

Other (env): The changelog generator for mono repository will use the same version for all packages. On the screen, a user will see all changes: `MAJOR BREAKING CHANGES`, `MINOR BREAKING CHANGES`, and all commits since the last release. The user must review it and provide the version. Closes ckeditor/ckeditor5#7323.

MAJOR BREAKING CHANGES (env): Task `generateChangelogForMonoRepository()` will generate the changelog uses the same version for all packages.


pomek:
@mlewand, sorry for making the confusion.

I changed the flow because we do not need to display all changes per package. If we want to propose the single version for all packages, I assumed that displaying everything once will be enough.

I know, when you see ~100 commits on the screen, it could be hard to parse. We can improve the look of the list. But how?

The tool didn't suggest a new version based on commits. I've fixed that.

mlewand:
All right, I guess that the idea might work :ok\_hand: as far as improving the list goes, I think that ordering changelog entries by package name would improve it a bit.

During the release when reviewing the changelog it makes a lot of sense see these grouped for readability, just like we do in the final changelog.

PR #659, opened 2020-07-29
Introduced support for "@skipsource" annotation

pomek:

Feature (jsdoc-plugins): Support for the `@skipsource` annotation.

Internal (docs): Added the plugin that supports the `@skipsource` annotation.

---

### Additional information

By adding the `@skipsource` annotation you can hide `See source` link in API docs produced by Umberto.

### DIFF
--- a/packages/ckeditor5-dev-docs/lib/index.js
+++ b/packages/ckeditor5-dev-docs/lib/index.js
@@ -54,6 +54,7 @@ async function build( config ) {
 			require.resolve( '@ckeditor/jsdoc-plugins/lib/export-fixer/export-fixer' ),
 			require.resolve( '@ckeditor/jsdoc-plugins/lib/custom-tags/error' ),
 			require.resolve( '@ckeditor/jsdoc-plugins/lib/custom-tags/observable' ),
+			require.resolve( '@ckeditor/jsdoc-plugins/lib/custom-tags/skipsource' ),
 			require.resolve( '@ckeditor/jsdoc-plugins/lib/observable-event-provider' ),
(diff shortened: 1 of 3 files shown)


ma2ciek:
I guess that this change is needed only to mirror the JSDuck's ability to display it? (cksource/umberto#864)

ma2ciek:
I'll be implementing a similar thing for the CKEditor 5 closed source code but rather based on paths to not blow the code as every single comment would need  annotation.

pomek:
@ma2ciek, it will be enough if you wrap the entire `@module` block with the `@skipsource` annotation.

ma2ciek:
OK, but that's still hundreds of files where you can forget to add it.

ma2ciek:
So if that change is not crucial for now I'd wait a bit.

pomek:
The solution was proposed by @f1ames and I'm assuming the annotation is the best and the easiest way to mark pieces of the code that shouldn't display the "See source" button across different projects and engines that parses comment blocks.

Keep in mind that non-developers modifying our docs as well and the solution should be intuitive and simple for everybody.

ma2ciek:
> Keep in mind that non-developers modifying our docs as well and the solution should be intuitive and simple for everybody.

I agree with you 100%. Thus, not writing anything is better than having to know which annotation and where should be needed :) It could be e.g. based on the `pirvate: true` filed in the package.json file, etc.

ma2ciek:
@pomek, do you want to keep it open? For me, the current behavior works fine.

pomek:
I was using the solution from PR for testing purposes on Umberto side. If we don't need it, the PR can be closed.

pomek:
Closing since we won't merge it anyway.

PR #662, opened 2020-07-30
Add support for the `translationsOutputFile` option allowing specifying target bundle for translations

ma2ciek:

Feature: Add support for the `translationsOutputFile` option for `CKEditorWebpackPlugin` allowing specifying the target bundle for translations. Closes ckeditor/ckeditor5#7688.

---

### Additional information

*   The `CKEditorWebpackPlugin` won't longer complain when no asset will be available during the build
*   Requires ckeditor/ckeditor5#7755.

### DIFF
--- a/packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
+++ b/packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
@@ -190,7 +190,7 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 	 * @fires error
 	 * @param {Object} options
 	 * @param {String} options.outputDirectory Output directory for the translation files relative to the output.
-	 * @param {String[]} options.compilationAssetNames Original asset names from the compiler (e.g. Webpack).
+	 * @param {Array.<String>} options.compilationAssetNames Original asset names from the compiler (e.g. Webpack).
 	 * @returns {Array.<Object>} Returns new and modified assets that will be added to original ones.
 	 */
 	getAssets( { outputDirectory, compilationAssetNames } ) {
@@ -238,7 +238,7 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 	/**
 	 * @param {Object} options
 	 * @param {String} options.outputDirectory Output directory for the translation files relative to the output.
-	 * @param {String[]} options.compilationAssetNames Original asset names from the compiler (e.g. Webpack).
+	 * @param {Array.<String>} options.compilationAssetNames Original asset names from the compiler (e.g. Webpack).
 	 */
 	_getAssetsWithTranslationsBundledToTheOutputFile( { outputDirectory, compilationAssetNames } ) {
 		const assetName = match( this._translationsOutputFile, compilationAssetNames );
@@ -325,7 +325,7 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 	 *
 	 * @private
 	 * @param {String} language The target language
-	 * @param {String[]} sortedMessageIds An array of sorted message ids.
+	 * @param {Array.<String>} sortedMessageIds An array of sorted message ids.
 	 * @returns {Object.<String,String|String[]>}
 	 */
 	_getTranslations( language, sortedMessageIds ) {
@@ -392,7 +392,7 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 /**
  *
  * @param {String|((name: string) => boolean)|RegExp} predicate
- * @param {String[]} options
+ * @param {Array.<String>} options
  * @returns {String|undefined}
  */
 function match( predicate, options ) {
--- a/packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
+++ b/packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
@@ -393,7 +393,7 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 };
 
 /**
- * @param {String|((name: string) => boolean)|RegExp} predicate
+ * @param {String|Function|RegExp} predicate
  * @param {Array.<String>} options
  * @returns {String|undefined}
  */
--- a/packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
+++ b/packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
@@ -194,9 +194,6 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 	 * @returns {Array.<Object>} Returns new and modified assets that will be added to original ones.
 	 */
 	getAssets( { outputDirectory, compilationAssetNames } ) {
-		compilationAssetNames = compilationAssetNames
-			.filter( name => name.endsWith( '.js' ) );
-
 		let bundledLanguage = this._mainLanguage;
 
 		if ( compilationAssetNames.length === 0 ) {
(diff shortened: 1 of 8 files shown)


ma2ciek:
@pomek, could you test it on the Vue integration by following the steps from https://github.com/ckeditor/ckeditor5/pull/7755/files#diff-0acf7b6d2c9b5050039b3d5741919a61? I mean both modes (`yarn serve` and `yarn build`)

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -23,15 +23,17 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 	 * @param {Boolean} [options.compileAllLanguages] When set to `true` languages will be found at runtime.
 	 * @param {Boolean} [options.addMainLanguageTranslationsToAllAssets] When set to `true` the service will not complain
 	 * about multiple JS assets and will output translations for the main language to all found assets.
-	 * @param {Boolean} [buildAllTranslationsToSeparateFiles] When set to `true` the service will output all translations
+	 * @param {Boolean} [options.buildAllTranslationsToSeparateFiles] When set to `true` the service will output all translations
 	 * to separate files.
+	 * @param {String|Function|RegExp} [options.translationsOutputFile]
```
pomek, line 28: What does it do?

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -239,6 +235,30 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 		];
 	}
 
+	/**
+	 * @param {Object} options
+	 * @param {String} options.outputDirectory Output directory for the translation files relative to the output.
+	 * @param {String[]} options.compilationAssetNames Original asset names from the compiler (e.g. Webpack).
```
pomek, line 241: `s/String[]/Array.<String>`

pomek: Missing the `@returns` tag.

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -305,7 +325,7 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 	 *
 	 * @private
 	 * @param {String} language The target language
-	 * @param {String} sortedMessageIds An array of sorted message ids.
+	 * @param {String[]} sortedMessageIds An array of sorted message ids.
```
pomek, line 328: `s/String[]/Array.<String>`

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -305,7 +325,7 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
...
 	 * @returns {Object.<String,String|String[]>}
```
pomek, line 329: Out of scope of this PR, but `s/String[]/Array.<String>`.

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -368,3 +388,28 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 		return path.join( pathToPackage, 'lang', 'translations' );
 	}
 };
+
+/**
+ *
+ * @param {String|((name: string) => boolean)|RegExp} predicate
+ * @param {String[]} options
+ * @returns {String|undefined}
+ */
```
pomek, line 397: `String|RegExp|Function` would be enough. OTOH, why do we need to support such complex construction?

pomek: `s/String[]/Array.<String>`

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -368,3 +388,28 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 		return path.join( pathToPackage, 'lang', 'translations' );
 	}
 };
+
+/**
+ *
```
pomek, line 393:
```suggestion
```

pomek: Or describe the util if something could be unclear.

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -368,3 +388,28 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
...
+		return options.find( option => predicate === option );
+	}
+
+	if ( predicate instanceof RegExp ) {
+		return options.find( option => predicate.test( option ) );
+	}
+
+	throw new Error(
+		'The CKEditorWebpackPlugin matching function got an unsupported type of a predicate.' +
+		`Got ${ predicate } (${ typeof predicate } ) where supported values are only 'string', 'regexp' and 'function'`
```
pomek, line 413:
```suggestion
		`Got '${ predicate }' (${ typeof predicate } ) where supported values are only 'string', 'regexp' and 'function'.`
```

packages/ckeditor5-dev-webpack-plugin/lib/index.js
```diff
@@ -31,24 +31,13 @@ const MultipleLanguageTranslationService = require( '@ckeditor/ckeditor5-dev-uti
...
-	 * @param {Boolean} [options.buildAllTranslationsToSeparateFiles] An option that makes all translations output to separate files.
-	 * @param {String} [options.sourceFilesPattern] An option that allows override the default pattern for CKEditor 5 source files.
-	 * @param {String} [options.packageNamesPattern] An option that allows override the default pattern for CKEditor 5 package names.
-	 * @param {String} [options.corePackagePattern] An option that allows override the default CKEditor 5 core package pattern.
+	 * @param {CKEditorWebpackPluginOptions} [options] Plugin options.
 	 */
-	constructor( options = {} ) {
+	constructor( options ) {
+		if ( !options ) {
+			throw new Error( 'The `CKEditorWebpackPlugin` plugin requires an object.' );
```
pomek, line 38: Docs says something else:

```
* @param {CKEditorWebpackPluginOptions} [options] Plugin options.
```

```diff
-* @param {CKEditorWebpackPluginOptions} [options] Plugin options.
+* @param {CKEditorWebpackPluginOptions} options Plugin options.
```

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
ma2ciek: Not sure if we document the `return` everywhere. Especially for such places that the code is rather internal. I can specify only the type as the function name suggests what it returns.

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
ma2ciek: We need to support at least strings and regexps. We need strings for specifying the name that might not exist and regexp for matching the existing name. I added support for function as it always allows for full customization.

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
pomek: If it returns Object, just put `@returns {Object}`. We don't need digging into its details.

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
```diff
@@ -368,3 +391,27 @@ module.exports = class MultipleLanguageTranslationService extends EventEmitter {
 		return path.join( pathToPackage, 'lang', 'translations' );
 	}
 };
+
+/**
+ * @param {String|((name: string) => boolean)|RegExp} predicate
```
pomek, line 396:
```suggestion
 * @param {String|Function|RegExp} predicate
```

packages/ckeditor5-dev-utils/lib/translations/multiplelanguagetranslationservice.js
pomek: I would put the comment in the code. In the future, it might help when we will be wondering why it's done this way and what we wanted to resolve.

PR #664, opened 2020-08-19
Provide changes required for env unification

ma2ciek:

Feature (tests): Manual test server accepts a new flag: `--identity-file` (alias: `-i`) that allows defining global constants in manual tests.

Other (tests): Manual test script will not search for tests in the `./manual/_utils` directory. If any of a manual test requires additional utils to work, those can be placed in the `_utils` directory.

Feature (jsdoc-plugins): Introduced a plugin that hides the package documentation if in the package's `package.json` file the `private` key is set to `true`. However, by adding the `@publicApi` annotation, you can mark blocks of the code that should not be hidden.

---

### Additional information

* A new `purge-private-api-docs` JSDoc plugin was introduced to remove private docs not ready to be public.


pomek (approved):
LGTM.

PR #670, opened 2020-09-03
Added support for the "--strict" flag while building documentation

pomek:

Feature (docs): Added support for the `--strict` flag while building documentation. See ckeditor/ckeditor5#7994.


mlewand (approved):
Works fine, I like the fact that it does an early return rather than wait for whole thing to build.

Though changelog entry should not say about a flag as we're not adding flag support here (it's in another repo).

PR #674, opened 2020-10-15
`getNewVersionType()` will suggest proper version for "BREAKING CHANGES"

pomek:

Fix (env): The `getNewVersionType()` util will return a proper version when generating the changelog for a single package. Closes ckeditor/ckeditor5#8265.

Feature (env): `generateChangelogForSinglePackage()` accepts an optional option: `options.releaseBranch` (which is equal `master` by default).

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogforsinglepackage.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogforsinglepackage.js
@@ -18,7 +18,7 @@ const getNewVersionType = require( '../utils/getnewversiontype' );
 const getCommits = require( '../utils/getcommits' );
 const getWriterOptions = require( '../utils/getwriteroptions' );
 const { getRepositoryUrl } = require( '../utils/transformcommitutils' );
-const transformCommitForSubRepositoryFactory = require( '../utils/transformcommitfactory' );
+const transformCommitFactory = require( '../utils/transformcommitfactory' );
 
 const SKIP_GENERATE_CHANGELOG = 'Typed "skip" as a new version. Aborting.';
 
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogforsinglepackage.js
```diff
@@ -45,12 +47,13 @@ module.exports = function generateChangelogForSinglePackage( options = {} ) {
...
 
-	const transformCommit = transformCommitForSubRepositoryFactory();
+	const transformCommit = transformCommitFactory();
 
 	logProcess( 'Collecting all commits since the last release...' );
 
 	const commitOptions = {
-		from: options.from ? options.from : 'v' + pkgJson.version
+		from: options.from ? options.from : 'v' + pkgJson.version,
+		releaseBranch: options.releaseBranch
```
ma2ciek, line 56: Should not it be `releaseBranch: options.releaseBranch || master`?

ma2ciek:
LGTM

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogforsinglepackage.js
pomek: I did the same way as we have for the mono repository.

https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogformonorepository.js#L100-L103

However, I am not sure I did it well.

ma2ciek: Maybe you could add a simple test showing that an error is thrown when the current branch does not match the `releaseBranch`?

pomek: https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-env/tests/release-tools/utils/getcommits.js#L78-L247

ma2ciek: Hm. Okey. Now the `undefined` can be passed to the `options.releaseBranch` but I guess it will work fine - https://github.com/ckeditor/ckeditor5-dev/blob/67567dcb6da6ff1d69d1e2ec06c507c35c392f88/packages/ckeditor5-dev-env/lib/release-tools/utils/getcommits.js.

PR #686, opened 2021-01-11
Changed a way how the packages are grouped in the mono-repository changelog

pomek:

Fix (env): Changed packages in the generated changelog will be grouped to proper sections by exact comparing their scopes instead of matching values. Closes ckeditor/ckeditor5#8596.

---

### Additional information

The case mentioned in the issue was resolved:

image


mlewand (approved):
Works fine 👍

PR #689, opened 2021-01-27
Support for collecting commits from external repositories in monorepo-changelog

pomek:

Feature (env): Support for collecting commits from additional mono-repositories in the `generateChangelogForMonoRepository()` function.

Thanks to the new option called `options.externalRepositories` which is an array of the `ExternalRepository` objects, you can specify additional repositories that should be checked while generating the changelog.

The `ExternalRepository` object requires the following properties:

* `{String}` `cwd` - An absolute path to the repository.
* `{String}` `packages` - Where to look for packages (a relative path to `cwd`)

The following properties are optional:

* `{String}` `scope` - Package names have to match to specified glob pattern in order to be processed.
* `{Array.<String>}` `skipPackages` - Name of packages which won't be touched.
* `{Boolean` `skipLinks` - If set on `true`, a URL to commit (hash) will be omitted.
* `{String}` `from` - A commit or tag name that will be the first param of the range of commits to collect. If not specified, the option will inherit its value from the function's `options` object.
* `{String}` `releaseBranch` A name of the branch that should be used for releasing packages. If not specified, the branch used for the main repository will be used.

---

### Additional information

While merging, copy the above description to the merge message.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogformonorepository.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogformonorepository.js
@@ -707,9 +707,9 @@ module.exports = function generateChangelogForMonoRepository( options ) {
  *
  * @param {String} cwd An absolute path to the repository.
  *
- * @param {String} packages Where to look for packages.
+ * @param {String} packages Subdirectory in a given `cwd` that should searched for packages. E.g. `'packages'`.
  *
- * @param {String} [scope] Package names have to match to specified glob pattern in order to be processed.
+ * @param {String} [scope] Glob pattern for package names to be processed.
  *
  * @param {Array.<String>} [skipPackages] Name of packages which won't be touched.
  *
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogformonorepository.js
```diff
@@ -592,3 +701,23 @@ module.exports = function generateChangelogForMonoRepository( options ) {
  *
  * @param {Boolean} next The next version defined during generating the changelog file.
  */
+
+/**
+ * @typedef {Object} ExternalRepository
+ *
+ * @param {String} cwd An absolute path to the repository.
+ *
+ * @param {String} packages Where to look for packages.
```
mlewand, line 710: `"packages"`.
```suggestion
 * @param {String} packages Subdirectory in a given `cwd` that should searched for packages. E.g. `'packages'`.
```

packages/ckeditor5-dev-env/lib/release-tools/tasks/generatechangelogformonorepository.js
```diff
@@ -592,3 +701,23 @@ module.exports = function generateChangelogForMonoRepository( options ) {
...
+ *
+ * @param {String} [scope] Package names have to match to specified glob pattern in order to be processed.
```
mlewand, line 712:
```suggestion
 * @param {String} [scope] Glob pattern for package names to be processed.
```

mlewand (approved):
Checked and it works fine. I just proposed minor corrections to the docs.

PR #690, opened 2021-02-08
Release tools can handle mono-repo that is a part of other mono-repo

pomek:

Feature (env): The `bumpVersions()` function can read the changelog file from an external directory (using `options.changelogDirectory`). Thanks to that, the private packages from the mono-repository can be released using a public changelog.

Feature (env): The `bumpVersions()` function allows skipping upgrading versions of dependencies between updated packages (using `options.skipUpdatingDependencies` which is `false` by default).

Fix (env): Simplified a check whether a package was published on NPM in the `releaseSubRepositories()` function. 

Other (env): The `bumpVersions()` function returns a promise with a collection that contains all updated packages. Previously the promise didn't resolve anything.

---

### Additional information

I'd like to do a live review because those scripts require much work that I don't want to share here.


mlewand (approved):
Tested and works fine 👍

PR #695, opened 2021-04-22
Update postcss to postcss 8

alexander-schranz:

Type: Message. Closes #000.

---

### Additional information


packages/ckeditor5-dev-utils/package.json
```diff
@@ -14,14 +14,14 @@
...
-    "postcss": "^7.0.17",
-    "postcss-import": "^12.0.0",
-    "postcss-loader": "^3.0.0",
-    "postcss-mixins": "^6.2.0",
-    "postcss-nesting": "^7.0.0",
+    "postcss": "^8.2.12",
+    "postcss-import": "^14.0.1",
+    "postcss-loader": "^4.2.0",
+    "postcss-mixins": "^7.0.3",
+    "postcss-nested": "^5.0.5",
```
alexander-schranz, line 21: Think here https://github.com/postcss/postcss-nested is better to be used instead of to none official https://github.com/csstools/postcss-nesting

packages/ckeditor5-dev-utils/package.json
```diff
@@ -14,14 +14,14 @@
...
     "javascript-stringify": "^1.6.0",
     "pofile": "^1.0.9",
-    "postcss": "^7.0.17",
-    "postcss-import": "^12.0.0",
-    "postcss-loader": "^3.0.0",
-    "postcss-mixins": "^6.2.0",
-    "postcss-nesting": "^7.0.0",
+    "postcss": "^8.2.12",
+    "postcss-import": "^14.0.1",
+    "postcss-loader": "^4.2.0",
```
pomek, line 19: This is a huge breaking change and many of our tools will not work – https://github.com/webpack-contrib/postcss-loader/releases/tag/v4.0.0.

pomek: This upgrade will require updating all webpack loaders we have (including these in [the main repository](https://github.com/ckeditor/ckeditor5)).
I'd downgrade it to the previous version. The rest of your changes look good.

alexander-schranz: I'm aware that this is a BC Break and this can only be part of the next major release.

I will have a look at the main repository didn't know that there are also postcss packages required thought it is all handled by this package. Whats the best way to test it as it seems the tests don't cover that? Is there any way to create a PR in the ckeditor repository that this version of ckeditor5-dev-utils is used (maybe when used the same branch name)?

We would like to move step by step to webpack 5. The postcss-loader 4 supports both 4 and 5 of webpack version. I think it would be great if ckeditor supports both webpack 4 and 5.

alexander-schranz: > I'd downgrade it to the previous version.

Then the only change in this PR would be `postcss-nesting` -> `postcss-nested`. As the `postcss-loader` 4 is the only one supporting `postcss` `8` and the plugin updates requires postcss 8.

pomek: Updating to `postcss` 8 requires reviewing guides in the documentation (many code snippets assume using `postcss` < 8). Also, our internal tool that prepares the documentation must be synchronized with that change. That's why I'd like to avoid major BC.

Could you describe your use case, why the change is important? I may not know the context. Hence you can make an impression that I'm against the change.

pomek: Btw, the editor build using webpack 5 should work. Please, see https://github.com/ckeditor/ckeditor5-dev/releases/tag/v24.3.0.

alexander-schranz: We are first trying to update to postcss 8 and postcss-loader 4 requires it which is the one supporting webpack 5. And had then issues here: https://github.com/sulu/sulu/pull/5549 as written there the closest issue we could found was: https://github.com/ckeditor/ckeditor5/issues/8112. 

And I thought the problems is that ckeditor is not yet build with postcss 8 and I think all dependencies need to be the same to get this work correctly.

pomek: That's reasonable. Unfortunately, I cannot give you any ETA when we can finally migrate to webapck 5. We focus on other things right now (please, see the iteration roadmap – https://github.com/ckeditor/ckeditor5/issues/9578), so the upgrade to webpack 5 will not happen soon. However, it's an important thing for integrators, and we will try the best we can do to unblock the upgrade.

packages/ckeditor5-dev-utils/package.json
```diff
@@ -14,14 +14,14 @@
...
+    "postcss-mixins": "^7.0.3",
```
keillera, line 20: There seems to be a security issue with a module  dependent on  postcss-mixins.
I would appreciate it a lot if you could raise the priority of this.

--
@ckeditor/ckeditor5-dev-utils@25.4.2 requires glob-parent@^3.1.0 via a transitive dependency on fast-glob@2.2.7
[CVE-2020-28469](https://github.com/advisories/GHSA-ww39-953v-wcq6)

ThomasTr:
Another CVE:

CVE-2021-23382
Vulnerable versions: < 8.2.13
Patched version: 8.2.13
The package postcss before 8.2.13 are vulnerable to Regular Expression Denial of Service (ReDoS) via getAnnotationURL() and loadAnnotation() in lib/previous-map.js. The vulnerable regexes are caused mainly by the sub-pattern /*\s* sourceMappingURL=(.*).

alexander-schranz:
I'm currently not able to work on this maybe @pomek which did implement the webpack 5 support #729 or somebody else want to take this over. Think it would be great as postcss seems not todo any security release for 7 version any more.

gagrison:
Can you also fix [Inefficient Regular Expression Complexity in nth-check](https://github.com/advisories/GHSA-rp65-9cf3-cjxr) by updating cssnano to at least 5.0.0

pomek:
Upgrade to postcss@8 is covered in https://github.com/ckeditor/ckeditor5-dev/pull/764.

PR #712, opened 2021-06-18
Make translation output to be ES5 compatible.

drumlab:
The translation webpack plugin should output ES5 JS (avoid using `const`) to make the output compatible to more use cases. This is necessary since webpack rules does not apply to webpack plugin output.


ma2ciek:
Hi @drumlab,

Thanks for the contribution. 

Could you sign the CLA?

pomek:
Closing due to lack of activity.

PR #715, opened 2021-07-12
Exit code when `translations:collect` reports an error

pomek:

Other (env): Set the proper exit code when found an error when collecting translations. Closes ckeditor/ckeditor5#10123.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/translations/createpotfiles.js
+++ b/packages/ckeditor5-dev-env/lib/translations/createpotfiles.js
@@ -34,11 +34,16 @@ module.exports = function createPotFiles( {
 	const packageContexts = getPackageContexts( packagePaths, corePackagePath );
 	const sourceMessages = collectSourceMessages( { sourceFiles, logger } );
 
-	const foundErrors = [
+	const errors = [].concat(
 		assertNoMissingContext( { packageContexts, sourceMessages, logger } ),
 		assertAllContextUsed( { packageContexts, sourceMessages, logger } ),
 		assertNoRepeatedContext( { packageContexts, logger } )
-	].filter( errors => errors.length ).length > 0;
+	);
+
+	for ( const error of errors ) {
+		logger.error( error );
+		process.exitCode = 1;
+	}
 
 	removeExistingPotFiles();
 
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-dev-env/lib/translations/createpotfiles.js
```diff
@@ -135,12 +141,14 @@ function assertNoMissingContext( { packageContexts, sourceMessages, logger } ) {
...
 	for ( const sourceMessage of sourceMessages ) {
 		if ( !contextIdOrigins.has( sourceMessage.id ) ) {
-			logger.error(
-				`Context for the message id is missing ('${ sourceMessage.id }' from ${ sourceMessage.filePath }).`
-			);
+			errors.push( `Context for the message id is missing ('${ sourceMessage.id }' from ${ sourceMessage.filePath }).` );
 		}
 	}
 
+	for ( const error of errors ) {
```
mlewand, line 148: This part is repeated in all three functions. Wouldn't it be better to do this logic outside of `assert*` functions once?

Instead of

```
const foundErrors = [
		assertNoMissingContext( { packageContexts, sourceMessages, logger } ),
		assertAllContextUsed( { packageContexts, sourceMessages, logger } ),
		assertNoRepeatedContext( { packageContexts, logger } )
	].filter( errors => errors.length ).length > 0;
```

one could do:

```
const errors = [].concat(
		assertNoMissingContext( { packageContexts, sourceMessages, logger } ),
		assertAllContextUsed( { packageContexts, sourceMessages, logger } ),
		assertNoRepeatedContext( { packageContexts, logger } )
	);

// (...)

for ( const error of errors ) {
	logger.error( error );

	process.exitCode = 1;
}
```

mlewand (commented):
The solution looks to be working just fine. I just have one question.

packages/ckeditor5-dev-env/lib/translations/createpotfiles.js
pomek: Done in 03983fb.

mlewand (approved):
Looking good 👌

PR #716, opened 2021-07-13
Fixed incorrect path interpretation for theme-lark package

przemyslaw-zan:
Fix (tests): Prevented `isDevDependency()` function from reading "theme" word from the `ckeditor5-theme-lark` package name. See ckeditor/ckeditor5#9998.

### DIFF
--- a/packages/ckeditor5-dev-tests/bin/check-dependencies.js
+++ b/packages/ckeditor5-dev-tests/bin/check-dependencies.js
@@ -428,7 +428,7 @@ function findMisplacedDependencies( dependencies, devDependencies, dependenciesT
  * @returns {Boolean}
  */
 function isDevDependency( absolutePaths ) {
-	return !absolutePaths.some( absolutePath => absolutePath.match( /[/\\](src|theme)/ ) );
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-dev-tests/bin/check-dependencies.js
```diff
@@ -428,7 +428,7 @@ function findMisplacedDependencies( dependencies, devDependencies, dependenciesT
  * @returns {Boolean}
  */
 function isDevDependency( absolutePaths ) {
-	return !absolutePaths.some( absolutePath => absolutePath.match( /src|theme/ ) );
+	return !absolutePaths.some( absolutePath => absolutePath.match( /[/\\](src|theme)/ ) );
```
psmyrek, line 431: I would just make sure we are checking the whole path segments. It does not cost anything but we will be a little safer:

```suggestion
	return !absolutePaths.some( absolutePath => absolutePath.match( /[/\\](src|theme)[/\\]/ ) );
```

psmyrek (commented):
Found one thing to consider.

psmyrek (approved):
LGTM 🎉

PR #719, opened 2021-08-09
Translations for DLL (dev)

pomek:

Other (utils): The webpack configuration for DLL builds now produces the translation files.

Feature (webpack-plugin): Introduced several new options that improve the output files produced by the plugin.

New options:
  * `corePackageContextsResourcePath` - (optional) a path to the file where all translation contexts are specified for the `ckeditor5-core` package. Defaults to `'@ckeditor/ckeditor5-core/lang/contexts.json'`.
  * `includeCorePackageTranslations` - (optional) a flag that determines whether all translations found in the core package should be added to the output bundle file. If set to `true`, translations from the core package will be saved even if are not used in the source code (*.js files). Defaults to `false`.
  * `skipPluralFormFunction`- (optional) a flag that determines whether the `getPluralForm()` function should not be added in the output bundle file. Defaults to `false`.

---

### Additional information

Required for: ckeditor/ckeditor5#10330.


psmyrek (approved):
LGTM 🎉

przemyslaw-zan (approved):
LGTM

PR #722, opened 2021-08-18
Fixed yarn clean script for Windows OS

przemyslaw-zan:
Fix (utils): The `tools.clean()` function will resolve paths correctly on Windows environments. Closes ckeditor/ckeditor5#10141.

### DIFF
--- a/packages/ckeditor5-dev-utils/tests/tools.js
+++ b/packages/ckeditor5-dev-utils/tests/tools.js
@@ -553,8 +553,8 @@ describe( 'utils', () => {
 
 		describe( 'clean', () => {
 			const files = [
-				path.posix.join( 'test', 'foo', 'bar' ),
-				path.posix.join( 'test', 'bar', 'foo' )
+				path.join( 'test', 'foo', 'bar' ),
+				path.join( 'test', 'bar', 'foo' )
 			];
 
 			let delArg;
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-dev-utils/tests/tools.js
```diff
@@ -577,6 +577,17 @@ describe( 'utils', () => {
...
 			} );
+
+			it( 'works with paths that contain backslash', () => {
+				return tools.clean( 'test\\path', '**\\packages' )
+					.then( () => {
+						expect( delArg ).to.equal( path.join( 'test/path', '**/packages' ) );
+						expect( loggerVerbosity ).to.equal( 'info' );
+						expect( infoSpy.calledTwice ).to.equal( true );
+						expect( infoSpy.firstCall.args[ 0 ] ).to.match( new RegExp( files[ 0 ] ) );
+						expect( infoSpy.secondCall.args[ 0 ] ).to.match( new RegExp( files[ 1 ] ) );
```
psmyrek, line 588: These lines are not related to the subject of this test case - you are testing here if the joined path, that is sent to the `del` package, contains only slashes, but the `infoSpy` still operates on the `files` array that does not match the provided path `'test/path/**/packages'` at all. So comparing the `infoSpy` args with the `files` when the joined path does not match them is confusing.

psmyrek (changes requested):
It seems that it works fine 🎉.

I've just found one minor thing to change.

packages/ckeditor5-dev-utils/tests/tools.js
przemyslaw-zan: Addresed

psmyrek (approved):
LGTM 🎉

PR #725, opened 2021-09-15
Fixed majority of tests on Windows

przemyslaw-zan:

Tests: Fixed majority of tests on Windows. Closes ckeditor/ckeditor5#10388.

---

### Additional information

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "homepage": "https://github.com/ckeditor/ckeditor5-dev#readme",
   "scripts": {
-    "test": "mocha packages/*/tests --recursive --timeout 5000",
+    "test": "mocha 'packages/*/tests/**/*.js' --recursive --timeout 5000",
     "coverage": "nyc --reporter=lcov --reporter=text-summary mocha packages/*/tests -- --recursive --timeout 5000",
     "changelog": "node ./scripts/changelog.js",
     "release:bump-version": "node ./scripts/bump-versions.js",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "homepage": "https://github.com/ckeditor/ckeditor5-dev#readme",
   "scripts": {
-    "test": "mocha 'packages/*/tests/**/*.js' --recursive --timeout 5000",
+    "test": "mocha 'packages/*/tests/**/*.js' --timeout 5000",
     "coverage": "nyc --reporter=lcov --reporter=text-summary mocha packages/*/tests -- --recursive --timeout 5000",
     "changelog": "node ./scripts/changelog.js",
     "release:bump-version": "node ./scripts/bump-versions.js",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
   },
   "homepage": "https://github.com/ckeditor/ckeditor5-dev#readme",
   "scripts": {
-    "test": "mocha packages/*/tests --recursive --timeout 5000",
+    "test": "mocha 'packages/*/tests/**/*.js' --recursive --timeout 5000",
```
pomek, line 30:
```suggestion
    "test": "mocha 'packages/*/tests/**/*.js' --timeout 5000",
```

It seems that the `--recursive` option is not needed.

packages/ckeditor5-dev-tests/tests/utils/manual-tests/compilehtmlfiles.js
```diff
@@ -14,6 +14,13 @@ const chokidar = require( 'chokidar' );
 const fakeDirname = path.dirname( require.resolve( '../../../lib/utils/manual-tests/compilehtmlfiles' ) );
 
 describe( 'compileHtmlFiles', () => {
+	// This test is disabled on Windows
+	// See https://github.com/ckeditor/ckeditor5/issues/10538
+	if ( process.platform === 'win32' ) {
+		console.log( 'This test is disabled on Windows' );
+		return;
+	}
```
pomek, line 22: It would be good to explain why the test is disabled.

```suggestion
	// Temporary these tests are disabled on Windows.
	// See https://github.com/ckeditor/ckeditor5/issues/10538.
	if ( process.platform === 'win32' ) {
		console.log( 'The "compileHtmlFiles()" tests do not work on Windows environments.' );
		console.log( 'See: https://github.com/ckeditor/ckeditor5/issues/10538 for more details.' );

		return;
	}
```

pomek (commented):
TBH, I wouldn't say I like that we have changes in `lib/*.js` files. I'd expect that only `tests/*.js` files will be fixed by specifying proper mocks.

These changes shouldn't break tools on Unix systems, but I'm unsure whether tools will work on Windows.

pomek:
Which program (terminal) should I use for testing changes? In the past, I used Git bash. Will it be ok?

przemyslaw-zan:
`Git bash` might not catch all possible problems that might occur on Windows platform, since some commands are different depending on the terminal - during my testing different errors popped up depending on the terminal choice. I'd go with the `Powershell` or `cmd.exe`, although ideally `bash` could be tested as well, just to make sure.

pomek:
Please, do not force me to use `cmd.exe` :D 

I use Powershell.

pomek:
image

Green.

pomek:
It's nice to see that these tests also pass on Windows. Thanks @przemyslaw-zan!

PR #734, opened 2021-11-22
Remove instance/static symbols when comparing doclets

pomek:

Fix (jsdoc-plugins): When detecting doclets of the same member, remove a prefix for instance/static symbols. Closes ckeditor/ckeditor5#8691.

---

### Additional information

The goal is to generate proper docs for the `DataController#set` method.

image

But the PR fixes all cases when a mixed method is overridden by a class. However, it's hard to say do we have more cases.


pomek:
Steps to verify:

* `ckeditor5-dev`:
	* checkout to a branch from this PR
* `ckeditor5`:
	* `yarn link @ckeditor/ckeditor5-dev-docs; yarn link @ckeditor/jsdoc-plugins`
	* `yarn run docs --skip-guides`
	* Review API docs.

PR #737, opened 2021-12-14
Created release script that updates all `ckeditor5` dependencies to the latest version.

przemyslaw-zan:
Feature (env): Created a script for updating versions of `ckeditor5` and `@ckeditor/ckeditor5-*` dependencies to a specified version.


packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+
+	try {
+		console.log( '\n📍 ' + chalk.blue( 'Updating \'packages/*\'...' ) );
+		const packagesChanged = updateDirectory( pathToPackages );
+
+		if ( packagesChanged ) {
+			console.log( '\n📍 ' + chalk.blue( 'Committing the changes...\n' ) );
+
+			execSync( 'git add -A', execOptions );
+			execSync( 'git commit -m "Updated all ckeditor5 dependencies in `packages/*` to the latest version."', execOptions );
```
psmyrek, line 43:
```suggestion
			execSync( 'git commit -m "Internal: Updated all CKEditor 5 dependencies in `packages/*` to the latest version. [skip ci]"', execOptions );
```

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+	};
+
+	try {
+		console.log( '\n📍 ' + chalk.blue( 'Updating \'packages/*\'...' ) );
+		const packagesChanged = updateDirectory( pathToPackages );
+
+		if ( packagesChanged ) {
+			console.log( '\n📍 ' + chalk.blue( 'Committing the changes...\n' ) );
+
+			execSync( 'git add -A', execOptions );
```
psmyrek, line 42: If the `dryRun` flag is set, no changes should be commited. Maybe something like here https://github.com/ckeditor/ckeditor5-dev/blob/master/packages/ckeditor5-dev-env/lib/release-tools/tasks/bumpversions.js#L217-L228 could be used instead: list the changes that would be commited if `dryRun` flag was not set.

Besides that, the `git add -A` operates on the entire working tree, and not just on the current path so I think it should not be used here. It would be better to just stage the files that are modified by this script only.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+			execSync( 'git commit -m "Updated all ckeditor5 dependencies in `packages/*` to the latest version."', execOptions );
+		}
+
+		console.log( '\n📍 ' + chalk.blue( 'Updating \'release/*\'...' ) );
+		updateDirectory( pathToRelease );
+
+		console.log( '\n📍 ' + chalk.green( 'Successfully updated package versions!' ) );
```
psmyrek, line 49: What if no packages were updated?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+			console.log( `\n${ chalk.yellow( 'Use' ) } git checkout . ${ chalk.yellow( 'to revert the changes.\n' ) }` );
+		}
+	} catch ( error ) {
+		console.log( '\n📍 ' + chalk.red( 'Updating package versions threw an error:' ) );
+		console.log( chalk.redBright( error ) );
+	}
+};
+
+/**
+ * Updates @ckeditor/ckeditor5-* and ckeditor5 dependencies in the specified directory to the latest version.
```
psmyrek, line 65: Please wrap the **@ckeditor/ckeditor5-*** and **ckeditor5** with \`\` as it is done in the description in the main function. The description here could also contain info that the `version` property from the `package.json` file is used as the "current version", because it has already been bumped.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @param {String} pathToUpdate
+ * @returns {Boolean} True if any files were updated, false otherwise
+ */
+function updateDirectory( pathToUpdate ) {
+	const pathToCheck = pathToUpdate + '/**/package.json';
```
psmyrek, line 71: It looks like there is no need to deeply traverse the provided directory, as the `package.json` file is located in the root folder of each package.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+
+	console.log( `\nLooking for package.json files in ${ pathToCheck }...` );
+
+	const packageJsonArray = glob.sync( pathToCheck ).filter( path => !path.includes( 'node_modules' ) );
```
psmyrek, line 75: As we are using a monorepo with workspaces, all the found paths should never point to the `node_modules` as it should never contain any `package.json` files, so filtering out is not needed. Additionally, if you change the `glob` pattern above, only the `package.json` files from the package's root directory will be found.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+
+	if ( !packageJsonArray.length ) {
+		console.log( 'No files were found.' );
+		return false;
```
psmyrek, line 79: According to the coding style that we have in the CKEditor 5 ecosystem, these 2 lines should be separated by a new line.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	console.log( `Found ${ packageJsonArray.length } files. Updating ckeditor5 dependencies...` );
```
psmyrek, line 82: If the CKEditor 5 name is used as the project name (not the `ckeditor5` dependency), it should be named in the capitalized form: CKEditor 5.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+
+	for ( const file of packageJsonArray ) {
+		const currentFileData = fs.readFileSync( file, 'utf-8' );
+		const parsedData = JSON.parse( currentFileData );
+		const version = parsedData.version;
+
+		for ( const dependency in parsedData.dependencies ) {
```
psmyrek, line 89: What about `devDependencies`?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+			// Update only cke5 deps except the *-dev ones.
+			if ( !/^@ckeditor\/ckeditor5-(?!dev)|^ckeditor5-(?!dev)/.test( dependency ) ) {
```
psmyrek, line 91: The second part of this rule is invalid. There is no such dependency like `ckeditor5-*`, but only `ckeditor5`.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+				continue;
+			}
+
+			parsedData.dependencies[ dependency ] = version;
```
psmyrek, line 95: Should we explicitly define the exact dependency version or use the current version with the `^` range?
Example: `"@ckeditor/ckeditor5-core": "31.1.0"` vs `"@ckeditor/ckeditor5-core": "^31.1.0"`?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		const newFileData = JSON.stringify( parsedData, null, 2 ) + '\n';
+		fs.writeFileSync( file, newFileData, 'utf-8' );
```
psmyrek, line 99: There is no need to save the data if nothing has changed.

psmyrek (changes requested):

psmyrek:
Please also add test in `packages/ckeditor5-dev-env/tests/index.js` for the newly-added task, so the coverage would not be decreased.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
przemyslaw-zan: Citing [https://github.com/cksource/ckeditor5-internal/issues/986#issue-993305327:](https://github.com/cksource/ckeditor5-internal/issues/986#issue-993305327:)

> I don't agree that the user should use `^` in their dependencies. We should not force anyone to update the editor if everything works for them in current version. Also, we don't support our packages working with different versions of our own packages, we require to keep all of them at the same version, so I don't understand why we use `^`.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,105 @@
...
+const path = require( 'path' );
+
+/**
+ * Updates `@ckeditor/ckeditor5-*` and `ckeditor5` dependencies in `packages/*` and `release/*` directories to the latest version.
+ * Changes in `packages/*` will be committed as well.
+ *
+ * See https://github.com/cksource/ckeditor5-internal/issues/1123
+ *
+ * @param {Object} options
+ * @param {String} options.cwd Current working directory (packages) from which all paths will be resolved.
```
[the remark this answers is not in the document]
przemyslaw-zan: This got here by mistake, my bad.

psmyrek (changes requested):

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,153 @@
...
+		console.log( getFeedback( packagesResult ) );
+
+		const releaseResult = updateDirectory( 'release', options.cwd );
+		console.log( getFeedback( releaseResult ) );
+
+		if ( packagesResult.updated ) {
+			console.log( '\n📍 ' + chalk.blue( 'Committing the changes from \'packages/*\'...\n' ) );
+
+			if ( options.dryRun ) {
+				console.log( chalk.yellow( 'DRY RUN mode - no changes will be committed.' ) );
```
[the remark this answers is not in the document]
przemyslaw-zan: I think it's fine like this, since it shows you what WOULD happen normally, but then `dry-run` "interrupts" the process, but if you feel strong about this, i can change this so that only the dry run message is displayed, and update it to say something along the lines of "Changes would now be commited, but this is dry run, so that step is skipped."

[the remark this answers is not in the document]
przemyslaw-zan: As i think about it, You're right, i will update it

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,153 @@
...
+module.exports = function updatePackageVersions( options ) {
+	const execOptions = {
+		stdio: 'inherit',
+		cwd: options.cwd
+	};
+
+	console.log( '\n📍 ' + chalk.blue( 'Updating CKEditor 5 dependencies...\n' ) );
+
+	try {
+		const packagesResult = updateDirectory( 'packages', options.cwd );
```
pomek, line 36: The `'packages'` and `'release'` directories should be specified in `options`, as the rest release tasks do.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,153 @@
...
+ */
+module.exports = function updatePackageVersions( options ) {
+	const execOptions = {
+		stdio: 'inherit',
+		cwd: options.cwd
+	};
+
+	console.log( '\n📍 ' + chalk.blue( 'Updating CKEditor 5 dependencies...\n' ) );
+
+	try {
```
pomek, line 35: Why do we need the try/catch block at all?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,153 @@
...
+		const packagesResult = updateDirectory( 'packages', options.cwd );
+		console.log( getFeedback( packagesResult ) );
+
+		const releaseResult = updateDirectory( 'release', options.cwd );
+		console.log( getFeedback( releaseResult ) );
```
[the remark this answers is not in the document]
pomek: `getFeedback` - what does it mean?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,153 @@
...
+				if ( options.diff ) {
+					console.log( 'git diff --word-diff ./packages' );
+					execSync( 'git diff --word-diff ./packages', execOptions );
+				} else {
+					console.log( 'git status --short ./packages' );
+					execSync( 'git status --short ./packages', execOptions );
+				}
+
+				console.log( chalk.yellow( 'Reverting the changes...' ) );
+				execSync( 'git checkout ./packages', execOptions );
```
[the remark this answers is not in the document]
pomek: One thing that may change the current implementation is that some of the directories are not tracked in git. Hence, reverting changes must be more innovative than calling `git checkout .` as it will not touch untracked files.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
pomek: _Marked as unresolved as the current implementation is not valid._

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+
+const chalk = require( 'chalk' );
+const { diffChars: diff } = require( 'diff' );
+const { execSync } = require( 'child_process' );
+const fs = require( 'fs' );
+const glob = require( 'glob' );
+
+/**
+ * Updates `@ckeditor/ckeditor5-*` and `ckeditor5` dependencies in `packages/*` and `release/*`
+ * directories to the latest version. Changes in `packages/*` will be committed as well.
```
psmyrek, line 16: The description should be adapted, as there are no predefined `packages/*` and `release/*` directories in this script, but all paths to update are passed as an argument. The script does not know in advance which paths will be updated.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+ * instead shows list of files that was changed in `packages/*` directory.
+ * @param {Boolean} options.pathsToUpdate Instead of list of files, it displays detailed
+ * list of changes from each file. Has no effect without options.dryRun.
+ */
+module.exports = function updatePackageVersions( options ) {
+	const totalResult = { found: 0, updated: 0, differences: [] };
+	const pathsToCommit = [];
+
+	options.pathsToUpdate.forEach( ( value, index, array ) => {
+		array[ index ].path = value.path.split( '\u005C' ).join( '\u002F' );
```
psmyrek, line 33: What is this magic? Consider whether it can not be done more simply, without using Unicode escape sequence `\uXXXX` and never modify the original input argument.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+	console.log( '\n📍 ' + chalk.blue( 'Updating CKEditor 5 dependencies...\n' ) );
+
+	for ( const pathToUpdate of options.pathsToUpdate ) {
+		console.log( `Looking for package.json files in '${ pathToUpdate.path }'...` );
+
+		const result = updateDirectory( pathToUpdate.path, options.cwd, options.dryRun );
+
+		totalResult.found += result.found;
+		totalResult.updated += result.updated;
+		if ( result.differences ) {
```
psmyrek, line 45: New line separator is required between these two lines, according to coding convention in CKEditor 5 ecosystem.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * See https://github.com/cksource/ckeditor5-internal/issues/1123
+ *
+ * @param {Object} options
+ * @param {String} options.cwd The root directory of the repository in which
+ * the script is being used.
+ * @param {Boolean} options.dryRun Prevents the script from committing, and
+ * instead shows list of files that was changed in `packages/*` directory.
+ * @param {Boolean} options.pathsToUpdate Instead of list of files, it displays detailed
+ * list of changes from each file. Has no effect without options.dryRun.
```
psmyrek, line 26: ?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+		console.log( '\n📍 ' + chalk.blue( 'Committing the changes...\n' ) );
+
+		for ( const path of pathsToCommit ) {
+			const execOptions = {
+				stdio: 'inherit',
+				cwd: path
+			};
+
+			console.log( chalk.blue( path ) );
+			execSync( `git add ${ path }`, execOptions );
```
psmyrek, line 77: Please add new line separator between these 2 lines.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * Updates `@ckeditor/ckeditor5-*` and `ckeditor5` dependencies in the specified directory to the latest version.
+ * Latest version is taken from the `version` property from the root of the `package.json` file, since that value
+ * should already have been bumped at this point of release process.
+ *
+ * @param {String} pathToUpdate directory containing files to update
+ * @param {String} cwd The root directory of the repository in which
+ * the script is being used.
+ * @returns {Object} Object containing numerical values counting files found and updated, as well as array of all changes made.
```
psmyrek, line 100: Missing description for `dryRun` argument.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+	let updatedFiles = 0;
+	const differences = [];
+
+	for ( const file of packageJsonArray ) {
+		const currentFileData = fs.readFileSync( file, 'utf-8' );
+		const parsedData = JSON.parse( currentFileData );
+		const version = parsedData.version;
+
+		// Update only the CKEditor 5 dependencies, except the *-dev and *-inspector.
+		const regex = /^@ckeditor\/ckeditor5-(?!dev|inspector)|^ckeditor5$/;
```
psmyrek, line 119: Nice catch with the `@ckeditor/ckeditor5-inspector` 👍🏻

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+			parsedData.dependencies[ dependency ] = `^${ version }`;
+		}
+
+		for ( const dependency in parsedData.devDependencies ) {
+			if ( !regex.test( dependency ) ) {
+				continue;
+			}
+
+			parsedData.devDependencies[ dependency ] = `^${ version }`;
+		}
```
psmyrek, line 135: Please merge this logic as it does almost the same thing.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+
+		const newFileData = JSON.stringify( parsedData, null, 2 ) + '\n';
+
+		if ( currentFileData !== newFileData ) {
+			updatedFiles++;
+			differences.push( diff( currentFileData, newFileData ) );
```
psmyrek, line 141: Currently, the diff in your script is needed only if `--dry-run` flag is set, so it does not need to be calculated always.

Or maybe do we want to display the changes always, even after committing them?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+
+	for ( const pathToUpdate of options.pathsToUpdate ) {
+		console.log( `Looking for package.json files in '${ pathToUpdate.path }'...` );
+
+		const result = updateDirectory( pathToUpdate.path, options.cwd, options.dryRun );
+
+		totalResult.found += result.found;
+		totalResult.updated += result.updated;
+		if ( result.differences ) {
+			totalResult.differences = [ ...totalResult.differences, ...result.differences ];
```
psmyrek, line 46: To consider: re-creating the same array with some new items is slower than just pushing new items.

```suggestion
			totalResult.differences.push( ...result.differences );
```

packages/ckeditor5-dev-env/tests/index.js
```diff
@@ -133,6 +135,26 @@ describe( 'dev-env/index', () => {
 		} );
 	} );
 
+	describe( 'updatePackageVersions()', () => {
+		it( 'should update versions in package.json files', () => {
```
psmyrek, line 139: If you could add some new tests it would be great, especially for the conditions you have in the script (i.e. the regex for matching the dependency names).

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,150 @@
...
+			if ( pathToUpdate.commit ) {
+				pathsToCommit.push( pathToUpdate.path );
+			}
+		}
+	}
+
+	if ( options.dryRun ) {
+		console.log( chalk.yellow( 'DRY RUN mode - displaying changes instead of committing.' ) );
+
+		// TODO: Changes will be displayed here
```
psmyrek, line 65: Do we want to show the changes only in `--dry-run` mode?

psmyrek (changes requested):

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
przemyslaw-zan: Added `^` to the verions.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
przemyslaw-zan: Adressed in 0c5b3ca. Function name i came up with is pretty generic, any ideas for improvement in that regard?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,157 @@
...
+ * should already have been bumped at this point of release process.
+ *
+ * @param {String} pathToUpdate directory containing files to update
+ * @param {String} cwd The root directory of the repository in which
+ * the script is being used.
+ * @returns {Object} Object containing numerical values counting files found and updated, as well as array of all changes made.
+ */
+function updateDirectory( pathToUpdate, cwd, dryRun ) {
+	const globPattern = pathToUpdate + '/*/package.json';
+	const packageJsonArray = glob.sync( globPattern, { cwd } );
```
psmyrek, line 107: The `cwd` here is probably not needed and could be anything, as the `globPattern` is an absolute path.

przemyslaw-zan:
@pomek [https://github.com/ckeditor/ckeditor5-dev/pull/737#discussion_r772935139](https://github.com/ckeditor/ckeditor5-dev/pull/737#discussion_r772935139) https://github.com/ckeditor/ckeditor5-dev/pull/737#discussion_r772927583

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
pomek: I guess there is no reason for displaying the diff when executing the script in normal mode (without the `--dry-run` option).

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,232 @@
...
+/**
+ * Updates `@ckeditor/ckeditor5-*` and `ckeditor5` dependencies in the specified directory to the latest version.
+ * Latest version is taken from the `version` property from the root of the `package.json` file, since that value
+ * should already have been bumped at this point of release process.
+ *
+ * @param {String} pathToUpdate Directory containing files to update.
+ * @param {Boolean} dryRun If set to true, diff of changes that would be made is calculated, and included in the returned object. Without
+ * this flag, file is updated normally.
+ * @returns {Object} Object containing numerical values counting files found and updated, as well as array of all changes made. If `dryRun`
+ * is set to true, `differences` array will be empty, as there is no point in calculating that which is only used during `dryRun`.
```
psmyrek, line 126: It would be nice to re-phrase it a bit to add also the property names that will hold the described values.

Also there is bug in this description: if the `dryRun` flag is `false` (off), but not `true`, as it is now, the `differences` array will be empty.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,232 @@
...
+ */
+function updateDirectory( pathToUpdate, dryRun ) {
+	const globPattern = pathToUpdate + '/*/package.json';
+	const packageJsonArray = glob.sync( globPattern );
+
+	if ( !packageJsonArray.length ) {
+		return { found: 0, updated: 0 };
```
psmyrek, line 133: Why empty `differences` array is not returned in this case?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,232 @@
...
+			}
+		}
+	}
+
+	return { found: packageJsonArray.length, updated: updatedFiles, differences };
+}
+
+/**
+ * This function takes an object and a property name, and modifies said property in that object. If any of the properties
+ * of that property matches the regex, its value will be updated to the value of `version` property from the root of the object.
```
psmyrek, line 167: Maybe something like:

_Updates the CKEditor 5 dependencies, except the `*-dev` and `*-inspector` ones, in the provided `package.json` file._

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,232 @@
...
+		}
+
+		parsedPkgJson[ propertyName ][ dependency ] = `^${ version }`;
+	}
+}
+
+/**
+ * Takes in raw changelog for a single file generated by jsdiff library, and returns formatted array of strings, containing
+ * human-readable, line by line changelog of a file, with removals and additions colored, as well as unnecessary chunks of
+ * unchanged file hidden.
```
psmyrek, line 189: This needs to be re-phrased, as well. The `jsdiff` is named as `diff` on NPM and it also should be wrapped in the backticks \`\`.

> unnecessary chunks of unchanged file hidden.

☝🏻 🤔

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,232 @@
...
+ *
+ * @param {Array<Object>} diff Array of changes for a single file generated by jsdiff library.
+ * @returns {Array<String>} Formatted changelog split into single lines.
+ */
+function formatDiff( diff ) {
+	const formattedDiff = [];
+	const regex = /(?<=":) (?=")/;
```
psmyrek, line 196: Why this regexp is needed?

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,232 @@
...
+		process.stdin.setRawMode( true );
+
+		process.stdin.on( 'keypress', ( str, key ) => {
+			if ( key.name === 'q' ) {
+				console.log( chalk.yellow( 'Manual exit.' ) );
+				process.exit();
+			}
+
+			const nextDiff = totalResult.differences.shift();
+			const formattedDiff = formatDiff( nextDiff.content );
```
psmyrek, line 74: Here it will crash if there are no differences.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,232 @@
...
+			].join( '' ) );
+
+			i++;
+		} else if ( !current.added && !current.removed && currentLines.length > 8 ) {
+			// Cutting out the middle of a long streak of unchanged lines.
+			const shortenedLines = [
+				...currentLines.slice( 0, 3 ),
+				chalk.yellow( `[...${ currentLines.length - 7 } lines without changes...]` ),
+				...currentLines.slice( -4 )
+			].join( '\n' );
```
psmyrek, line 221: Maybe it would be better to dim the unchanged lines? Yellow text draws attention.

psmyrek (changes requested):

packages/ckeditor5-dev-env/tests/index.js
przemyslaw-zan: 49cb082

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
przemyslaw-zan: I need to split these lines and merge them back together in order to pretty print the diff. Indentations are also kept in these lines, so there are more than one space. I suppose i could split using last space of a string instead, but this seems safer.

```json
    "@ckeditor/ckeditor5-core": "^31.1.0",
```

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,232 @@
...
+		totalResult.found += result.found;
+		totalResult.updated += result.updated;
+
+		if ( result.differences ) {
+			totalResult.differences.push( ...result.differences );
+		}
+
+		if ( !result.found ) {
+			console.log( 'No files were found.\n' );
+		} else if ( result.found && !result.updated ) {
```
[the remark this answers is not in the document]
przemyslaw-zan: How did i miss something so obious 😅

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
psmyrek: What if, for whatever reason, the script changes something else in the `package.json` file, other than the CKEditor 5 dependency versions? Now, only changes to the dependency versions are displayed in red and green colors, but other changes that could occur in the file are not highlighted in the console.

The `--dry-run` flag should display all the changes that the script has made, to be sure that running it without this flag will not surprise us.

If the dependency version has changed, it can be highlighted as it is now (only the version, not the full line). But when something else has changed that is not expected, let the entire changed line be highlighted.

psmyrek (changes requested):

packages/ckeditor5-dev-env/package.json
```diff
@@ -15,12 +15,14 @@
...
     "conventional-commits-parser": "^3.1.0",
     "del": "^5.1.0",
+    "diff": "^5.0.0",
     "fs-extra": "^9.0.0",
     "git-raw-commits": "^2.0.7",
     "glob": "^7.1.6",
     "inquirer": "^7.1.0",
     "minimatch": "^3.0.4",
     "mkdirp": "^1.0.4",
+    "mock-fs": "^5.1.2",
```
psmyrek, line 25: It should be in `devDependencies`.

packages/ckeditor5-dev-env/tests/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,304 @@
...
+ */
+
+'use strict';
+
+const { expect } = require( 'chai' );
+const fs = require( 'fs' );
+const mockFs = require( 'mock-fs' );
+const sinon = require( 'sinon' );
+
+describe( 'updatePackageVersions()', () => {
```
psmyrek, line 13: Thanks for adding some tests for this new util 👍🏻 

It would be great if you could increase the coverage, as it currently is less than 50% for this new util. [Here](https://coveralls.io/builds/45248775/source?filename=packages%2Fckeditor5-dev-env%2Flib%2Frelease-tools%2Ftasks%2Fupdate-package-versions.js) you can check the not covered lines (i.e. the `--dry-run` flag with diff formatting is not covered at all).

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * Updates the CKEditor 5 dependencies, except the `*-dev` and `*-inspector` ones, in the provided `package.json` file.
+ *
+ * @param {Object} parsedPkgJson Object to update.
+ * @param {String} propertyName Name of the property to update.
+ */
+function updateObjectProperty( parsedPkgJson, propertyName ) {
+	// Update only the CKEditor 5 dependencies, except the *-dev and *-inspector.
+	const regex = /^@ckeditor\/ckeditor5-(?!dev|inspector)|^ckeditor5$/;
```
pomek, line 177: I'm afraid this is not enough. What about `ckeditor5-react`, `ckeditor5-vue`, `ckeditor5-angular`? They will be updated even if used as dev-deps.

The easiest way to resolve the problem is to introduce a callback that will return a boolean. According to the callback result, a version of the package will be modified if returned `true`. 

Also, we can hardcore names of packages that always should be ignored, like mentioned integrations, inspector, and dev utils (`ckeditor5-dev-*`).

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
przemyslaw-zan: Okay, i've actually taken this into account, but then i forgot to color the rest of changes, so good thing You pointed that out 😅

Single removals and additions are simple enough, they are just colored lines. But how do we want to show replacements?

a) Replacement on the same line





b) Replacement on the same line, but trim the whitespaces off the addition





c) Replacement on the line below 





I'd suppose option b) is preferable, so this is what I'll do for now.

przemyslaw-zan:
I've encountered some problems while trying to write more tests as requested. The function is currently set up as a synchronous one. However, when `--dry-run` is used, it has asynchronous elements, as it waits for the user input. This did not cause problems untill i've tried to write some tests that would cover said option. This causes problem of multiple scripts awaiting anwser at the same time. I've tried converting it to asynchronous function by wraping it's content into `Promise` and using `return resolve()` instead of `process.exit()`, but this does not end the script, and it still awaits anwser, even if normally it shouldnt, as there are no changes. This makes me think that somehow, `return resolve()` does not kill `readline`. Moreover, I've failed to come up with a way to simulate keypreses to test it, and only found [dead-end unawsered question](https://stackoverflow.com/questions/64339263/is-it-possible-to-programmatically-simulate-a-keypress-using-the-readline-librar). What should we do with this? I've found [readline-sync](https://www.npmjs.com/package/readline-sync) package, and succesfylly replaced asynchronous `readline` _(with just small inconvinience of the script no longer taking `enter` or arrow keys, etc. as an input, only `a-z`, `0-9` and other characters)_. But even after that, testing is a problem. I can't simulate any keypress, if the synchronous function is still executing, waiting for input. But even in asynchronous form, how would we give input to the function, before `.then()` can be called? I suppose `setTimeout()` could be used, but that feels terribly _hacky_...

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,257 @@
...
+		if ( totalResult.updated ) {
+			console.log( '\n📍 ' + chalk.green( `Updated total of ${ totalResult.updated } files!\n` ) );
+		} else {
+			console.log( '\n📍 ' + chalk.green( 'No files needed an update.\n' ) );
+		}
+	}
+};
+
+/**
+ * Updates `@ckeditor/ckeditor5-*` and `ckeditor5` dependencies in the specified directory to the latest version.
```
psmyrek, line 120: This is not true, because there are some package names added to exceptions.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,257 @@
...
+function updateObjectProperty( parsedPkgJson, propertyName ) {
+	const version = parsedPkgJson.version;
+	const regex = /^@ckeditor\/ckeditor5-([a-z]+)|^ckeditor5$/;
+	const exceptions = [
+		'dev',
+		'inspector',
+		'react',
+		'vue',
+		'angular'
+	];
```
psmyrek, line 181: For me, it would be fine to just add the exceptions to the regexp:

```js
const regex = /^@ckeditor\/ckeditor5-(?!dev|inspector|react|vue|angular)|^ckeditor5$/;
```

Decide for yourself which is more readable.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,257 @@
...
+		} else if ( current.added ) {
+			// Other additions (trimming whitespaces in replacements)
+			formattedDiff.push( chalk.green( previous && previous.removed ? current.value.trim() : current.value ) );
+		} else if ( current.removed ) {
+			// Other removals
+			formattedDiff.push( chalk.red( current.value ) );
+		} else {
+			// Unchanged lines
+			formattedDiff.push( current.value );
+		}
```
psmyrek, line 242: There is quite a lot of conditions. Could it be simplified?

For example, probably the order of conditions could be better:

```
if ( validReplaceSequence ) {
    // ...
} else if ( current.added ) {
    // ...
} else if ( current.removed ) {
    // ...
} else if ( currentLines.length > 8 ) {
    // ...
} else {
    // ...
}
```

psmyrek:
> I've encountered some problems while trying to write more tests as requested. The function is currently set up as a synchronous one. However, when `--dry-run` is used, it has asynchronous elements, as it waits for the user input. This did not cause problems untill i've tried to write some tests that would cover said option. This causes problem of multiple scripts awaiting anwser at the same time. I've tried converting it to asynchronous function by wraping it's content into `Promise` and using `return resolve()` instead of `process.exit()`, but this does not end the script, and it still awaits anwser, even if normally it shouldnt, as there are no changes. This makes me think that somehow, `return resolve()` does not kill `readline`. Moreover, I've failed to come up with a way to simulate keypreses to test it, and only found [dead-end unawsered question](https://stackoverflow.com/questions/64339263/is-it-possible-to-programmatically-simulate-a-keypress-using-the-readline-librar). What should we do with this? I've found [readline-sync](https://www.npmjs.com/package/readline-sync) package, and succesfylly replaced asynchronous `readline` _(with just small inconvinience of the script no longer taking `enter` or arrow keys, etc. as an input, only `a-z`, `0-9` and other characters)_. But even after that, testing is a problem. I can't simulate any keypress, if the synchronous function is still executing, waiting for input. But even in asynchronous form, how would we give input to the function, before `.then()` can be called? I suppose `setTimeout()` could be used, but that feels terribly _hacky_...

I don't know if I understood what you mean exactly, but I think it would be sufficient to just mock the `readline.emitKeypressEvents()`, `process.exit()` and the `process.stdin` object with `on()` and `setRawMode()` functions. Then, your mocked `process.stdin.on()` function will be called that will register a callback for the `keypress` event, and then you can simulate a keystroke by trying to call a callback function registered to the `keypress` event, with appropriate arguments that are expected in the source file.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
przemyslaw-zan: IMO regexes are hard to read by their nature, and thus array with an explicit name is way better. Also, if for any reason we need to add more in the future, array is more expandable. Let's keep it the way it is.

pomek: I'd go with supporting strings and regexps:

```js
// Pattern that describes CKEditor 5 packages.
const CKEditor5Pattern = /^@ckeditor\/ckeditor5-(.*)|^ckeditor5$/; // [a-z]+ does not catch `ckeditor5-word-count`.

// Names of patterns to ignore.
const patternsToSkip = [
	'@ckeditor/ckeditor5-inspector',
	/@ckeditor/ckeditor5-dev-.*/, // All dev utils. Including the private too.
	'@ckeditor/jsdoc-plugins',
	/* ... */
];
```

pomek (changes requested):
Please, do not forget about https://github.com/ckeditor/ckeditor5-dev/pull/737/files#r777370255.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,281 @@
...
+		console.log( chalk.yellow( `      ${ chalk.bold( 'A' ) } - display diff from all files` ) );
+		console.log( chalk.yellow( `      ${ chalk.bold( 'Q' ) } - exit` ) );
+
+		if ( !totalResult.differences.length ) {
+			console.log( chalk.yellow( 'The script has not changed any files.' ) );
+			process.exit();
+		}
+
+		readline.emitKeypressEvents( process.stdin );
+		process.stdin.setRawMode( true );
```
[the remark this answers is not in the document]
psmyrek: Handling of the `CMD+C` or `CTRL+C` could be added as described in https://stackoverflow.com/questions/10021373/what-is-the-windows-equivalent-of-process-onsigint-in-node-js/14861513#14861513.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
przemyslaw-zan: I believe that exception for `'@ckeditor/jsdoc-plugins',` is not needed, as `/^@ckeditor\/ckeditor5-(.*)|^ckeditor5$/` will not catch it anyway

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,281 @@
...
+				pathsToCommit.push( path );
+			}
+		}
+	}
+
+	if ( dryRun ) {
+		console.log( `⚠️ ${ chalk.yellow( 'DRY RUN mode' ) } ⚠️` );
+		console.log( chalk.yellow( `${ chalk.bold( 'Any key' ) } - display next file diff` ) );
+		console.log( chalk.yellow( `      ${ chalk.bold( 'A' ) } - display diff from all files` ) );
+		console.log( chalk.yellow( `      ${ chalk.bold( 'Q' ) } - exit` ) );
```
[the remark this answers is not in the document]
przemyslaw-zan: Sure, implemented.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,281 @@
...
+			if ( pathToUpdate.commit ) {
+				totalResult.toCommit += result.updated;
+				pathsToCommit.push( path );
+			}
+		}
+	}
+
+	if ( dryRun ) {
+		console.log( `⚠️ ${ chalk.yellow( 'DRY RUN mode' ) } ⚠️` );
+		console.log( chalk.yellow( `${ chalk.bold( 'Any key' ) } - display next file diff` ) );
```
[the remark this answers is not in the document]
przemyslaw-zan: Sure, implemented.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
```diff
@@ -0,0 +1,281 @@
...
+	const formattedDiff = [];
+	const regex = /(?<=":) (?=")/;
+
+	for ( let i = 0; i < diff.length; i++ ) {
+		const previous = diff[ i - 1 ];
+		const current = diff[ i ];
+		const next = diff[ i + 1 ];
+		const currentLines = current.value.split( '\n' );
+
+		const validReplaceSequence = current.removed && next.added && regex.test( current.value ) && regex.test( next.value );
```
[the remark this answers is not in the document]
przemyslaw-zan: More or less. I've changed `Display` to `Format`, because the diff will be displayed either way, but if this check is passed, it will merge them into a single line for readability, or _pretty-print_ them if you will.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-ckeditor5-dependencies.js
```diff
@@ -0,0 +1,352 @@
...
+				const execOptions = {
+					stdio: 'inherit',
+					cwd: path
+				};
+
+				console.log( `${ chalk.green( '+' ) } '${ chalk.underline( path ) }'` );
+
+				execSync( `git add ${ path }/*/package.json`, execOptions );
+				execSync( 'git commit -m "Internal: Updated CKEditor 5 packages to the latest version. [skip ci]"', execOptions );
+			}
```
pomek, line 91: The current implementation creates more than a single update commit. `git commit` should be executed outside the loop.

packages/ckeditor5-dev-env/lib/release-tools/tasks/update-package-versions.js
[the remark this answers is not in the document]
pomek: => https://github.com/ckeditor/ckeditor5/issues/11084.

PR #738, opened 2021-12-15
Created the script for bumping a year in license headers.

przemyslaw-zan:
Feature (env): Created the script for bumping a year in license headers. Closes ckeditor/ckeditor5#10909.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
@@ -5,69 +5,107 @@
 
 'use strict';
 
+const chalk = require( 'chalk' );
 const fs = require( 'fs' );
 const glob = require( 'glob' );
 const minimatch = require( 'minimatch' );
 
 /**
- * See https://github.com/ckeditor/ckeditor5/issues/10909
+ * Updates year in all licenses in the provided directory, based on provided glob patterns.
+ *
+ * Replaces:
+ * 	Copyright (c) [initial year]-xxxx
+ *
+ * With:
+ *	Copyright (c) [initial year]-[current year]
  *
  * @param {Object} params
  * @param {String} params.cwd Current working directory from which all paths will be resolved
- * @param {Function} params.callback Optional function to modify the license header
  * @param {Number} params.initialYear Year from which the licenses should begin
  * @param {Array} params.globPatterns array of objects, where each object has string property 'pattern',
  * and optionally 'options' property.
  */
 module.exports = function bumpYear( params ) {
-	params.globPatterns
+	const filesToUpdate = params.globPatterns
 		.map( globPattern => glob.sync( globPattern.pattern, globPattern.options ) )
 		.reduce( ( previous, current ) => [ ...previous, ...current ] )
 		.filter( fileName => {
-			// Filter out nested `node_modules`.
-			if ( minimatch( fileName, '**/node_modules/**' ) ) {
-				return false;
-			}
-
-			// Filter out stuff from `src/lib/`.
-			if ( minimatch( fileName, '**/src/lib/**' ) ) {
-				return false;
-			}
-
-			// Filter out builds.
-			if ( minimatch( fileName, '**/ckeditor5-build-*/**' ) ||
-				minimatch( fileName, '**/build/**' ) ) {
-				return false;
-			}
-
 			// Filter out directories.
 			if ( fs.statSync( fileName ).isDirectory() ) {
 				return false;
 			}
 
-			return true;
-		} )
-		.forEach( fileName => {
-			const currentYear = new Date().getFullYear();
-			const yearOnlyRegexp = /(?<=^\/\*\*$[\s\S]+^ \* @license.+)(\d{4}-\d{4})(?=[\s\S]+^ \*\/$)/m;
-			const fullLicenseRegexp = /^\/\*\*$[\s\S]+^ \* @license.+\d{4}-\d{4}[\s\S]+^ \*\/$/m;
-
-			const fileContent = fs.readFileSync( fileName, 'utf-8' );
-
-			let updatedFileContent = fileContent.replace( yearOnlyRegexp, `${ params.initialYear }-${ currentYear }` );
-
-			if ( params.callback ) {
-				updatedFileContent = updatedFileContent.replace( fullLicenseRegexp, params.callback );
+			// Filter out nested `node_modules`.
+			if ( minimatch( fileName, '**/node_modules/**' ) ) {
+				return false;
 			}
 
-			if ( fileContent === updatedFileContent ) {
+			// Filter out builds.
+			if ( minimatch( fileName, '**/build/**' ) ) {
+				return false;
+			}
+
+			return true;
+		} );
+
+	const filesWithoutHeader = [];
+	const totalFiles = filesToUpdate.length;
+	let updatedFiles = 0;
+
+	filesToUpdate.forEach( fileName => {
+		const fileContent = fs.readFileSync( fileName, 'utf-8' );
+		const currentYear = new Date().getFullYear();
+
+		// This syntax has to be used in order to insert variable into the regex.
+		const yearRegexp = new RegExp( `(?<=Copyright \\(c\\) ${ params.initialYear }-)\\d{4}`, 'gm' );
+
+		let updatedFileContent = fileContent.replace( yearRegexp, currentYear );
+
+		// TODO: Remove this line after all repositories are updated.
+		updatedFileContent = updatedFileContent.replace( /- Frederico Knabben/, 'Holding sp. z o.o' );
+
+		// Optional replacing of second line with either longer or shorter version.
+		/* Short to long
+		updatedFileContent = updatedFileContent.replace(
+			/For licensing, see LICENSE\.md\./,
+			'For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license'
+		); */
+		/* Long to short
+		updatedFileContent = updatedFileContent.replace(
+			/For licensing, see LICENSE\.md or https:\/\/ckeditor\.com\/legal\/ckeditor-oss-license/,
+			'For licensing, see LICENSE.md.'
+		); */
+
+		if ( fileContent === updatedFileContent ) {
 			// License headers are only required in JS files.
 			// Also, the file might have already been updated.
-				if ( fileName.endsWith( '.js' ) && !fileContent.match( yearOnlyRegexp ) ) {
-					console.warn( `The file "${ params.cwd }/${ fileName }" misses a license header.` );
-				}
-			} else {
-				fs.writeFileSync( fileName, updatedFileContent, 'utf-8' );
+			if ( fileName.endsWith( '.js' ) && !fileContent.match( yearRegexp ) ) {
+				filesWithoutHeader.push( `${ params.cwd }/${ fileName }` );
 			}
-		} );
+		} else {
+			fs.writeFileSync( fileName, updatedFileContent, 'utf-8' );
+		}
+
+		updatedFiles++;
+
+		const percent = ( ( updatedFiles / totalFiles ) * 100 ).toFixed( 2 );
+		const trimmedFileName = fileName.length < 50 ? fileName : '...' + fileName.slice( -47 );
+
+		const output = [
+			`[ ${ updatedFiles } / ${ totalFiles } ]`,
+			chalk.green( `(${ percent }%)` ),
+			chalk.underline( trimmedFileName )
+		].join( ' ' );
+
+		process.stdout.clearLine();
+		process.stdout.cursorTo( 0 );
+		process.stdout.write( output );
+	} );
+
+	if ( filesWithoutHeader.length ) {
+		console.warn( chalk.red( 'Following files are missing their license headers:' ) );
+		for ( const file of filesWithoutHeader ) {
+			console.log( file );
+		}
+	}
 };
--- a/packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
@@ -39,7 +39,7 @@ module.exports = function bumpYear( params ) {
 				'**/node_modules/**'
 			];
 
-			options.nodir = options.nodir != null ? options.nodir : true;
+			options.nodir = true;
 			options.ignore = options.ignore ? [ ...options.ignore, ...defaultIgnore ] : defaultIgnore;
 
 			return glob.sync( globPattern.pattern, options );
--- a/packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
@@ -20,9 +20,9 @@ const glob = require( 'glob' );
  *
  * @param {Object} params
  * @param {String} params.cwd Current working directory from which all paths will be resolved.
- * @param {Number} [params.initialYear='2003'] Year from which the licenses should begin.
- * @param {Array} params.globPatterns array of objects, where each object has string property 'pattern',
- * and optionally 'options' property.
+ * @param {Array} params.globPatterns An array of objects, where each object has string property `pattern`,
+ * and optionally `options` property for this `glob` pattern.
+ * @param {String} [params.initialYear='2003'] Year from which the licenses should begin.
  */
 module.exports = function bumpYear( params ) {
 	if ( !params.initialYear ) {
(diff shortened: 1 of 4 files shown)


przemyslaw-zan:
I've already spotted a problem, I wrote this script to catch licenses in specifically `.js` files, but there are other license formats:

[https://github.com/ckeditor/ckeditor5/blob/master/scripts/switch-to-dev-dev.sh#L3-L4](https://github.com/ckeditor/ckeditor5/blob/master/scripts/switch-to-dev-dev.sh#L3-L4)

Previous script did not had this problem, since it was way less specific. New script has to be more specific, since now initial year is no longer constant, which causes it to also catch lines such as:

[https://github.com/ckeditor/ckeditor5/blob/master/LICENSE.md?plain=1#L17](https://github.com/ckeditor/ckeditor5/blob/master/LICENSE.md?plain=1#L17)

Because of that and the fact that we want option to pass the callback to modify the license, new script has to make sure that entire license is caught, and only the license, and thus the regex has to be stricter (which allowed me to catch some missformated licenses, which i already corrected in [CKEditor5 PR](https://github.com/ckeditor/ckeditor5/pull/11012)).

In my opinion, fix is simple: use different regexes depending on the format of the file.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,73 @@
...
+			const currentYear = new Date().getFullYear();
+			const yearOnlyRegexp = /(?<=^\/\*\*$[\s\S]+^ \* @license.+)(\d{4}-\d{4})(?=[\s\S]+^ \*\/$)/m;
+			const fullLicenseRegexp = /^\/\*\*$[\s\S]+^ \* @license.+\d{4}-\d{4}[\s\S]+^ \*\/$/m;
+
+			const fileContent = fs.readFileSync( fileName, 'utf-8' );
+
+			let updatedFileContent = fileContent.replace( yearOnlyRegexp, `${ params.initialYear }-${ currentYear }` );
+
+			if ( params.callback ) {
+				updatedFileContent = updatedFileContent.replace( fullLicenseRegexp, params.callback );
```
psmyrek, line 60: I think that we do not need such callback, that is executed for each matched param. It would be sufficient to just call the callback with the current license string and whatever is returned from the callback, it is used as the new license string. 

I think also that the default license should be placed somewhere as the default value, to not define the callback in all the places/repos. Moreover, the callback should be used only if the default license is for any reasons not applicable.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,73 @@
...
+				return false;
+			}
+
+			// Filter out directories.
+			if ( fs.statSync( fileName ).isDirectory() ) {
+				return false;
+			}
+
+			return true;
+		} )
```
psmyrek, line 49: These rules should not be part of the main script. This script should only take the glob patterns and find files that matches them.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+'use strict';
+
+const fs = require( 'fs' );
+const glob = require( 'glob' );
+const minimatch = require( 'minimatch' );
+
+/**
+ * See https://github.com/ckeditor/ckeditor5/issues/10909
```
psmyrek, line 13: More descriptive sentences would be better instead of the link. I think the link to the issue is not needed at all.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,73 @@
...
+			if ( fs.statSync( fileName ).isDirectory() ) {
+				return false;
+			}
+
+			return true;
+		} )
+		.forEach( fileName => {
+			const currentYear = new Date().getFullYear();
+			const yearOnlyRegexp = /(?<=^\/\*\*$[\s\S]+^ \* @license.+)(\d{4}-\d{4})(?=[\s\S]+^ \*\/$)/m;
+			const fullLicenseRegexp = /^\/\*\*$[\s\S]+^ \* @license.+\d{4}-\d{4}[\s\S]+^ \*\/$/m;
```
psmyrek, line 53: Would it be possible to use only one regular expression to match the whole license string and then the year would be just a matched group inside it? Having 2 regular expressions seems like unnecessary duplication. Also it looks like the regular expression pattern is too generic, i.e. it could include the `Copyright (c)` pattern as well (which is AFAIK present in all the files), so there would be less false matches returned.

psmyrek (changes requested):

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
psmyrek: The `glob` package has the [`ignore` option](https://github.com/isaacs/node-glob#options) - maybe it could be used to do the filtering on the low-level?

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
przemyslaw-zan: The problem i had was, how to replace only the year, while also matching the entire regexp? Solution i found, was to use positive lookahead and lookbehind in order to find whole license, but only replace the part i want. So we look for the entire license, but only match the year, which we can then easily replace: `(?<=...)(2003-2021)(?=...)`. If you know of a better way to match whole license, split it in to 3 groups _(prefix / year / suffix),_ use one `.replace` function to replace only the year, and then use another `.replace` to modify the entire thing, then we can implement it. This solution produces two regexes, but results in much cleaner `.replace` functions. (Other solutions i was tinkering with ended up with much messier `.replace`, and were hard to make them work.)

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
przemyslaw-zan: I agree that we could remove `src/lib/` from there. However:

*   filtering out directories is obvious
*   it would probably make sense to filter out `node_modules` everywhere
*   while not all projects might have `build` directory, we probably want to filter out all that do anyway

The last point is debatable, however, at least the first two surely should be kept in there?

psmyrek: After rethinking it, you are right. The `**/node_modules/**` and `**/build/**` directories should be always skipped. Also filtering out the directories here makes sense, as we want to target only files. But the `**/src/lib/**` and `**/ckeditor5-build-*/**` patterns are strictly related to the CKEditor 5 repository so they should not be here.

My idea was to speed up this script a bit, so that it doesn't have to go to the deeply nested `node_modules` and other directories (i.e. `**/build/**`), which we'll filter out anyway. To accomplish this, it looks like it is enough to provide the `ignore` option (an array of patterns) to the `glob` that should not be traversed. As soon as `glob` finds a path that matches the ignored pattern, it does not go deeper. Would you like to investigate a bit to check if this is doable? If yes, then 🤞🏻. If not, then please just extract the two `**/src/lib/**` and `**/ckeditor5-build-*/**` patterns out from this file, as they are related to the CKEditor 5 only.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
psmyrek: If the answer for [this question](https://github.com/ckeditor/ckeditor5/issues/10909#issuecomment-996748334) is that we should unify the license globally, then there is even a simpler solution: define the default license on top of this file to something like:

```js
const DEFAULT_LICENSE = `
/**
 * @license Copyright (c) %START_YEAR%-%END_YEAR%, CKSource Holding sp. z o.o. All rights reserved.
 * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
 */
`;
```

Then, it is sufficient to just replace the `%START_YEAR%` and `%END_YEAR%` placeholders with appropriate values.
IMHO there is no need to use regular expressions to search for the years in the license content.

przemyslaw-zan: This does not account for different file formats that have different syntax for comments, since we need to replace the license year everywhere.

[https://github.com/ckeditor/ckeditor5/blob/master/LICENSE.md?plain=1#L5](https://github.com/ckeditor/ckeditor5/blob/master/LICENSE.md?plain=1#L5)

This line does not even have `@license` prefix, so we either

1.  replace only `Copyright (c) 2003-2021`
2.  write a whole lot of different cases to cover all of the possibilities

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
przemyslaw-zan:
```
pattern: '!(build|coverage|node_modules|external)/**',
options: {
    ignore: [
        '**/ckeditor5-build-*/**',
        '**/src/lib/**'
    ]
}
```

☝️ this works. Funny thing is, that current version on master has this line:

[https://github.com/ckeditor/ckeditor5/blob/84b676ee99137617418498b54494852ffbfcbba9/scripts/bump-year.js#L49](https://github.com/ckeditor/ckeditor5/blob/84b676ee99137617418498b54494852ffbfcbba9/scripts/bump-year.js#L49)

But it does not work, and it updates `**/ckeditor5-build-*/**` anyway

psmyrek: In my previous comment my point was to get rid of the `.filter()` method at all and provide the common ignore patterns as the default option to the `glob`. Then, `glob` will not need to go inside the deeply nested directories that we are skipping anyway. The common options for glob could be something like:

```
{
  nodir: true,
  ignore: [
    '**/node_modules/**'
  ]
}
```

These options should be then merged with the options received as an arguments to this script.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,115 @@
...
+			return true;
+		} );
+
+	const filesWithoutHeader = [];
+	const totalFiles = filesToUpdate.length;
+	let updatedFiles = 0;
+
+	filesToUpdate.forEach( fileName => {
+		const fileContent = fs.readFileSync( fileName, 'utf-8' );
+		const currentYear = new Date().getFullYear();
```
psmyrek, line 61: It is not needed to calculate current year for each changed file.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,115 @@
...
+
+		// This syntax has to be used in order to insert variable into the regex.
+		const yearRegexp = new RegExp( `(?<=Copyright \\(c\\) ${ params.initialYear }-)\\d{4}`, 'gm' );
```
psmyrek, line 64: The `m` modifier is not needed as it only affects the behavior of `^` and `$`, which are not used here.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,115 @@
...
+
+		let updatedFileContent = fileContent.replace( yearRegexp, currentYear );
+
+		// TODO: Remove this line after all repositories are updated.
+		updatedFileContent = updatedFileContent.replace( /- Frederico Knabben/, 'Holding sp. z o.o' );
```
psmyrek, line 69: This does not replace all old sentences.

Because of that, my proposal is to focus on the mandatory scope (changing the years) and do not change the license.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,115 @@
...
+		const percent = ( ( updatedFiles / totalFiles ) * 100 ).toFixed( 2 );
+		const trimmedFileName = fileName.length < 50 ? fileName : '...' + fileName.slice( -47 );
+
+		const output = [
+			`[ ${ updatedFiles } / ${ totalFiles } ]`,
+			chalk.green( `(${ percent }%)` ),
+			chalk.underline( trimmedFileName )
+		].join( ' ' );
+
+		process.stdout.clearLine();
```
psmyrek, line 104: To prevent blinking on Windows:

```suggestion
		process.stdout.clearLine( 1 );
```

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,115 @@
...
+		process.stdout.cursorTo( 0 );
+		process.stdout.write( output );
+	} );
+
+	if ( filesWithoutHeader.length ) {
+		console.warn( chalk.red( 'Following files are missing their license headers:' ) );
+		for ( const file of filesWithoutHeader ) {
```
psmyrek, line 111:
```suggestion
		console.warn( chalk.red( 'Following files are missing their license headers:' ) );

		for ( const file of filesWithoutHeader ) {
```

psmyrek (changes requested):

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,95 @@
...
+ * 	Copyright (c) [initial year]-xxxx
+ *
+ * With:
+ *	Copyright (c) [initial year]-[current year]
+ *
+ * @param {Object} params
+ * @param {String} params.cwd Current working directory from which all paths will be resolved.
+ * @param {Number} [params.initialYear='2003'] Year from which the licenses should begin.
+ * @param {Array} params.globPatterns array of objects, where each object has string property 'pattern',
+ * and optionally 'options' property.
```
psmyrek, line 25: I know the `params` argument is an object so order of properties does not matter, but to conform to the convention it is better to define optional arguments last. Also there is a typo in the `array` (should be in uppercase), and the type of the `initialYear` should be a string, according to its default value.

```suggestion
 * @param {Array} params.globPatterns An array of objects, where each object has string property `pattern`,
 * and optionally `options` property for this `glob` pattern.
 * @param {String} [params.initialYear='2003'] Year from which the licenses should begin.
```

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
```diff
@@ -0,0 +1,95 @@
...
+
+	const filesToUpdate = params.globPatterns
+		.map( globPattern => {
+			const options = globPattern.options || {};
+
+			const defaultIgnore = [
+				'**/node_modules/**'
+			];
+
+			options.nodir = options.nodir || true;
```
psmyrek, line 42: What is the point of `options.nodir = options.nodir || true`?

The `options.nodir` will be always `true` in this ☝🏻 instruction, even if received `globPattern.options.nodir` is `false`.

psmyrek (commented):
Minor, mostly cosmetic improvements. Other than that, it works fine.

packages/ckeditor5-dev-env/lib/release-tools/tasks/bump-year.js
przemyslaw-zan: Right, my bad. I wanted to have default `true` value, but still prioritise whatever value would be passed from the options.

I've replaced that with `options.nodir = options.nodir == null ? options.nodir : true`, this will only use default value for `undefined` and `null`, but not for `false` as `false != null` returns `true`,

psmyrek: But still, why do you want to allow directories to be targeted, since a few lines below you expect that all the returned results from `glob` are files? You are trying to open and read the content from all the found paths, so when the given path points to a directory, the script will crash.

I think that the `nodir` should have a hardcoded value equal `true`, because the script is able to work only on files.

psmyrek:
The old `scripts/bump-year.js` should now call the new script from `ckeditor5-dev-env`.

scripts/bump-year.js
```diff
@@ -17,54 +17,18 @@ git commit -am "Internal: Bumped the year." && git push
...
 				}
-			} else {
-				fs.writeFile( fileName, updatedData, err => {
-					if ( err ) {
-						throw err;
-					}
-				} );
+			},
+			{
+				pattern: '!(node_modules|coverage|.nyc_output)/**'
```
psmyrek, line 31: The `node_modules` directory is already ignored globally.

psmyrek (changes requested):

scripts/bump-year.js
przemyslaw-zan, line 31:
```suggestion
				pattern: '!(coverage|.nyc_output)/**'
```

psmyrek (approved):
LGTM 🎉

PR #751, opened 2022-02-08
Added handling for the "©" character when updating licenses

przemyslaw-zan:
Other (env): Added handling for the `©` character when updating licenses. Closes ckeditor/ckeditor5#11119.

### DIFF
--- a/packages/ckeditor5-dev-env/lib/tasks/bump-year.js
+++ b/packages/ckeditor5-dev-env/lib/tasks/bump-year.js
@@ -55,7 +55,7 @@ module.exports = function bumpYear( params ) {
 
 	filesToUpdate.forEach( fileName => {
 		const fileContent = fs.readFileSync( fileName, 'utf-8' );
-		const yearRegexp = new RegExp( `(?<=Copyright \\(c\\) ${ params.initialYear }-)\\d{4}`, 'g' );
+		const yearRegexp = new RegExp( `(?<=Copyright (?:\\(c\\)|\u00A9) ${ params.initialYear }-)\\d{4}`, 'g' );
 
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-dev-env/lib/tasks/bump-year.js
```diff
@@ -55,7 +55,7 @@ module.exports = function bumpYear( params ) {
 
 	filesToUpdate.forEach( fileName => {
 		const fileContent = fs.readFileSync( fileName, 'utf-8' );
-		const yearRegexp = new RegExp( `(?<=Copyright \\(c\\) ${ params.initialYear }-)\\d{4}`, 'g' );
+		const yearRegexp = new RegExp( `(?<=Copyright (?:\\(c\\)|\u00A9) ${ params.initialYear }-)\\d{4}`, 'g' );
```
psmyrek, line 58: What do you think about specifying the © character explicitly?

```suggestion
		const yearRegexp = new RegExp( `(?<=Copyright (?:\\(c\\)|©) ${ params.initialYear }-)\\d{4}`, 'g' );
```

psmyrek: By specifying the © character explicitly you don't need to switch to the console or Internet source to see what it is (the `\u00A9` tells me nothing) and this is safe. In fact, we already do it in CKEditor 5:

https://github.com/ckeditor/ckeditor5/blob/82efb635e3f5ac964d1e17097cf655455948ea0a/packages/ckeditor5-special-characters/src/specialcharacterstext.js#L61

pomek (approved):
LGTM.

PR #756, opened 2022-03-02
Addressed issues with karma after upgrading to webpack 5

pomek:

Fix (tests): Karma integrated with the IntelliJ IDE rebuilds the test bundle when detecting a change.

Fix (tests): Paths to files are clickable from a console if a test fails.

Fix (tests): Karma should not execute the test bundle twice when detecting a change (a macOS case).


psmyrek (approved):
LGTM 🎉

PR #764, opened 2022-03-22
Bumped `postcss` to v8

przemyslaw-zan:

Feature (utils): The `styles.getPostCssConfig()` function returns a configuration compatible with `postcss@8`. See ckeditor/ckeditor5#11460.

MAJOR BREAKING CHANGE (utils): Upgraded the `postcss` version to 8 and marked the package as peer dependency instead of dependency. Starting from this version, when using the `@ckeditor/ckeditor5-dev-utils`, you need to manually install the `postcss` dependency.

Tests (env): Fixed broken tests on Windows.

Tests (env): Fixed deprecation warnings.


packages/ckeditor5-dev-utils/lib/styles/themeimporter.js
```diff
@@ -41,30 +38,35 @@ const getPackageName = require( './utils/getpackagename' );
...
-} );
-
-// Imports a complementary theme file corresponding with a CSS file being processed by
-// PostCSS, if such theme file exists.
-//
-// @private
-// @param {Options} Plugin options.
-// @returns {Promise}
+};
+
```
pomek, line 61:
```suggestion
};

module.exports.postcss = true;

```

packages/ckeditor5-dev-utils/package.json
```diff
@@ -18,11 +18,11 @@
     "is-interactive": "^1.0.0",
     "javascript-stringify": "^1.6.0",
     "pofile": "^1.0.9",
-    "postcss": "^7.0.39",
-    "postcss-import": "^12.0.0",
-    "postcss-loader": "^3.0.0",
-    "postcss-mixins": "^6.2.0",
-    "postcss-nesting": "^7.0.0",
+    "postcss": "^8.4.12",
```
pomek, line 21:
```suggestion
```

`ckeditor5-dev-utils` does not use the package anywhere.

packages/ckeditor5-dev-utils/tests/stream.js
```diff
@@ -97,7 +97,7 @@ describe( 'stream', () => {
 				const file = new Vinyl( {
 					cwd: './',
 					path,
-					contents: new Buffer( '' )
+					contents: new Buffer.from( '' ) // eslint-disable-line new-cap
```
pomek, line 100:
```suggestion
					contents: Buffer.from( '' )
```

It's enough.

pomek (approved):
LGTM.

PR #769, opened 2022-05-26
Display toast notifications while the manual tests are being compiled.

oleq:

Feature (tests): Display toast messages while manual tests are being compiled. Closes ckeditor/ckeditor5#11831.

---

### Additional information

~This is WIP (missing tests).~ See to learn more ckeditor/ckeditor5#11831.

~Note: In this version, the notification wont display for changes to HTML files. JS only.~

---


pomek (commented):
I played with the proposed changes, and I'm under a huge impression.

I wondered whether we could detect which test has changed to display a notification only if we run the same test in a browser, but it might be difficult as the watcher does not inform which file has changed. We can see an update in test files, but we won't detect a change in source files. Hence, let's stay with the proposed version.

oleq:
I wrote some tests and fixed existing ones to make it work. But honestly, I feel a little bit lost because the sheer number of mocks there is overwhelming and I wasn't able to write tests for some (most?) of the new functionality. Maybe someone else could take it over and improve the coverage? Or maybe this is enough?

oleq:
BTW: Are we tracking the issue with manual tests reloading twice on every save @pomek ? The indicator makes this even more visible and we could cut the reload time by half anyway.

PR #780, opened 2022-07-22
Support for TypeScript when collecting translation entries

psmyrek:

Internal: Support for TypeScript when collecting translation entries. Closes ckeditor/ckeditor5#12029.

---

### Additional information

It looks like `acorn`, even in the newest version, doesn't understand TypeScript sources.

TypeScript support is provided in another package - `@babel/parser`, and the `@babel` ecosystem has also a separate package for walking in the generated tree - `@babel/traverse`.

Initially, I thought it would be possible to prepare an `acorn`-compatible AST from the `@babel/parser` to re-use the existing `acorn-walk` package that currently finds all `t()` calls. According to the documentation from `@babel/parser` it should be possible, because it is shipped with the `estree` plugin, which produces the AST that is `acorn`-compatible. Unfortunately, it didn't quite work, because some deeply nested `t()` calls were not detected, so the script reported multiple unused contexts for texts that in fact are used in our sources.

Luckily, the default AST format produced by the `@babel/parser` is very close to the `acorn` format (see [the list of all differences](https://babeljs.io/docs/en/babel-parser#output)). Also the `@babel/traverse` interface is almost identical to `acorn-walk`, so replacing both was fairly easy.

There is also a related PR - https://github.com/ckeditor/ckeditor5/pull/12128 - that restores the `ckeditor5-utils` package in the `translations:collect` script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,9 @@
     "release:bump-version": "node ./scripts/bump-versions.js",
     "release:publish": "node ./scripts/publish.js",
     "lint": "eslint --quiet \"**/*.js\"",
-    "precommit": "lint-staged"
+    "precommit": "lint-staged",
+    "clean": "npx rimraf package-lock.json yarn.lock ./**/node_modules",
+    "reinstall": "yarn run clean && yarn install"
   },
   "lint-staged": {
     "**/*.js": [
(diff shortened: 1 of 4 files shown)


pomek (approved):
LGTM.

PR #781, opened 2022-08-31
Prevent publishing an empty package on npm

psmyrek:

Feature (env): Added a new check during the release process that verifies if all required files exist in a package directory before releasing a package on npm.

---

### Additional information

See the related change in `ckeditor5` repo: https://github.com/ckeditor/ckeditor5/pull/12392.


psmyrek:
During the live review with @pomek, we agreed to the following changes:
- Currently, the verification if the `files` property matches any file is terminated upon the first unmatched entry. We need to check all entries in the `files` property and remember the number of matches for each entry.
- In "dry run" mode, we should always show the verification results in a table (two columns: _Entry name_ and _Number of matches_).
- In non-"dry run" mode, we should show the verification results just before asking if the missing file(s) are OK.
- If the answer to the question of whether to continue the release process is negative, then the release process is terminated and the default message is displayed. It would be good to include some hints (what to do in this case to start the release process again).

pomek (approved):
LGTM.

PR #783, opened 2022-09-06
Split multi-scoped commit into multiple single-scoped commits

psmyrek:

Feature (env): Split multi-scoped commit into multiple single-scoped commits, so the changelog generator can produce entries for all touched scopes. Closes ckeditor/ckeditor5#10605.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitfactory.js
+++ b/packages/ckeditor5-dev-env/lib/release-tools/utils/transformcommitfactory.js
@@ -5,6 +5,7 @@
 
 'use strict';
 
+const { cloneDeepWith } = require( 'lodash' );
 const utils = require( './transformcommitutils' );
 const getChangedFilesForCommit = require( './getchangedfilesforcommit' );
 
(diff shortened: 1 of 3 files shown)


pomek (changes requested):
I believe we need a few more tests.

* A message containing many entries:

	```
    Feature (foo): Lorem ipsum dolor sit amet.
    
    Internal (bar): Aliquam elementum.
    
    Fix (foo, bar): Suspendisse ac egestas ipsum.
	```
    
* As above but for a [merge commit](https://github.com/ckeditor/ckeditor5-dev/blob/9dc15a54220fe3837ac475cbf6da24e58bbc9dca/packages/ckeditor5-dev-env/tests/release-tools/utils/transformcommitfactory.js#L404-L427).

psmyrek:
The first case (_A message containing many entries_) is already handled in `'multi-entries commit'` -> `'handles scoped and non-scoped changes'` test.

The test for merge commit has been just added, thanks.

pomek (approved):
LGTM.

PR #791, opened 2022-09-28
Standardized the DLL build script

przemyslaw-zan:
Internal (tests): Print a log once the DLL building process is finished.


packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
```diff
@@ -144,6 +144,8 @@ module.exports = function runManualTests( options ) {
 				await buildDllInRepository( path.dirname( pkgJsonPath ) );
 			}
 		}
+
+		console.log( '\n📍 DLL building complete!\n' );
```
pomek, line 148:
```suggestion
		console.log( '\n📍 DLL building complete.\n' );
```

There is no need to shout :D

pomek (approved):
LGTM.

PR #792, opened 2022-10-04
Introduced new `tag-event` plugin for TypeDoc

psmyrek:

Feature (typedoc-plugins): Created the `tag-event` plugin that collects all occurrences of the `@eventName` tags in the CKEditor 5 code base. Closes ckeditor/ckeditor5#12526.

---

### Additional information

This plugin does not parse the `@event` tag, because it has a special meaning in the TypeDoc and it would be difficult to get it to work as we expect. This is why I added support for the custom `@eventName` tag, which now replaces the old `@event` tag.


pomek (approved):
LGTM.

PR #797, opened 2022-10-05
Fix (tests): Disable watch mode for building manual when no files provided

martnpaneq:

Fix (tests): Disabled watch mode for building manual tests when no files are provided. Closes ckeditor/ckeditor5#12189.

---

### Additional information

* Moved default values for files from `bin/test-manual.js` to `lib/tasks/runmanualtests.js`.
* Added deduplication for file to test glob transform.
* Removed duplicated logic in tests

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
+++ b/packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
@@ -42,9 +42,10 @@ const transformFileOptionToTestGlob = require( '../utils/transformfileoptiontote
 module.exports = function runManualTests( options ) {
 	const log = logger();
 	const buildDir = path.join( process.cwd(), 'build', '.manual-tests' );
-	const files = ( options.files && options.files.length ) ? options.files : [ '*' ];
-	const sourceFiles = files
-		.flatMap( file => transformFileOptionToTestGlob( file, true ) )
+	const isFilesFlagProvided = ( options.files && options.files.length );
+	const files = isFilesFlagProvided ? options.files : [ '*', 'ckeditor5' ];
+	const dedupedFileTestGlobs = [ ...new Set( files.flatMap( file => transformFileOptionToTestGlob( file, true ) ) ) ];
+	const sourceFiles = dedupedFileTestGlobs
 		.reduce( ( result, manualTestPattern ) => {
 			return [
 				...result,
@@ -60,7 +61,7 @@ module.exports = function runManualTests( options ) {
 	const language = options.language;
 	const additionalLanguages = options.additionalLanguages;
 	const silent = options.silent || false;
-	const disableWatch = options.disableWatch || false;
+	const disableWatch = options.disableWatch || !isFilesFlagProvided;
 	let socketServer;
 
 	function onTestCompilationStatus( status ) {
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
```diff
@@ -42,9 +42,10 @@ const transformFileOptionToTestGlob = require( '../utils/transformfileoptiontote
 module.exports = function runManualTests( options ) {
 	const log = logger();
 	const buildDir = path.join( process.cwd(), 'build', '.manual-tests' );
-	const files = ( options.files && options.files.length ) ? options.files : [ '*' ];
-	const sourceFiles = files
-		.flatMap( file => transformFileOptionToTestGlob( file, true ) )
+	const isFilesFlagProvided = ( options.files && options.files.length );
+	const files = isFilesFlagProvided ? options.files : [ '*', 'ckeditor5' ];
+	const dedupedFileTestGlobs = [ ...new Set( files.flatMap( file => transformFileOptionToTestGlob( file, true ) ) ) ];
```
psmyrek, line 47: Good catch with that 👍🏻 Previously, calling `yarn manual -f table,utils,table` would produce duplicated entries. Now it will be fixed.

Maybe it would be good to add a small test for that?

packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
```diff
@@ -60,7 +61,7 @@ module.exports = function runManualTests( options ) {
 	const language = options.language;
 	const additionalLanguages = options.additionalLanguages;
 	const silent = options.silent || false;
-	const disableWatch = options.disableWatch || false;
+	const disableWatch = options.disableWatch || !isFilesFlagProvided;
```
psmyrek, line 64: One thing to consider: now it will not be possible to explicitly enable the watch mode if no files are provided.
I'm not sure if this is important as the tests probably will throw an error anyway in that case...

packages/ckeditor5-dev-tests/tests/tasks/runmanualtests.js
```diff
@@ -630,6 +608,53 @@ describe( 'runManualTests', () => {
...
+	} );
+
+	it( 'should set default values for files', () => {
+		return runManualTests( defaultOptions )
+			.then( () => {
+				expect( spies.transformFileOptionToTestGlob.calledTwice ).to.equal( true );
+				expect( spies.transformFileOptionToTestGlob.firstCall.args[ 0 ] ).to.equal( '*' );
+				expect( spies.transformFileOptionToTestGlob.secondCall.args[ 0 ] ).to.equal( 'ckeditor5' );
+			} );
+	} );
```
psmyrek, line 656: Please add additional test if non-default files are provided to check the arguments passed to the `transformFileOptionToTestGlob()`.

psmyrek (changes requested):

packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
martnpaneq: Added a test for this case.

packages/ckeditor5-dev-tests/lib/tasks/runmanualtests.js
martnpaneq: Added log that shows up at the end of the script when file watching is disabled.

psmyrek (approved):
LGTM 🎉

pomek (approved):
LGTM.

PR #798, opened 2022-10-07
Introduced new `tag-observable` plugin for TypeDoc

psmyrek:

Feature (typedoc-plugins): Created the `tag-observable` plugin that parses all occurrences of the `@observable` tag associated with class properties in the CKEditor 5 code base. Closes ckeditor/ckeditor5#12527.

---

### Additional information

To detect if the property is observable from Umberto point of view, a new helper could be introduced there:
```js
function isObservable( item ) {
    // Check if the `@observable` tag is set.
    if ( item.comment && item.comment.tags ) {
        return item.comment.tags.some( tag => tag.tag === 'observable' );
    }

    return false;
};
```

Previously, I tried to add a new flag `isObservable` to TypeDoc, but it seems that it is not so easy. Flags in TypeDoc are stored as appropriate bit offsets on which some bit operations are performed, so adding a new flag from the outside may break everything.


pomek:
The `@observable` tag also lands in a `blockTags` section.

```
{
	"blockTags": [
		{
			"tag": "@observable",
			"content": []
		}
	]
}
```

pomek (approved):
LGTM.

PR #799, opened 2022-10-12
Introduced new `event-param-fixer` plugin for TypeDoc

psmyrek:

Feature (typedoc-plugins): Introduced new `event-param-fixer` plugin for TypeDoc. It creates the `eventInfo` parameter that is of type `EventInfo` class, and then inserts it as the first parameter for each found event. Closes ckeditor/ckeditor5#12606.


pomek (approved):
LGTM.

PR #803, opened 2022-10-19
Internal (dev-utils): Add CKEditorWebpackPlugin only when lang dir detected

martnpaneq:
Internal (dev-utils): Add CKEditorWebpackPlugin only when lang dir detected. Closes #12584.


psmyrek (approved):
LGTM 🎉

PR #818, opened 2022-11-24
Improved detection of module declaration and JSDoc link in Typedoc plugins

psmyrek:

Internal: Improved detection of module declaration and JSDoc link in Typedoc plugins. Instead of using numeric values, which are not constant, we now use enums exported from TypeScript. Closes ckeditor/ckeditor5#12877.


pomek (approved):
LGTM.

PR #819, opened 2022-11-24
Added option to run tests and manual with external packages names as …

martnpaneq:
Other (tests): Added support for executing automated and manual tests from a root directory of an external repository. Closes ckeditor/ckeditor5#12757.

---

### Additional information

*Running `yarn test -f collaboration-features` and `yarn manual -f collaboration-features` should run only in roots of specified external repositories.*

*Specifying multiple external repositories is also possible `yarn run test -f collaboration-features -f ckeditor5-internal`*

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/utils/transformfileoptiontotestglob.js
+++ b/packages/ckeditor5-dev-tests/lib/utils/transformfileoptiontotestglob.js
@@ -26,7 +26,7 @@ const EXTERNAL_DIR_PATH = path.join( process.cwd(), 'external' );
  */
 module.exports = function transformFileOptionToTestGlob( pattern, isManualTest = false ) {
 	if ( doesPatternMatchExternalRepositoryName( pattern ) ) {
-		return getExternalReposityGlob( pattern, { isManualTest } );
+		return getExternalRepositoryGlob( pattern, { isManualTest } );
 	}
 
 	// Directory to look: `packages/ckeditor5-*/...`.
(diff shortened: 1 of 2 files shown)


psmyrek (commented):
It looks like it does not work as requested.

Let's assume we have the following directory structure:

```
ckeditor5/
├── external/
│   └── ckeditor5-example/
│       ├── packages/
│       │   ├── ckeditor5-foo/
│       │   │   └── tests/
│       │   │       └── manual/
│       │   └── ckeditor5-bar/
│       │       └── tests/
│       │           └── manual/
│       └── tests/
│           └── manual/
├── packages/
│   ├── ckeditor5-adapter-ckfinder/
│   │   └── tests/
│   │       └── manual/
│   └── ckeditor5-alignment/
│       └── tests/
│           └── manual/
└── tests/
    └── manual/
```

After calling `yarn manual -f ckeditor5-example` I would expect that only manual tests from `ckeditor5/external/ckeditor5-example/tests/manual/**` are built, but currently it takes all manual tests from `ckeditor5/external/ckeditor5-example/packages/*/tests/manual/**`.

We have a common util that handles the `-f` option for manual and automated tests in `ckeditor5-dev-tests` package (path: `lib/utils/transformfileoptiontotestglob.js`). I think we should change it a bit (instead of creating a new util) to align to the requested changes in parsing the `-f` option.

Proposal:
1. In the `transformFileOptionToTestGlob` util first collect all available repositories: the root `ckeditor5` and all external ones.
2. Then check if the provided file pattern matches any of them.
   - If yes, do not push the `/packages/*/` fragment to the produced path.
   - If not, nothing changes and current flow is valid.

psmyrek (approved):
LGTM 🎉 

I've corrected a bit the logic to always produce POSIX path globs, as it is required by the `glob` and `minimatch` utils.

psmyrek:
Also I've changed the commit message from:
```
Internal (dev): Added option to run tests and manual with external packages names as -f argument. Closes ckeditor/ckeditor5#12757
```
to
```
Other (tests): Added support for executing automated and manual tests from a root directory of an external repository. Closes ckeditor/ckeditor5#12757.
```

- Bumped the type from `Internal` to `Other` to show it in the changelog. From my point of view, this is more than just an invisible fix, but rather a _small feature_.
- The previous scope `dev` was invalid. Changed to `tests`.
- Changed the message a bit and added a dot a the end.

PR #820, opened 2022-11-25
Translate source loader should pass through the source map object

pomek:

Fix (webpack-plugin): Fixed generating source maps after switching the order of loaders in the CKEditor 5 translations webpack plugin. The source loader should pass through the source map object to avoid generating a new one. Closes ckeditor/ckeditor5#12928.

---

### Additional information


psmyrek (approved):
LGTM 🎉

PR #825, opened 2022-12-15
Cleanup after removed `ckeditor5-dev-env` and add missing exports for `web-crawler`, `transifex` and `release-tools`.

martnpaneq:

Internal: Cleanup after removed `ckeditor5-dev-env` and add missing exports for `web-crawler`, `transifex` and `release-tools`.

---

### Additional information

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ script:
   - yarn run coverage
 after_script:
   - export END_TIME=$( date +%s )
-  - ./packages/ckeditor5-dev-tests/bin/notify-travis-status.js
+  - ./packages/ckeditor5-dev-ci/bin/notifytravisstatus.js
 after_success:
   - yarn add coveralls --ignore-workspace-root-check
   - cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
--- a/package.json
+++ b/package.json
@@ -2,9 +2,6 @@
   "name": "ckeditor5-dev",
   "version": "31.1.13",
   "private": true,
-  "dependencies": {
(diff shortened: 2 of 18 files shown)


martnpaneq:
I was wondering if instead of exporting `ckeditor5-dev-web-crawler` internal functions and constants mabe it's a better idea to extract the cli function from `ckeditor5` package: https://github.com/ckeditor/ckeditor5/blob/master/scripts/web-crawler/index.js#L32

pomek:
The [`parseArguments()`](https://github.com/ckeditor/ckeditor5/blob/master/scripts/web-crawler/index.js#L32-L59) without the `--docs` and `--manual` modifiers could be moved to `ckeditor5-dev-web-crawler` package. It should simplify the crawler integration itself in CKE 5.

pomek:
> The [`parseArguments()`](https://github.com/ckeditor/ckeditor5/blob/master/scripts/web-crawler/index.js?rgh-link-date=2022-12-19T19%3A04%3A16Z#L32-L59) without the `--docs` and `--manual` modifiers could be moved to `ckeditor5-dev-web-crawler` package. It should simplify the crawler integration itself in CKE 5.

We agreed to abond the idea as it requires changes in both: CKE 5 and CKE 5 Dev.

martnpaneq:
Turns out that it's not as easy to just move [parseArguments()](https://github.com/ckeditor/ckeditor5/blob/master/scripts/web-crawler/index.js#L32-L59) to `ckeditor5-dev-web-crawler`. It's used in `ckeditor5` and in `docs` and the implementations differ. So I think we should leave it as it is now.

PR #831, opened 2023-01-11
Added validation mechanism for API docs generated from TypeDoc

psmyrek:

Feature (docs): Added validation mechanism for API docs generated from TypeDoc. Closes ckeditor/ckeditor5#12965.

Internal (typedoc-plugins): Added `event:` prefix for event names to make sure they are unique in the parent's scope.

---

### Additional information


packages/ckeditor5-dev-docs/lib/validators/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+				errors.add( error );
+			} );
+
+			errors.forEach( error => typeDoc.logger.warn( error ) );
+		}
+
+		typeDoc.logger.info( 'Validation completed.' );
+
+		return result;
```
pomek, line 47:
```suggestion
		return errors.size === 0;
```
or
```suggestion
		return !errors.size;
```

Then, we can remove the `result` variable and simplify the `onError` callback.

packages/ckeditor5-dev-docs/lib/validators/index.js
```diff
@@ -0,0 +1,49 @@
...
+		for ( const validator of validators ) {
+			const errors = new Set();
+
+			validator( project, error => {
+				result = false;
+
+				errors.add( error );
+			} );
+
+			errors.forEach( error => typeDoc.logger.warn( error ) );
```
pomek, line 42: Is it possible to add a node from the source code where an error occurs? See: https://github.com/ckeditor/ckeditor5-dev/blob/7472b759bc6ccc380dc6016a63ec4a6da737a0ed/packages/typedoc-plugins/lib/tag-event/index.js#L75.

packages/ckeditor5-dev-docs/lib/validators/index.js
psmyrek: I simplified the main loop, thanks.

packages/ckeditor5-dev-docs/lib/validators/index.js
psmyrek: I already tried to do that, but it turns out that not all reflections have a symbol: `project.getSymbolFromReflection( reflection )` returns `undefined`. No idea how to get the internal `node` for such reflections.

psmyrek: I tried again and this time it worked. Now, all errors found by the validators will be reported along with the node from the source code that has them.

obraz

pomek:
It works as expected. The last thing we need before merging is to cover [684a617](https://github.com/ckeditor/ckeditor5-dev/pull/831/commits/684a6174bf38d43a99c6e8229571ce55c71fde21) (tests).

PR #836, opened 2023-02-03
Fixed the `tag-event` plugin that crashed when trying to parse a complex events parameters

psmyrek:

Fix (typedoc-plugins): Fixed the `tag-event` plugin that crashed when trying to parse event parameters, which are passed as type arguments. Closes ckeditor/ckeditor5#13376.

---

### Additional information

It seems that the API docs that previously caused the crash are now generated correctly.

---

from type arg

---

complex event arg


pomek (approved):
LGTM.

PR #837, opened 2023-02-08
Excluded non-public structures from API docs

przemyslaw-zan:

Other: Excluded non-public structures from API docs. Closes ckeditor/ckeditor5#13302.

---

### Additional information


martnpaneq (approved):

martnpaneq (approved):

packages/typedoc-plugins/lib/purge-private-api-docs/index.js
```diff
@@ -42,25 +42,61 @@ function onEventEnd() {
...
+				context.project.removeReflection( reflection );
+
+				continue;
+			}
+
+			removeUrlSourcesFromReflection( reflection );
+
+			const moduleFileName = reflection.sources[ 0 ].fileName;
+
+			const localReflections = Object.values( context.project.reflections ).filter( refl => {
```
psmyrek, line 55: OK, from this line it was unnecessarily complicated. There is no need to iterate twice over all reflections in the whole project, since we must check only children of the given module.

It seems to me that it is enough to write a function, that will get a private module reflection as input and then recursively check which children to remove. Something like:

```js
function removeNonPublicMembers( reflection ) {
    if ( isNonPublicMember( reflection ) ) {
        context.project.removeReflection( reflection );

        return;
    }

    if ( !reflection.children ) {
        return;
    }

    reflection.children.forEach( removeNonPublicMembers );
}
```

In the `isNonPublicMember( reflection )` function check the conditions described in the issue. Return `true` when:
- the reflection is not inherited or inherited from a non-public package
and
- the reflection is private, protected or internal.

psmyrek (approved):
LGTM 🎉

PR #840, opened 2023-02-15
Swapped order of ts-loader and ck-debug-loader

przemyslaw-zan:

Fix (tests): Swapped an order of loaders when processing the TypeScript files. The `ck-debug-loader` is executed before `ts-loader` to avoid the removal of `CK_DEBUG_*` comments from the source code. See ckeditor/ckeditor5#13381.

MAJOR BREAKING CHANGE (tests): The code specified in the `CK_DEBUG_*` flags must be valid TypeScript code as it is processed before `ts-loader`.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getwebpackconfig.js
+++ b/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getwebpackconfig.js
@@ -87,12 +87,6 @@ module.exports = function getWebpackConfigForAutomatedTests( options ) {
 					test: /\.ts$/,
 					use: [
 						{
-							loader: require.resolve( '../ck-debug-loader' ),
-							options: {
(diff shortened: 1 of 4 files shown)


pomek (changes requested):
We need two tests (single per environment) that prevent reverting the order of loaders in the future.

So, a test should check if `ck-debug-loader` is specified after `ts-loader` or executed before. The assertion depends on the description of the test.

pomek:
Merge message: missing scope, missing "why".

pomek:
Today we decided to merge it as a breaking change.

pomek:
The proposed commit message is still invalid. Please, see: https://ckeditor.com/docs/ckeditor5/latest/framework/guides/contributing/git-commit-message-convention.html.

pomek (approved):
LGTM.

PR #842, opened 2023-02-22
Skip derived class if it no longer exists

psmyrek:

Fix (typedoc-plugins): The `typedoc-plugin-event-inheritance-fixer` plugin should skip the derived class and not try to add any inherited events if the class no longer exists. Perhaps it could be removed from the project by the `typedoc-plugin-purge-private-api-docs` plugin.

---

### Additional information

Added new test that removes one derived class.

On `#master` branch:

On `#typedoc-event-inheritance-fix` branch:


pomek (approved):
LGTM.

PR #843, opened 2023-02-23
Fixed coverage with ck-debug flags

pomek:

Fix (tests): The `ck-debug-loader` should not be included in TypeScript files to avoid the "Unable to lookup source" error when checking the coverage. Instead, `babel-loader` should handle the `CK_DEBUG` flags. Closes ckeditor/ckeditor5#13529.


martnpaneq (approved):
Looks like the fix works fine :)

image

przemyslaw-zan:
I could not reproduce the issue. I've tried this PR and it works as well, so LGTM i suppose.

przemyslaw-zan (approved):

PR #845, opened 2023-02-24
When publishing packages, adds the types field to pkg.json

pomek:

Feature (release-tools): The `releaseSubrepositories()` task adds the `types` field (if not specified) to `package.json` while publishing packages based on the `main` field if a package is written in TypeScript. Closes ckeditor/ckeditor5#13518.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/tasks/releasesubrepositories.js
+++ b/packages/ckeditor5-dev-release-tools/lib/tasks/releasesubrepositories.js
@@ -415,6 +415,7 @@ module.exports = async function releaseSubRepositories( options ) {
 		// The keys in returned map are file patterns, and their values represent number of matched files. If there is no `#files` key
 		// in `package.json`, then empty map is returned.
 		function getMatchedFilesToPublish( packageJson, repositoryPath ) {
+			// TODO: Include the `main` and `types` properties if they are specified.
 			if ( !packageJson.files ) {
 				return new Map();
 			}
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-dev-release-tools/lib/tasks/releasesubrepositories.js
```diff
@@ -727,10 +727,17 @@ module.exports = async function releaseSubRepositories( options ) {
...
 
 			if ( hasTypeScriptEntryPoint ) {
 				tools.updateJSONFile( packageJsonPath, jsonFile => {
-					jsonFile.main = jsonFile.main.replace( '.ts', '.js' );
+					const { main } = jsonFile;
+
+					jsonFile.main = main.replace( /\.ts$/, '.js' );
+
+					if ( !hasTypesProperty ) {
+						jsonFile.types = main.replace( /\.ts$/, '.d.ts' );
```
martnpaneq, line 739: All looks fine, just a loose thought. Is it a problem when we specify `"types": "index.d.ts"` when the `index.d.ts` file does not exist? Probably this is not an issue, but maybe it's worth checking.

martnpaneq (approved):

psmyrek (approved):
LGTM 🎉

packages/ckeditor5-dev-release-tools/lib/tasks/releasesubrepositories.js
pomek: I added a check.

image

I agree it is not the best design here, but let's be safe than sorry.

PR #847, opened 2023-03-01
Allow specifying `tsconfig` path in automated tests

filipsobol:

Other (tests): Allow specifying the `tsconfig` option in automated and manual tests. The path is resolved using the current working directory). If not specified, it will be resolved to default values that the `ts-config` resolves.


pomek (commented):
LGTM. However, it deserves to be covered in unit tests to avoid accidentally removing it in the future.

Both webpack config functions contain them.

filipsobol:
@pomek I added few tests that should cover this change.

pomek (approved):
LGTM.

pomek:
Now it looks good.

image

PR #853, opened 2023-03-07
Typedoc - event plugin should not guessing an event owner

psmyrek:

Other (typedoc-plugins): Changed the way events are defined. The value of the `@eventName` tag must be a valid identifier (link) to a class or an interface, to which given event should be attached. Closes ckeditor/ckeditor5#13553.

Other (typedoc-plugins): Extended the scope of the `typedoc-plugin-event-inheritance-fixer` plugin. So far only inherited events in classes and in the `Observable` interface have been checked. Now, support for inherited events has been added to all interfaces, so all derived classes and interfaces will contain all inherited events.

Feature (typedoc-plugins): Added support for multiple `@eventName` tags defined for one exported type. One exported type can now be used to create multiple events, even in different (not related to each other) classes or interfaces.

Internal (typedoc-plugins): Added all TypeDoc plugins (as paths to the plugin source) to the export from the package. This makes it easier to use them when creating a TypeDoc instance.

MAJOR BREAKING CHANGE (typedoc-plugins): Changed the way events are defined. Previously, the value from the `@eventName` tag was simply the name of the event and the `typedoc-plugin-tag-event` plugin was trying to guess which reflection should be the owner of this event. Now, the `@eventName` value must be a valid identifier (link) to the event owner, which can be a class or an interface, to which given event will be attached. The owner of every event must be explicitly defined. The identifier can be relative or absolute.

---

### Additional information


pomek (approved):
LGTM.

PR #856, opened 2023-03-13
Building docs when processing a non-es6 module

pomek:

Fix (typedoc-plugins): The building documentation process should not crash when processing a non-es6 module file while building from the TypeScript sources. Closes ckeditor/ckeditor5#13662.

Fix (docs): The validator should not crash when processing a non-es6 module file while building from the TypeScript sources. Closes ckeditor/ckeditor5#13662.

---

### Additional information


psmyrek (approved):
LGTM 🎉

PR #861, opened 2023-03-27
Bumped slack-notify, fixed timestamp

przemyslaw-zan:

Other (ci): Bumped a version of the `slack-notify` library.

Fix (ci): Fixed the `notifyTravisStatus()` function providing incorrect values for build times lasting 1 hour or longer.

---

### Additional information


packages/ckeditor5-dev-ci/lib/notifytravisstatus.js
pomek, line 1:
```js
	slack.onError = err => {
		console.log( 'API error occurred:', err );
	};
```

This callback should be moved as `catch()` as it is not supported in the latest version of the library.

pomek (changes requested):
See: https://github.com/ckeditor/ckeditor5-dev/pull/861/files#r1150241864.

pomek (approved):
LGTM.

PR #864, opened 2023-04-06
Mark GitHub release as a pre-release when publishing a non-latest version

psmyrek:

Other (release-tools): Mark GitHub release as a pre-release when publishing a non-latest version. Closes ckeditor/ckeditor5#13664.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/tasks/releasesubrepositories.js
+++ b/packages/ckeditor5-dev-release-tools/lib/tasks/releasesubrepositories.js
@@ -873,7 +873,8 @@ module.exports = async function releaseSubRepositories( options ) {
 			repositoryOwner: releaseDetails.repositoryOwner,
 			repositoryName: releaseDetails.repositoryName,
 			version: `v${ releaseDetails.version }`,
-			description: releaseDetails.changes
+			description: releaseDetails.changes,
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-dev-release-tools/lib/tasks/releasesubrepositories.js
```diff
@@ -873,7 +873,8 @@ module.exports = async function releaseSubRepositories( options ) {
 			repositoryOwner: releaseDetails.repositoryOwner,
 			repositoryName: releaseDetails.repositoryName,
 			version: `v${ releaseDetails.version }`,
-			description: releaseDetails.changes
+			description: releaseDetails.changes,
+			isPrerelease: npmTag !== 'latest'
```
pomek, line 877: What if we publish only on GH? We can pass the modifier, but publishing a release on GH only and passing an npm version does not make sense.

I would change the implementation to assert the returned value from `semver.prerelease( version )`.

Does it make sense?

psmyrek: Of course, thanks for noticing such a scenario.

pomek (approved):
LGTM.

PR #867, opened 2023-04-27
Fixed crash in the Travis notifier if an author of a commit could not be obtained

psmyrek:

Fix (ci): Fixed crash in the Travis notifier script if an author of a commit could not be obtained. Instead, ping the entire channel and mention an author from its name instead of a login that can be `null`.

### DIFF
--- a/packages/ckeditor5-dev-ci/lib/notifytravisstatus.js
+++ b/packages/ckeditor5-dev-ci/lib/notifytravisstatus.js
@@ -34,11 +34,10 @@ module.exports = async function notifyTravisStatus( options ) {
 		notifyCommitUrl: options.notifyCommitUrl,
 		repositorySlug: options.repositorySlug
 	} );
-	const commitDetails = await getCommitDetails( commitUrl, options.githubToken );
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-dev-ci/lib/notifytravisstatus.js
```diff
@@ -34,19 +34,19 @@ module.exports = async function notifyTravisStatus( options ) {
 		notifyCommitUrl: options.notifyCommitUrl,
 		repositorySlug: options.repositorySlug
 	} );
-	const commitDetails = await getCommitDetails( commitUrl, options.githubToken );
+	const { githubAccount, commitAuthor, commitMessage } = await getCommitDetails( commitUrl, options.githubToken );
 
 	const [ buildRepoOwner, buildRepoName ] = options.repositorySlug.split( '/' );
-
-	const slackAccount = members[ commitDetails.author ];
+	const slackAccount = members[ githubAccount ] || null;
```
psmyrek, line 40: Setting `null` here only to match the possible `slackAccount` values with the `githubAccount` values: both may be `string` or `null`.

pomek (approved):
LGTM.

PR #868, opened 2023-04-27
Added repository preparation script

przemyslaw-zan:

Feature (release-tools): Added repository preparation script. Closes ckeditor/ckeditor5#13954.

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
+++ b/packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
@@ -5,64 +5,98 @@
 
 'use strict';
 
-const upath = require( 'upath' );
 const fs = require( 'fs-extra' );
-
-const pkgJsonTemplatePath = upath.join( __dirname, '..', 'templates', 'release-package.json' );
+const upath = require( 'upath' );
 
 /**
  * @param {Object} options
+ * @param {String} options.outputDirectory Relative path to the destination directory where packages will be stored.
  * @param {String} [options.cwd] Root of the repository to prepare. `process.cwd()` by default.
- * @param {String} [options.packagesDir] Path to directory containing packages. `packages` by default.
- * @param {String} [options.outputDir] Path to the output directory of this script. `release` by default.
- * @param {Object} options.pkgJsonContent Object containing values to use in the created package json file.
- * @param {Array<String>} options.rootFilesToCopy Array containing list of files or directories to copy.
- * @param {Array<String>} [options.packagesToCopy] Optional argument specifying packages to copy from the `packagesDir`.
- * This can contain nested package paths. If left empty, all non-nested packages from the `packagesDir` will be processed.
+ * @param {String} [options.packagesDirectory] Relative path to a location of packages.
+ * If specified, all of the found packages will be copied.
+ * @param {Array.<String>} [options.packagesToCopy] List of packages that should be processed.
+ * If not specified, all packages found in `packagesDirectory` are considered.
+ * @param {Object} [options.rootPackage] Definition of a root package (found in the `cwd` directory).
+ * If not specified, the root package will not be created.
+ * @param {Object} options.rootPackage.packageJson Object containing values to use in the created the `package.json` file.
+ * @param {Array.<String>} options.rootPackage.files Array containing a list of files or directories to copy.
  */
 module.exports = async function prepareRepository( options ) {
 	const {
+		outputDirectory,
 		cwd = process.cwd(),
-		packagesDir = 'packages',
-		outputDir = 'release',
-		pkgJsonContent,
-		rootFilesToCopy,
-		packagesToCopy
+		packagesDirectory,
+		packagesToCopy,
+		rootPackage
 	} = options;
 
-	const outputDirPath = upath.join( cwd, outputDir );
-	const packagesDirPath = upath.join( cwd, packagesDir );
-	const packagesOutputDir = upath.join( outputDirPath, packagesDir );
+	if ( !rootPackage && !packagesDirectory ) {
+		return;
+	}
 
-	await fs.ensureDir( packagesOutputDir );
+	const outputDirectoryPath = upath.join( cwd, outputDirectory );
+	await fs.emptyDir( outputDirectoryPath );
 
-	// Creating root package.json file.
-	const pkgJsonOutputPath = upath.join( outputDirPath, 'package.json' );
-	const pkgJsonTemplate = await fs.readJson( pkgJsonTemplatePath );
-	const pkgJsonNewContent = { ...pkgJsonTemplate, ...pkgJsonContent };
-	await fs.writeJson( pkgJsonOutputPath, pkgJsonNewContent, { spaces: 2, EOL: '\n' } );
+	const copyPromises = [];
 
-	// Copying root package files.
-	const copyRootItemsPromises = rootFilesToCopy.map( item => {
+	if ( rootPackage ) {
+		const copyRootItemsPromises = await processRootPackage( {
+			cwd,
+			rootPackage,
+			outputDirectoryPath
+		} );
+
+		copyPromises.push( ...copyRootItemsPromises );
+	}
+
+	if ( packagesDirectory ) {
+		const copyPackagesPromises = await processMonorepoPackages( {
+			cwd,
+			packagesDirectory,
+			packagesToCopy,
+			outputDirectoryPath
+		} );
+
+		copyPromises.push( ...copyPackagesPromises );
+	}
+
+	return Promise.all( copyPromises );
+};
+
+async function processRootPackage( { cwd, rootPackage, outputDirectoryPath } ) {
+	const pkgJsonOutputPath = upath.join( outputDirectoryPath, 'package.json' );
+	await fs.writeJson( pkgJsonOutputPath, rootPackage.packageJson, { spaces: 2, EOL: '\n' } );
+
+	return rootPackage.files.map( item => {
 		const itemPath = upath.join( cwd, item );
-		const itemOutputPath = upath.join( outputDirPath, item );
+		const itemOutputPath = upath.join( outputDirectoryPath, item );
 
 		return fs.copy( itemPath, itemOutputPath );
 	} );
+}
 
-	// Copying packages.
-	const packageDirs = packagesToCopy || await fs.readdir( packagesDirPath );
+async function processMonorepoPackages( { cwd, packagesDirectory, packagesToCopy, outputDirectoryPath } ) {
+	const packagesDirectoryPath = upath.join( cwd, packagesDirectory );
+	const packagesOutputDirectoryPath = upath.join( outputDirectoryPath, packagesDirectory );
+	const packageDirs = packagesToCopy || await fs.readdir( packagesDirectoryPath );
 
-	const copyPackagesPromises = packageDirs.map( packageDir => {
-		return fs.copy(
-			upath.join( packagesDirPath, packageDir ),
-			upath.join( packagesOutputDir, packageDir )
-		);
+	await fs.ensureDir( packagesOutputDirectoryPath );
+
+	return packageDirs.map( async packageDir => {
+		const packagePath = upath.join( packagesDirectoryPath, packageDir );
+		const isDir = ( await fs.lstat( packagePath ) ).isDirectory();
+
+		if ( !isDir ) {
+			return;
+		}
+
+		const pkgJsonPath = upath.join( packagePath, 'package.json' );
+		const hasPkgJson = await fs.existsSync( pkgJsonPath );
+
+		if ( !hasPkgJson ) {
+			return;
+		}
+
+		return fs.copy( packagePath, upath.join( packagesOutputDirectoryPath, packageDir ) );
 	} );
-
-	// Waiting for all copy processes to end.
-	await Promise.all( [
-		...copyRootItemsPromises,
-		...copyPackagesPromises
-	] );
-};
+}
--- a/packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
+++ b/packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
@@ -13,42 +13,51 @@ const pkgJsonTemplatePath = upath.join( __dirname, '..', 'templates', 'release-p
 /**
  * @param {Object} options
  * @param {String} [options.cwd] Root of the repository to prepare. `process.cwd()` by default.
- * @param {String} options.packagesDir Path to directory containing packages.
- * @param {String} options.outputDir Path to the output directory of this script.
- * @param {Array<String>} options.rootFilesToCopy Array containing list of files or directories to copy.
+ * @param {String} [options.packagesDir] Path to directory containing packages. `packages` by default.
+ * @param {String} [options.outputDir] Path to the output directory of this script. `release` by default.
  * @param {Object} options.pkgJsonContent Object containing values to use in the created package json file.
+ * @param {Array<String>} options.rootFilesToCopy Array containing list of files or directories to copy.
+ * @param {Array<String>} [options.packagesToCopy] Optional argument specifying packages to copy from the `packagesDir`.
+ * This can contain nested package paths. If left empty, all non-nested packages from the `packagesDir` will be processed.
  */
-module.exports = async function prepareRepository( options = {} ) {
-	const cwd = upath.toUnix( options.cwd || process.cwd() );
+module.exports = async function prepareRepository( options ) {
+	const {
+		cwd = process.cwd(),
+		packagesDir = 'packages',
+		outputDir = 'release',
+		pkgJsonContent,
+		rootFilesToCopy,
+		packagesToCopy
+	} = options;
 
-	const outputDir = upath.join( cwd, options.outputDir );
-	const packagesDir = upath.join( cwd, options.packagesDir );
-	const packagesOutputDir = upath.join( outputDir, options.packagesDir );
+	const outputDirPath = upath.join( cwd, outputDir );
+	const packagesDirPath = upath.join( cwd, packagesDir );
+	const packagesOutputDir = upath.join( outputDirPath, packagesDir );
 
 	await fs.ensureDir( packagesOutputDir );
 
 	// Creating root package.json file.
-	const pkgJsonOutputPath = upath.join( outputDir, 'package.json' );
+	const pkgJsonOutputPath = upath.join( outputDirPath, 'package.json' );
 	const pkgJsonTemplate = await fs.readJson( pkgJsonTemplatePath );
-	const pkgJsonContent = { ...pkgJsonTemplate, ...options.pkgJsonContent };
-	await fs.writeJSON( pkgJsonOutputPath, pkgJsonContent );
+	const pkgJsonNewContent = { ...pkgJsonTemplate, ...pkgJsonContent };
+	await fs.writeJson( pkgJsonOutputPath, pkgJsonNewContent, { spaces: 2, EOL: '\n' } );
 
 	// Copying root package files.
-	const copyRootItemsPromises = options.rootFilesToCopy.map( item => {
+	const copyRootItemsPromises = rootFilesToCopy.map( item => {
 		const itemPath = upath.join( cwd, item );
-		const itemOutputPath = upath.join( outputDir, item );
+		const itemOutputPath = upath.join( outputDirPath, item );
 
 		return fs.copy( itemPath, itemOutputPath );
 	} );
 
 	// Copying packages.
-	const packageNames = await fs.readdir( packagesDir );
+	const packageDirs = packagesToCopy || await fs.readdir( packagesDirPath );
 
-	const copyPackagesPromises = packageNames.map( packageName => {
-		const packageDir = upath.join( packagesDir, packageName );
-		const packageOutputDir = upath.join( packagesOutputDir, packageName );
-
-		return fs.copy( packageDir, packageOutputDir );
+	const copyPackagesPromises = packageDirs.map( packageDir => {
+		return fs.copy(
+			upath.join( packagesDirPath, packageDir ),
+			upath.join( packagesOutputDir, packageDir )
+		);
 	} );
 
 	// Waiting for all copy processes to end.
--- a/packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
+++ b/packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const fs = require( 'fs-extra' );
+const glob = require( 'glob' );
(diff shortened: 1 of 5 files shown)


martnpaneq:
Today we had a discussion with the team about unifying usage of `async` / `await` in this project.
We had an idea to either use it in every single place, or stop using it when it's not required.

We settled on the latter, to not use it, when it's not needed. Here are our arguments:
1. The code should be as short as possible - it's easier to understand and maintain.
2. It's easier to test without async.
3. When using `async` only when required, it's clear which function are truly async (like workers or BE calls), and which not.
4. Using non-blocking functions make sense for a node server (to speed up responses), but it makes no difference for a local script.

We should probably talk about it once again when @przemyslaw-zan gets back from holidays.

przemyslaw-zan:
In case of specifically this script, synchronous code takes over two and a half times longer to execute:

| Async | Sync |
|-------|------|
| ~6s   | ~17s |

_Based on several dozen runs of the script._

Lets use `async` code wherever it makes sense, such as here, but lets also keep in mind that not every script has to be `async`.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * @param {Object} options
+ * @param {String} [options.cwd] Root of the repository to prepare. `process.cwd()` by default.
+ * @param {String} [options.packagesDir] Path to directory containing packages. `packages` by default.
+ * @param {String} [options.outputDir] Path to the output directory of this script. `release` by default.
+ * @param {Object} options.pkgJsonContent Object containing values to use in the created package json file.
+ * @param {Array<String>} options.rootFilesToCopy Array containing list of files or directories to copy.
+ * @param {Array<String>} [options.packagesToCopy] Optional argument specifying packages to copy from the `packagesDir`.
+ * This can contain nested package paths. If left empty, all non-nested packages from the `packagesDir` will be processed.
```
pomek, line 21: Let's sort these properties from required to optional ones.

```suggestion
 * @param {Object} options
 * @param {String} options.outputDirectory Relative path to the destination directory where packages will be stored.
 * @param {String} [options.cwd] Root of the repository to prepare. `process.cwd()` by default.
 * @param {String} [options.packagesDirectory] Relative path to a location of packages. If specified, all of the found packages will be copied.
 * @param {Array.<String>} [options.packagesToCopy] List of packages that should be processed. If not specified, all packages found in `packagesDirectory` are considered.
* @param {Object} [options.rootPackage] Definition of a root package (found in the `cwd` directory). If not specified, the root package will not be created.
 * @param {Object} options.rootPackage.packageJson Object containing values to use in the created the `package.json` file.
 * @param {Array.<String>} options.rootPackage.files Array containing a list of files or directories to copy.
```

As I am unsure about double nested properties, we can consider creating a typedef describing the root package fields.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,68 @@
...
+	// Copying root package files.
+	const copyRootItemsPromises = rootFilesToCopy.map( item => {
+		const itemPath = upath.join( cwd, item );
+		const itemOutputPath = upath.join( outputDirPath, item );
+
+		return fs.copy( itemPath, itemOutputPath );
+	} );
+
+	// Copying packages.
+	const packageDirs = packagesToCopy || await fs.readdir( packagesDirPath );
```
pomek, line 54: Will it copy a file too? We should filter out non-directories – symlinks and files.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,68 @@
...
+			upath.join( packagesDirPath, packageDir ),
+			upath.join( packagesOutputDir, packageDir )
+		);
+	} );
+
+	// Waiting for all copy processes to end.
+	await Promise.all( [
+		...copyRootItemsPromises,
+		...copyPackagesPromises
+	] );
```
pomek, line 67:
```suggestion
	return Promise.all( [
		...copyRootItemsPromises,
		...copyPackagesPromises
	] );
```

But it's minor.

packages/ckeditor5-dev-release-tools/tests/tasks/preparerepository.js
```diff
@@ -0,0 +1,237 @@
...
+					readJson: sinon.stub().callsFake( stubReject ),
+					readdir: sinon.stub().callsFake( stubReject )
+				},
+				process: {
+					cwd: sinon.stub( process, 'cwd' ).returns( 'current/working/dir' )
+				}
+			};
+
+			stubs.fs.readJson.withArgs( sinon.match( pkgJsonTemplatePathPattern ) ).resolves( pkgJsonTemplate );
+			stubs.fs.readdir.withArgs( 'current/working/dir/packages' ).resolves( packages );
```
pomek, line 72: In such a case, I would assume it will be clearer if you use `callsFake()` instead of `withArgs()` and, depending on an argument, return a proper value.

pomek: I would aim to remove regexps from tests and use `String.endsWith()` instead.

packages/ckeditor5-dev-release-tools/tests/tasks/preparerepository.js
```diff
@@ -0,0 +1,237 @@
...
+			expect( prepareRepository ).to.be.a( 'function' );
+		} );
+
+		it( 'should ensure the existence of the output directory', async () => {
+			await prepareRepository( options );
+
+			expect( stubs.fs.ensureDir.callCount ).to.equal( 1 );
+			expect( stubs.fs.ensureDir.getCall( 0 ).args.length ).to.equal( 1 );
+			expect( stubs.fs.ensureDir.getCall( 0 ).args[ 0 ] ).to.equal( 'current/working/dir/release/packages' );
+		} );
```
pomek, line 101: Out loud thinking: shouldn't we clear the directory too?

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,68 @@
...
+	const pkgJsonNewContent = { ...pkgJsonTemplate, ...pkgJsonContent };
+	await fs.writeJson( pkgJsonOutputPath, pkgJsonNewContent, { spaces: 2, EOL: '\n' } );
+
+	// Copying root package files.
+	const copyRootItemsPromises = rootFilesToCopy.map( item => {
+		const itemPath = upath.join( cwd, item );
+		const itemOutputPath = upath.join( outputDirPath, item );
+
+		return fs.copy( itemPath, itemOutputPath );
+	} );
```
pomek, line 51: This is an opt-in step. It makes sense when releasing the `ckeditor5` package, but it doesn't for, e.g., `ckeditor5-dev`.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,68 @@
...
+
+	// Copying packages.
+	const packageDirs = packagesToCopy || await fs.readdir( packagesDirPath );
+
+	const copyPackagesPromises = packageDirs.map( packageDir => {
+		return fs.copy(
+			upath.join( packagesDirPath, packageDir ),
+			upath.join( packagesOutputDir, packageDir )
+		);
+	} );
```
pomek, line 61: The preparation step should be able to prepare a single package, e.g., `mrgit`. Hence, the `packagesDirectory` argument must be specified. Otherwise, process only a root package if its configuration is available.

The task may create the output directory without any files when missing the proper configuration. Make sure to cover the case in a unit test.

packages/ckeditor5-dev-release-tools/tests/tasks/preparerepository.js
```diff
@@ -0,0 +1,237 @@
...
+			expect( stubs.fs.copy.getCall( 0 ).args.length ).to.equal( 2 );
+			expect( stubs.fs.copy.getCall( 0 ).args[ 0 ] ).to.equal( 'current/working/dir/src' );
+			expect( stubs.fs.copy.getCall( 0 ).args[ 1 ] ).to.equal( 'current/working/dir/release/src' );
+
+			expect( stubs.fs.copy.getCall( 1 ).args.length ).to.equal( 2 );
+			expect( stubs.fs.copy.getCall( 1 ).args[ 0 ] ).to.equal( 'current/working/dir/CHANGELOG.md' );
+			expect( stubs.fs.copy.getCall( 1 ).args[ 1 ] ).to.equal( 'current/working/dir/release/CHANGELOG.md' );
+		} );
+
+		it( 'should copy files of all packages by default', async () => {
```
pomek, line 200: By default, it should not copy anything. A user must define the package's directory, and then the script can take it.

pomek (changes requested):
As mentioned.

I can see that the things I described are not mentioned in the original scope. However, we talked a lot about the new toolings, and we must be aware that the same scripts will be used outside the CKEditor 5 project, e.g., pkg-gen, Umberto, mrgit, etc. Let's make it universal as it is possible.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,68 @@
...
+		cwd = process.cwd(),
+		packagesDir = 'packages',
+		outputDir = 'release',
+		pkgJsonContent,
+		rootFilesToCopy,
+		packagesToCopy
+	} = options;
+
+	const outputDirPath = upath.join( cwd, outputDir );
+	const packagesDirPath = upath.join( cwd, packagesDir );
```
psmyrek, line 34: We should support the case that `packagesDir` is not provided.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,68 @@
...
+	const packagesOutputDir = upath.join( outputDirPath, packagesDir );
+
+	await fs.ensureDir( packagesOutputDir );
+
+	// Creating root package.json file.
+	const pkgJsonOutputPath = upath.join( outputDirPath, 'package.json' );
+	const pkgJsonTemplate = await fs.readJson( pkgJsonTemplatePath );
+	const pkgJsonNewContent = { ...pkgJsonTemplate, ...pkgJsonContent };
```
psmyrek, line 42: What if `pkgJsonContent` will not contain all fields that are provided in `pkgJsonTemplate`? Currently, `null` values will be stored in the generated `package.json`. We should probably filter out those falsy keys.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
psmyrek: ... and probably directories that are not "packages" - e.g. does not contain `package.json` file.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
przemyslaw-zan: In that case, what is the purpose of the template? It provides only the `null` values. It sounds like we should just simply write the `pkgJsonContent` alone as the new `package.json`.

przemyslaw-zan:
To be decided:
- [Typedef describing the root package fields.](https://github.com/ckeditor/ckeditor5-dev/pull/868#discussion_r1185917146)
- [Should the output directory be cleared?](https://github.com/ckeditor/ckeditor5-dev/pull/868#discussion_r1185922993) (Current implementation clears it.)
- [Is `pkgJsonTemplate` needed?](https://github.com/ckeditor/ckeditor5-dev/pull/868#discussion_r1185936133)

pomek:
We decide to simplify the input arguments:

```js
prepareRepository( {
	packageJson: {
		dependencies: [],
		description: '...',
		name: require('./package.json').name, // Read from the original file.
		version: require('./package.json').version,  // Read from the original file.
		files: [
			// Files to copy. It must be defined. Otherwise, throw.
		];
	},
} )
```

The expected output:

* release/ckeditor5
* release/ckeditor5-utils
* And other CKEditor 5 packages

> Should the output directory be cleared?

Let's throw an error if the directory is not clear. An example case: when publishing packages, npm reports an error. You can tap the arrow up twice and execute the preparation script by mistake.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,115 @@
...
+ * @param {String} [options.packagesDirectory] Relative path to a location of packages.
+ * If specified, all of the found packages will be copied.
+ * @param {Array.<String>} [options.packagesToCopy] List of packages that should be processed.
+ * If not specified, all packages found in `packagesDirectory` are considered.
+ * @param {Object} [options.rootPackageJson] Object containing values to use in the created the `package.json` file.
+ * If not specified, the root package will not be created.
+ * @param {String} options.rootPackageJson.name Name of the package. Required value.
+ * @param {Array.<String>} options.rootPackageJson.files Array containing a list of files or directories to copy. Required value.
+ */
+module.exports = async function prepareRepository( options ) {
```
pomek, line 24:
```suggestion
 * @param {Array.<String>} options.rootPackageJson.files Array containing a list of files or directories to copy. Required value.
 * @returns {Promise}
 */
module.exports = async function prepareRepository( options ) {
```

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,115 @@
...
+	if ( !rootPackageJson && !packagesDirectory ) {
+		return;
+	}
+
+	const outputDirPath = upath.join( cwd, outputDirectory );
+	await fs.ensureDir( outputDirPath );
+	const outputDirContent = await fs.readdir( outputDirPath );
+
+	if ( outputDirContent.length ) {
+		throw new Error( `Output directory is not empty: "${ outputDirPath }"` );
```
pomek, line 42:
```suggestion
		throw new Error( `Output directory is not empty: "${ outputDirPath }".` );
```

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,115 @@
...
+};
+
+async function processRootPackage( { cwd, rootPackageJson, outputDirPath } ) {
+	if ( !rootPackageJson.name ) {
+		throw new Error( '"rootPackageJson" option object must have a "name" field.' );
+	}
+
+	if ( !rootPackageJson.files ) {
+		throw new Error( '"rootPackageJson" option object must have a "files" field.' );
+	}
```
pomek, line 78: I would extract these checks to another function that could be called before processing the root package, e.g., `verifyRootPackage()`. Thanks to that, the body function is clear – it does something and does not care about the validation.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,115 @@
...
+			outputDirPath
+		} );
+
+		copyPromises.push( ...copyPackagesPromises );
+	}
+
+	return Promise.all( copyPromises );
+};
+
+async function processRootPackage( { cwd, rootPackageJson, outputDirPath } ) {
```
pomek, line 71:
```suggestion
async function processRootPackage( { cwd, packageJson, outputDirectoryPath } ) {
```

I am a big fan of using full names, especially with keys.

pomek: Also, please add JSDoc to describe input values. Yeah, I miss TS here.

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,115 @@
...
+
+	return rootPackageJson.files.map( item => {
+		const itemPath = upath.join( cwd, item );
+		const itemOutputPath = upath.join( rootPackageOutputPath, item );
+
+		return fs.copy( itemPath, itemOutputPath );
+	} );
+}
+
+async function processMonorepoPackages( { cwd, packagesDirectory, packagesToCopy, outputDirPath } ) {
```
pomek, line 94: Please, use full names + describe types in JSDoc.

packages/ckeditor5-dev-release-tools/tests/tasks/preparerepository.js
```diff
@@ -0,0 +1,321 @@
...
+				outputDirectory: 'release'
+			};
+
+			function stubReject( stubName, args ) {
+				if ( args.length === 0 ) {
+					throw new Error( `Stub "${ stubName }" expected to receive an argument.` );
+				}
+
+				throw new Error( `No output configured for stub "${ stubName }" with the following args: ${ JSON.stringify( args ) }` );
+			}
```
pomek, line 32: What kind of magic is here? Why do we need it?

packages/ckeditor5-dev-release-tools/tests/tasks/preparerepository.js
```diff
@@ -0,0 +1,321 @@
...
+			stubs.fs.readdir.withArgs( 'current/working/dir/release' ).resolves( [ 'someFile.txt' ] );
+			stubs.fs.readdir.withArgs( 'current/working/dir/packages' ).resolves( [] );
+			options.packagesDirectory = 'packages';
+
+			try {
+				await prepareRepository( options );
+				throw new Error( 'Expected to throw.' );
+			} catch ( err ) {
+				expect( err.message ).to.equal( 'Output directory is not empty: "current/working/dir/release"' );
+			}
```
pomek, line 111:
```suggestion
			await prepareRepository( options )
				.then(
					() => {
						throw new Error( 'Expected to throw.' );
					},
					err => {
						expect( err.message ).to.equal( 'Output directory is not empty: "current/working/dir/release"' );
					}
				);
```

I would avoid using `try/catch` and `await`. It looks bad.

packages/ckeditor5-dev-release-tools/tests/tasks/preparerepository.js
```diff
@@ -0,0 +1,321 @@
...
+
+				expect( stubs.fs.copy.callCount ).to.equal( 2 );
+
+				expect( stubs.fs.copy.getCall( 0 ).args.length ).to.equal( 2 );
+				expect( stubs.fs.copy.getCall( 0 ).args[ 0 ] ).to.equal( 'current/working/dir/packages/ckeditor5-core' );
+				expect( stubs.fs.copy.getCall( 0 ).args[ 1 ] ).to.equal( 'current/working/dir/release/ckeditor5-core' );
+
+				expect( stubs.fs.copy.getCall( 1 ).args.length ).to.equal( 2 );
+				expect( stubs.fs.copy.getCall( 1 ).args[ 0 ] ).to.equal( 'current/working/dir/packages/ckeditor5-utils' );
+				expect( stubs.fs.copy.getCall( 1 ).args[ 1 ] ).to.equal( 'current/working/dir/release/ckeditor5-utils' );
```
pomek, line 268: Instead of checking what has been copied, I would check if the `'current/working/dir/packages/textFile.txt'` file was not touched.

packages/ckeditor5-dev-release-tools/tests/tasks/preparerepository.js
```diff
@@ -0,0 +1,321 @@
...
+				await prepareRepository( options );
+
+				expect( stubs.fs.copy.callCount ).to.equal( 1 );
+
+				expect( stubs.fs.copy.getCall( 0 ).args.length ).to.equal( 2 );
+				expect( stubs.fs.copy.getCall( 0 ).args[ 0 ] ).to.equal( 'current/working/dir/packages/ckeditor5-core' );
+				expect( stubs.fs.copy.getCall( 0 ).args[ 1 ] ).to.equal( 'current/working/dir/release/ckeditor5-core' );
+			} );
+
+			it( 'should allow copying nested packages via the "packagesToCopy" option', async () => {
```
pomek, line 304: Still determining if we need such a test. I cannot imagine a real use case. Do you?

packages/ckeditor5-dev-release-tools/lib/tasks/preparerepository.js
```diff
@@ -0,0 +1,115 @@
...
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+const fs = require( 'fs-extra' );
+const upath = require( 'upath' );
+
+/**
```
pomek, line 11: Please write a few words about the purpose of this function.

I wonder if it shouldn't be called `prepareRepositoryToRelease()` as `prepareRepository()` does not explain what it prepares. It is placed in the `release-tools` package, though. I am just thinking.

pomek (commented):
Minor changes.

pomek:
Almost R+. Unfortunately, the integration script prints an error:

```
[Error: ENOENT: no such file or directory, stat '/Users/pomek/Projects/ckeditor/ckeditor5/src/*.js'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/Users/pomek/Projects/ckeditor/ckeditor5/src/*.js'
}
```

```js
#!/usr/bin/env node

/**
 * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
 * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
 */

/* eslint-env node */

'use strict';

const releaseTools = require( '@ckeditor/ckeditor5-dev-release-tools' );
const updateVersionReferences = require( './update-version-references' );

const abortController = new AbortController();

const PACKAGES_DIRECTORY = 'packages';
const RELEASE_DIRECTORY = 'release';
const pkgJson = require( '../../package.json' );

process.on( 'SIGINT', () => {
	abortController.abort( 'SIGINT' );
} );

( async () => {
	await releaseTools.prepareRepository( {
		outputDirectory: RELEASE_DIRECTORY,
		packagesDirectory: PACKAGES_DIRECTORY,
		rootPackageJson: {
			name: pkgJson.name,
			files: [
				// Do not add the entire `build/` directory as it contains files produced by internal scripts:
				// automated/manual tests, translations, documentation, content styles.
				// If you need to release anything from the directory, insert a relative path to the file/directory.
				'src/*.js',
				'src/*.d.ts',
				'build/ckeditor5-dll.js',
				'build/ckeditor5-dll.manifest.json',
				'build/translations/*.js'
			]
		}
	} ).catch( err => {
		console.log( err );
		process.exit( 1 );
	} );
} )();
```

psmyrek (approved):
LGTM 🎉

PR #869, opened 2023-04-28
Created a util that allows executing a task in parallel

pomek:

Feature (release-tools): Introduced util called `executeInParallel()` that allows executing a task in parallel in all packages found in a directory specified in the configuration. See ckeditor/ckeditor5#13970.

Feature (utils): The `tools.createSpinner()` allows for creating a counter spinner that improves UX when executing a task specified the number of times.

---

### Additional information

- Example implementation: https://github.com/ckeditor/ckeditor5/pull/14031.

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
@@ -27,7 +27,8 @@ const WORKER_SCRIPT = path.join( __dirname, 'parallelworker.js' );
  * @param {Object} options
  * @param {String} options.packagesDirectory Relative path to a location of packages to execute a task.
  * @param {String} options.processDescription Description of the task displayed on a screen.
- * @param {Function} options.taskToExecute A callback that is executed on all found packages. It receives an absolute path to a package as an argument. It can be synchronous or may return a promise.
+ * @param {Function} options.taskToExecute A callback that is executed on all found packages.
+ * It receives an absolute path to a package as an argument. It can be synchronous or may return a promise.
  * @param {AbortSignal} options.signal Signal to abort the asynchronous process.
  * @param {String} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
  * @param {Number} [options.concurrency=require( 'os' ).cpus().length / 2] Number of CPUs that will execute the task.
--- a/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
@@ -61,6 +61,11 @@ module.exports = function executeInParallel( options ) {
 		.catch( err => {
 			counter.finish( { emoji: '❌' } );
 
+			// `err` can be `undefined` if a process was aborted.
+			if ( !err ) {
+				return Promise.resolve();
+			}
+
 			return Promise.reject( err );
 		} )
 		.finally( () => {
--- a/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
@@ -8,13 +8,13 @@
 'use strict';
 
 const crypto = require( 'crypto' );
-const path = require( 'path' );
+const upath = require( 'upath' );
 const fs = require( 'fs' );
 const { Worker } = require( 'worker_threads' );
-const glob = require( 'glob' );
+const { globSync } = require( 'glob' );
 const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
 
-const WORKER_SCRIPT = path.join( __dirname, 'parallelworker.js' );
+const WORKER_SCRIPT = upath.join( __dirname, 'parallelworker.js' );
 
 /**
  * This util allows executing a specified task in parallel using Workers. It can be helpful when executing a not resource-consuming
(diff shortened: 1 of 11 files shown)


packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
```diff
@@ -0,0 +1,124 @@
...
+ * @returns {Promise}
+ */
+module.exports = function executeInParallel( options ) {
+	const {
+		packagesDirectory,
+		processDescription,
+		signal,
+		taskToExecute,
+		cwd = process.cwd(),
+		concurrency = require( 'os' ).cpus().length / 2
```
martnpaneq, line 43: What about `require( 'os' ).cpus().length - 1` to maximize speed?

pomek: Would you like to measure if it speeds up significantly?

martnpaneq: It is dependent on the processor. When I made some tests with the number of cores, using 8 cores was about 20% faster than using 4 cores.
I just wrote it to say that when more cores are available, then why not to use them?
Leaving 1 cpu unused so that the CPU is responsive seams seasonable.

pomek: Usually, we go through the release on a call. As we know, zoom takes some resources while sharing a screen. I doubt that it will be faster than when using a lower number of CPUs. But I got your point, and this option will be configurable via CLI, so an integrator can use it as much as possible.

packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * @see https://nodejs.org/api/worker_threads.html
+ * @param {Object} options
+ * @param {String} options.packagesDirectory
+ * @param {String} options.processDescription
+ * @param {Function} options.callback
+ * @param {AbortSignal} options.signal
+ * @param {String} [options.cwd=process.cwd()]
+ * @param {Number} [options.concurrency=require( 'os' ).cpus().length / 2]
+ * @returns {Promise}
```
pomek, line 34:
```suggestion
 * @param {Object} options
 * @param {String} options.packagesDirectory Relative path to a location of packages to execute a task.
 * @param {String} options.processDescription Description of the task displayed on a screen.
 * @param {Function} options.taskToExecute A callback that is executed on all found packages. It receives an absolute path to a package as an argument. It can be synchronous or may return a promise.
 * @param {AbortSignal} options.signal Signal to abort the asynchronous process.
 * @param {String} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
 * @param {Number} [options.concurrency=require( 'os' ).cpus().length / 2] Number of CPUs that will execute the task.
 * @returns {Promise}
```

psmyrek (approved):
LGTM 🎉

packages/ckeditor5-dev-release-tools/tests/utils/executeinparallel-integration.js
```diff
@@ -0,0 +1,123 @@
...
+
+			for ( const { value, packageName, source } of data ) {
+				expect( value > timeBefore, `comparing timeBefore (${ packageName })` ).to.equal( true );
+				expect( value < timeAfter, `comparing timeAfter (${ packageName })` ).to.equal( true );
+
+				fs.unlinkSync( source );
+			}
+		} );
+
+		it( 'should store current time in all found packages (callback returns a promise)', async () => {
```
pomek, line 85:
```suggestion
		it( 'should store current time in all found packages (callback is a synchronous function)', async () => {
```

packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
```diff
@@ -0,0 +1,134 @@
...
+
+	const normalizedCwd = upath.toUnix( cwd );
+	const packages = globSync( `${ packagesDirectory }/*/`, {
+		cwd: normalizedCwd,
+		absolute: true
+	} );
+	const packagesInThreads = getPackagesGroupedByThreads( packages, concurrency );
+
+	const callbackModule = upath.join( cwd, crypto.randomUUID() + '.js' );
+	fs.writeFileSync( callbackModule, `'use strict';\nmodule.exports = ${ taskToExecute };`, 'utf-8' );
```
martnpaneq, line 55: I managed to run the worker without creating additional file. If you have time you can try to check it out:
`eval: true` is the key here.

```js
createWorker( `(${ workerContent })()`, { packages: packagesChunk } )
```

```js
function createWorker( file, workerData ) {
	return new Promise( function( resolve, reject ) {
		const ckeditor5Worker = new Worker( file, { workerData, eval: true } );
		ckeditor5Worker.on( 'error', reject );
		ckeditor5Worker.on( 'message', console.log );
		ckeditor5Worker.on( 'exit', resolve );
	} );
}
```

packages/ckeditor5-dev-release-tools/tests/utils/executeinparallel.js
```diff
@@ -0,0 +1,550 @@
...
+
+			// Workers did not emit an error.
+			getExitCallback( firstWorker )( 0 );
+			getExitCallback( secondWorker )( 0 );
+
+			await promise;
+		} );
+
+		it( 'should create the temporary module properly when using Windows-style paths', async () => {
+			stubs.process.cwd.returns( 'C:\\Users\\ckeditor' );
```
martnpaneq, line 190: *minor*

I don't think we should test the OS specific paths in this place.
This way such tests should be present in all new release functions.

Repeating such tests causes the test file to grow in size when it's not needed.

pomek: This test aims to confirm if the normalized `process.cwd()` path is used when storing the temporary module. Not sure if needed. I prefer to add more than less.

martnpaneq (approved):

packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
pomek: No eval, please :D

PR #870, opened 2023-05-02
Add updateVersions() util for the release process.

martnpaneq:

Feature (release-tools):  Introduced the `updateVersions()` task for the redesigned release process for updating the version of all packages found to release. Closes ckeditor/ckeditor5#13952.

---

### Additional information


packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
```diff
@@ -0,0 +1,63 @@
...
+	if ( !semver.gt( newVersion, oldVersion ) ) {
+		throw new Error( `Version ${ newVersion } must be greater than ${ oldVersion } or match 0.0.0-<...>.` );
+	}
+}
+
+function checkVersionIsAvailableInNpm( version, packageName ) {
+	try {
+		tools.shExec( `npm show ${ packageName }@${ version } version`, { verbosity: 'silent' } );
+
+		throw new Error( `Version ${ version } is already used in npm by ${ packageName }.` );
```
psmyrek, line 57: Just to align to other messages:

```suggestion
		throw new Error( `Provided version "${ version }" is already used in npm by the "${ packageName }".` );
```

packages/ckeditor5-dev-release-tools/tests/release/updateversions.js
```diff
@@ -0,0 +1,118 @@
...
+			updateVersions( { version: '0.0.0-nightly-123.0' } );
+
+			expect( outputJsonSyncStub ).to.be.called;
+		} );
+
+		it( 'should throw when new version is not greater than old version', () => {
+			gtStub.returns( false );
+
+			expect( () => updateVersions( { version: '1.0.1' } ) )
+				.to.throw( Error, 'Version 1.0.1 must be greater than 1.0.0 or match 0.0.0-<...>' );
```
psmyrek, line 97: The version `1.0.1` is actually greater than `1.0.0` 😃

packages/ckeditor5-dev-release-tools/tests/release/updateversions.js
```diff
@@ -0,0 +1,118 @@
...
+				'/ckeditor5-dev/packages/package2',
+				'/ckeditor5-dev/packages/package3',
+				'/ckeditor5-dev'
+			] );
+
+			updateVersions( { version: '1.0.1', packagesDirectory: 'packages' } );
+
+			expect( syncStub.firstCall.args[ 0 ] ).to.deep.equal( [ 'packages/*', './' ] );
+			expect( outputJsonSyncStub.callCount ).to.equal( 4 );
+			expect( outputJsonSyncStub.firstCall.args[ 0 ] ).to.equal( '/ckeditor5-dev/packages/package1/package.json' );
```
psmyrek, line 50: Not true on Windows, where instead of slashes there are backslashes: `\ckeditor5-dev\packages\package1\package.json`.

packages/ckeditor5-dev-release-tools/tests/release/updateversions.js
```diff
@@ -0,0 +1,118 @@
...
+		} );
+
+		it( 'should call outputJsonSyncStub 1 time when the version is correct packages are undefined and 1 root', () => {
+			syncStub.returns( [ '/ckeditor5-dev' ] );
+
+			updateVersions( { version: '1.0.1' } );
+
+			expect( syncStub.firstCall.args[ 0 ] ).to.equal( './' );
+			expect( outputJsonSyncStub.callCount ).to.equal( 1 );
+			expect( outputJsonSyncStub.firstCall.args[ 0 ] ).to.equal( '/ckeditor5-dev/package.json' );
```
psmyrek, line 61: Not true on Windows, where instead of slashes there are backslashes: `\ckeditor5-dev\package.json`.

packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+'use strict';
+
+const { globSync } = require( 'glob' );
+const path = require( 'path' );
+const fs = require( 'fs-extra' );
+const semver = require( 'semver' );
+const getPackageJson = require( '../utils/getpackagejson' );
+const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
```
psmyrek, line 13: Just the order: external imports are first, then the local ones.

```suggestion
const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
const getPackageJson = require( '../utils/getpackagejson' );
```

psmyrek (commented):
It works as expected 🎉 
Found a few minor points to discuss and tests to align.

packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
```diff
@@ -0,0 +1,63 @@
...
+		fs.outputJsonSync( packageJsonPath, { ...packageJson, version }, { replacer: null, spaces: 2 } );
+	}
+};
+
+function getRandomPackageJson( packagesDirectory, cwd ) {
+	const globs = packagesDirectory ? packagesDirectory + '/*' : './';
+	const packagesPaths = globSync( globs, { cwd, absolute: true } );
+	const randomPackagePath = packagesPaths[ Math.floor( Math.random() * packagesPaths.length ) ];
+	return getPackageJson( randomPackagePath );
+}
```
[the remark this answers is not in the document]
martnpaneq: This is intentional to not read the root package, because the root package sometimes does not exist in npm (`collaboration-features` or `ckeditor5-internal` root), but their versions should be also bumped.

packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
```diff
@@ -6,55 +6,70 @@
...
+ * Updates version of root package and subpackages if `packagesDirectory` path is provided.
+ *
  * @param {Object} options
  * @param {String} options.version
  * @param {String} [options.packagesDirectory]
  * @param {String} [options.cwd]
  */
 module.exports = function updateVersions( { packagesDirectory, version, cwd = process.cwd() } ) {
-	const packagesGlobs = packagesDirectory ? [ packagesDirectory + '/*', './' ] : './';
+	const normalizedPackagesDir = normalizePath( packagesDirectory );
```
psmyrek, line 24: This line will throw an error if `packagesDirectory` is `undefined`. Your unit tests have a mocked version of `normalizePath()` so they don't crash in that case.

packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
```diff
@@ -0,0 +1,80 @@
...
+	const packagesPaths = globSync( globs, { cwd, absolute: true } );
+	const randomPackagePath = packagesPaths[ Math.floor( Math.random() * packagesPaths.length ) ];
+	return getPackageJson( randomPackagePath );
+}
+
+/**
+ * @param {String} newVersion
+ * @param {String} oldVersion
+ */
+function checkVersionGreaterThanOldVersion( newVersion, oldVersion ) {
```
pomek, line 56:
```suggestion
function checkVersionGreaterThanOldVersion( newVersion, currentVersion ) {
```

Otherwise, what does `oldVersion` mean?

packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @param {String} packagesDirectory
+ * @param {String } cwd
+ * @returns {Object}
+ */
+function getRandomPackageJson( packagesDirectory, cwd ) {
+	const globs = packagesDirectory ? packagesDirectory + '/*' : './';
+	const packagesPaths = globSync( globs, { cwd, absolute: true } );
+	const randomPackagePath = packagesPaths[ Math.floor( Math.random() * packagesPaths.length ) ];
+	return getPackageJson( randomPackagePath );
```
pomek, line 49:
```suggestion
	const randomPackagePath = packagesPaths[ Math.floor( Math.random() * packagesPaths.length ) ];

	return getPackageJson( randomPackagePath );
```

We usually add a blank line before the return statement.

packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+/**
+ * @param {String} packagesDirectory
+ * @param {String } cwd
+ * @returns {Object}
+ */
+function getRandomPackageJson( packagesDirectory, cwd ) {
+	const globs = packagesDirectory ? packagesDirectory + '/*' : './';
+	const packagesPaths = globSync( globs, { cwd, absolute: true } );
```
pomek, line 47: I believe we **should not** look for packages again.

pomek (changes requested):
I think the flow can be simplified a bit.

packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
```diff
@@ -0,0 +1,80 @@
...
+	const normCwd = toUnix( cwd );
+	const normPackagesDir = packagesDirectory ? normalizeTrim( packagesDirectory ) : packagesDirectory;
+	const packagesGlobs = normPackagesDir ? [ normPackagesDir + '/*', './' ] : './';
+	const packagesPaths = globSync( packagesGlobs, { cwd: normCwd, absolute: true } );
+
+	checkVersionGreaterThanOldVersion( version, getPackageJson( normCwd ).version );
+	checkVersionIsAvailableInNpm( version, getRandomPackageJson( normPackagesDir, normCwd ).name );
+
+	for ( const packagePath of packagesPaths ) {
+		const packageJsonPath = join( packagePath, 'package.json' );
```
psmyrek, line 33: Because `glob` returns directory paths instead of paths to `package.json` files, I guess we should make sure that `package.json` file actually exists in each found directory. Sometimes there are empty directories left between git checkouts and in that case this script will crash during the release.

packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
martnpaneq: Unfortunately this glob pattern is different than the previous one.

We don't want to search in the root for `collaboration-features` and `ckeditor5-internal`, because they don't exist on npm and the npm check would throw an error.

That's why we have to have two calls to `globSync`.

packages/ckeditor5-dev-release-tools/tests/release/updateversions.js
```diff
@@ -0,0 +1,138 @@
...
+		} );
+
+		it( 'should pass cwd to globSync', () => {
+			updateVersions( { version: '1.0.1', cwd: 'Users/username/ckeditor5-dev/custom-dir' } );
+
+			expect( syncStub.firstCall.args[ 1 ] ).to.deep.equal( {
+				cwd: 'Users/username/ckeditor5-dev/custom-dir',
+				absolute: true,
+				nodir: true
+			} );
```
pomek, line 116: This assertion should be duplicated in the first two tests. Then, this test scenario can be removed.

packages/ckeditor5-dev-release-tools/tests/release/updateversions.js
```diff
@@ -0,0 +1,138 @@
...
+			expect( readJsonSyncStub.firstCall.args[ 0 ] ).to.contain( 'package1' );
+		} );
+
+		it( 'should pass "/ckeditor5-dev" to readJsonSync to search for random packages when packagesDirectory is undefined', () => {
+			syncStub.returns( [ '/ckeditor5-dev' ] );
+
+			updateVersions( { version: '1.0.1' } );
+
+			expect( readJsonSyncStub.firstCall.args[ 0 ] ).to.equal( '/ckeditor5-dev' );
+		} );
```
pomek, line 136:
```suggestion
		it( 'should pass path with "package1" to readJsonSync to search for random packages when packagesDirectory is defined', () => {
			syncStub.returns( [
				'/ckeditor5-dev/packages/package1',
				'/ckeditor5-dev'
			] );

			updateVersions( { version: '1.0.1', packagesDirectory: 'packages' } );

			expect( readJsonSyncStub.firstCall.args[ 0 ] ).to.contain( 'package1' );
		} );

		it( 'should pass "/ckeditor5-dev" to readJsonSync to search for random packages when packagesDirectory is undefined', () => {
			syncStub.returns( [ '/ckeditor5-dev' ] );

			updateVersions( { version: '1.0.1' } );

			expect( readJsonSyncStub.firstCall.args[ 0 ] ).to.equal( '/ckeditor5-dev' );
		} );
```

Both tests should be merged with the proposal:

```
it( 'should not throw when the version is available (not in use)', () => {
```

The first one for the root package, the second one for the packages found in a packages directory.

packages/ckeditor5-dev-release-tools/lib/release/updateversions.js
```diff
@@ -0,0 +1,90 @@
...
+	if ( !semver.valid( newVersion ) ) {
+		throw new Error( `Provided version ${ newVersion } must be a valid semver version.` );
+	}
+
+	if ( !semver.gt( newVersion, currentVersion ) ) {
+		throw new Error( `Provided version ${ newVersion } must be greater than ${ currentVersion }.` );
+	}
+}
+
+/**
```
pomek, line 76: One week later, I am trying to remember the purpose of the npm check. It deserves a comment about why we do it.

This check aims to verify if the proposed version is available. In other words, check if the package can be published under the given version.

pomek (changes requested):
Please, pay even more attention to writing _real scenario_ tests. Even with stubs/mocks we can create an environment that behaves like our daily workspace.

I suggest calling tomorrow to discuss and improve the reported points.

packages/ckeditor5-dev-release-tools/tests/release/updateversions.js
martnpaneq: I accepted almost all of the comments, but this will be an exception :)
Adding this test logic to test 1 and 2 would result in:
1. Duplicating test code.
2. Cluttering the test with unnecessary logic.

Adding a new test just for cwd looks cleaner :)

packages/ckeditor5-dev-release-tools/tests/tasks/updateversions.js
```diff
@@ -0,0 +1,143 @@
...
+			stubs.getPackageJson.returns( { version: '1.0.0', name: 'stub-package' } );
+
+			expect( () => updateVersions( { version: '1.0.1' } ) ).to.throw( Error, 'custom error' );
+		} );
+
+		it( 'should accept a "0.0.0-nightly-123*" version for nightly releases', () => {
+			updateVersions( { version: '0.0.0-nightly-123.0' } );
+
+			expect( stubs.outputJsonSync ).to.be.called;
+		} );
```
pomek, line 96: [As I mentioned in the previous review](https://github.com/ckeditor/ckeditor5-dev/pull/870#discussion_r1187781952), it won't work, and this test checks nothing.

I integrated the script and passed `0.0.0-nightly-20220509.0` as the version value.

```
Error: Provided version 0.0.0-nightly-20220509.0 must be greater than 37.1.0.
    at checkVersionGreaterThanOldVersion (/Users/pomek/Projects/ckeditor/ckeditor5/external/ckeditor5-dev/packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js:72:9)
    at Object.updateVersions (/Users/pomek/Projects/ckeditor/ckeditor5/external/ckeditor5-dev/packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js:37:2)
    at /Users/pomek/Projects/ckeditor/ckeditor5/scripts/release/prepare-packages.js:44:15
```

PR #871, opened 2023-05-02
Added script that updates packages dependencies to a target version

psmyrek:

Feature (release-tools): Added a script that updates dependencies to a target version. Closes ckeditor/ckeditor5#13953.

---

### Additional information

Related PR in CKEditor 5 repository that checks if a package is CKEditor 5 dependency: https://github.com/ckeditor/ckeditor5/pull/14016.

Notes:
- <s>Introduced new folder `lib/tasks-new` that temporarily contains new things related to the new release process not to mix them up with the old ones.</s>
- Bumped `glob` to v10+.


packages/ckeditor5-dev-release-tools/tests/tasks-new/updatedependencies.js
```diff
@@ -0,0 +1,315 @@
...
+			it( 'should search for packages only in `cwd` if `packagesDirectory` option is not provided', () => {
+				updateDependencies( {} );
+
+				expect( stubs.glob.globSync.calledOnce ).to.equal( true );
+				expect( stubs.glob.globSync.getCall( 0 ).args[ 0 ] ).to.deep.equal( [
+					'package.json'
+				] );
+			} );
+
+			it( 'should remove leading and trailing path separators from the `packagesDirectory`', () => {
```
martnpaneq, line 116: Looks like tests for removing trailing path separators should be added to a util and not this function.
Su[ch util should be probably used by most of the tasks in the release process that accept `packagesDirectory` as argument.

packages/ckeditor5-dev-release-tools/lib/tasks/updatedependencies.js
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * - The root package located in `options.cwd` path.
+ * - All packages located in the `options.packagesDirectory` path relative to `options.cwd`.
+ *
+ * The eligible dependencies are distinguished by the return value from the `options.shouldUpdateVersionCallback` function. Only if this
+ * callback returns a truthy value for a given dependency, its version will be updated.
+ *
+ * @param {Object} options
+ * @param {String} options.version Target version to which all of the eligible dependencies will be updated.
+ * @param {String} options.shouldUpdateVersionCallback Callback function that decides whether to update a version for a dependency.
```
martnpaneq, line 24: I know that we don't wan't to leak any private packages names, but I have a feeling that this callback should be optional.

pomek: Assuming so, how would you like to determine if a package should be updated? I do not see any other solution that allows us to use tooling outside the CKEditor 5 project. 

E.g., the package generator. We do not want to update CKEditor 5 feature packages while releasing the tool. We only want to update tools and the generator itself.

martnpaneq: We talked about it and it's not as easy as I imagined. So not using default value is okay for me :)

martnpaneq (approved):

packages/ckeditor5-dev-release-tools/tests/tasks-new/updatedependencies.js
psmyrek: Extracted the `normalizePath()` to `lib/utils/normalizepath.js`, but I would like to keep two tests here to make sure that the path from `packagesDirectory` will be normalized.

martnpaneq (approved):

packages/ckeditor5-dev-release-tools/lib/tasks/updatedependencies.js
```diff
@@ -0,0 +1,87 @@
...
+};
+
+/**
+ * Updates the version for each eligible dependency.
+ *
+ * @param {Object} object Object containing dependencies to update, where the key is a package name and the value is its version.
+ * @param {String} version Target version to which all of the eligible dependencies will be updated.
+ * @param {Function} callback Callback function that decides whether to update a version for a dependency.
+ */
+function updateVersion( object, version, callback ) {
```
pomek, line 77:
```suggestion
function updateVersion( version, callback, dependencies ) {
```

As `object` can be undefined, I would swap the order of arguments.

packages/ckeditor5-dev-release-tools/lib/tasks/updatedependencies.js
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * The eligible dependencies are distinguished by the return value from the `options.shouldUpdateVersionCallback` function. Only if this
+ * callback returns a truthy value for a given dependency, its version will be updated.
+ *
+ * @param {Object} options
+ * @param {String} options.version Target version to which all of the eligible dependencies will be updated.
+ * @param {String} options.shouldUpdateVersionCallback Callback function that decides whether to update a version for a dependency.
+ * @param {String} [options.packagesDirectory] Relative path to a location of packages to update their dependencies.
+ * @param {String} [options.cwd] Current working directory from which all paths will be resolved.
+ */
```
pomek, line 28:
```suggestion
/**
 * The purpose of this script is to update all eligible dependencies to a version specified in the `options.version`. The following packages
 * are taken into consideration:
 *
 * - The root package located in `options.cwd` path.
 * - All packages located in the `options.packagesDirectory` path relative to `options.cwd`.
 *
 * The eligible dependencies are distinguished by the return value from the `options.shouldUpdateVersionCallback` function. Only if this
 * callback returns a truthy value for a given dependency, its version will be updated.
 *
 * @param {Object} options
 * @param {String} options.version Target version or a range version to which all eligible dependencies will be updated, e.g., `1.0.0`, `^1.0.0`, etc.
 * @param {Function} options.shouldUpdateVersionCallback Callback function that decides whether to update a version for a dependency. It receives a package name as an argument and should return a boolean value.
 * @param {String} [options.packagesDirectory] Relative path to a location of packages to update their dependencies. If not specified, only the root package is checked.
 * @param {String} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
 */
```

packages/ckeditor5-dev-release-tools/tests/tasks/updatedependencies.js
```diff
@@ -0,0 +1,315 @@
...
+				},
+				process: {
+					cwd: sandbox.stub( process, 'cwd' ).returns( '/work/project' )
+				},
+				devUtils: {
+					logger: sandbox.stub().returns( {
+						error: sandbox.stub(),
+						warning: sandbox.stub(),
+						info: sandbox.stub()
+					} )
```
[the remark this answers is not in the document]
psmyrek: True. I even consider not testing the messages printed in the console at all...

przemyslaw-zan (approved):
WFM & LGTM

pomek (approved):
LGTM.

PR #875, opened 2023-05-10
Remove non-published files and clean up `package.json`

psmyrek:

Feature (release-tools): Added script that removes unnecessary files and directories from all packages prepared for the release. Additionally, it deletes unnecessary fields from the `package.json` file. Closes ckeditor/ckeditor5#13955.

---

### Additional information

PR with integration with the new release process: https://github.com/ckeditor/ckeditor5/pull/14085


packages/ckeditor5-dev-release-tools/lib/tasks/cleanuppackages.js
```diff
@@ -0,0 +1,175 @@
...
+
+	if ( firstPathSegments > secondPathSegments ) {
+		return -1;
+	}
+
+	if ( firstPathSegments < secondPathSegments ) {
+		return 1;
+	}
+
+	return 0;
```
przemyslaw-zan, line 174: This can be simpler:

```suggestion
	return secondPathSegments - firstPathSegments;
```

Also, removing the sorting function alltogether does not fail any unit tests, so a test covering the use case should be added.

przemyslaw-zan (changes requested):
Small nitpick.

pomek:
Could you make sure to remove dot files too?

image

packages/ckeditor5-dev-release-tools/lib/tasks/cleanuppackages.js
psmyrek: You are right, thanks for finding this. It turns out that in the tests I only checked the existence of the required files or directories. I did not verify the absence of unnecessary files or directories.

psmyrek:
Added support for dot files and dot directories, which were skipped by default by `glob`.

przemyslaw-zan (approved):
LGTM

packages/ckeditor5-dev-release-tools/lib/tasks/cleanuppackages.js
```diff
@@ -0,0 +1,173 @@
...
+
+const fs = require( 'fs-extra' );
+const upath = require( 'upath' );
+const { globSync } = require( 'glob' );
+const { logger } = require( '@ckeditor/ckeditor5-dev-utils' );
+
+/**
+ * The purpose of the script is to clean all packages prepared for the release. The cleaning consists of two stages:
+ *
+ * - Remove unnecessary files and directories from the package directory.
```
pomek, line 16: It is worth mentioning that it filters out files/directories not included in the `files` property in pkg.json.

packages/ckeditor5-dev-release-tools/lib/tasks/cleanuppackages.js
pomek, line 1: Could we add a generic spinner once we start this task? I think it takes some time, so it would be good to draw something on the screen.

pomek (commented):
I proposed a few minor changes.

packages/ckeditor5-dev-release-tools/lib/tasks/cleanuppackages.js
```diff
@@ -0,0 +1,173 @@
...
+
+	log.info( 'Task: cleanUpPackages()' );
+
+	const { packagesDirectory, packageJsonFieldsToRemove, cwd } = parseOptions( options );
+
+	const packageJsonPaths = globSync( packagesDirectory + '/*/package.json', {
+		cwd,
+		nodir: true,
+		absolute: true
+	} );
```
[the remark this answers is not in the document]
psmyrek: It doesn't matter to me, but if you suggest it, it does for you 😄
Changed as suggested.

packages/ckeditor5-dev-release-tools/lib/tasks/cleanuppackages.js
psmyrek: Added more details.

packages/ckeditor5-dev-release-tools/lib/tasks/cleanuppackages.js
psmyrek: This will not work now, because the script is synchronous.

Anyway, scanning and deleting files for 60+ packages takes about 1-2 seconds on my PC, so I'm not sure if it's worth rewriting the script to an asynchronous one and adding a logger...

packages/ckeditor5-dev-release-tools/tests/tasks/cleanuppackages.js
```diff
@@ -0,0 +1,686 @@
...
+
+		beforeEach( () => {
+			sandbox = sinon.createSandbox();
+
+			// Calls to `fs` and `glob` are stubbed, but they are passed through to the real implementation, because we want to test the
+			// real behavior of the script. The whole filesystem is mocked by the `mock-fs` util for testing purposes. A virtual project is
+			// prepared in tests on this mocked filesystem.
+			stubs = {
+				fs: {
+					readJsonSync: sandbox.stub().callsFake( ( ...args ) => fs.readJsonSync( ...args ) ),
```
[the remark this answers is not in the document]
psmyrek: This will not work, because `sinon` does not know what is stubbed (no arguments in `stub()` method) and it says: `Unable to find wrapped method`.

Also I tried with `const readJsonSyncStub = sandbox.stub( fs, 'readJsonSync' ).callThrough()` but then `readJsonSyncStub.callCount` is always 0. It seems that the imported `fs` lib in tests is another instance than the one from the source code.

The only solution that works for me is to use `mockery` and explicitly calling the real methods from the stubs.

packages/ckeditor5-dev-release-tools/lib/tasks/cleanuppackages.js
pomek: Spinner is a synchronous code so it should work.

martnpaneq (approved):

martnpaneq (approved):

pomek (approved):
LGTM.

PR #880, opened 2023-05-16
Added karma notifications

przemyslaw-zan:
Feature (tests): Added karma notification for when the tests are completed. Notification can be enabled with `-n`/`--notify` CLI option. Its enabled by default when running all tests, without the `-f`/`--files` option. Closes ckeditor/ckeditor5#13736.

---

### Additional information

(This happened on **Windows 10**)

I've encountered a small issue where the first attempt to use the notifier did not work, and only following attempts worked successfully. It happened when I've started with hard-coded value: `reporters: [ 'progress', 'notify' ]`, so it might not be an issue.

Reviewers should make sure that the current setup works on first try.


arkflpc:
Works on Linux under KDE.

psmyrek (commented):
Notification works in Windows 11:

Zrzut ekranu 2023-05-17 134720

The not cool thing is that it adds a shortcut in the Start menu...

Zrzut ekranu 2023-05-17 134842

obraz

Could you please check if this could be prevented somehow?

psmyrek:
Possible workaround:
```js
const notifier = require( 'node-notifier' );
const isWindowsNotifier = notifier.Notification === notifier.WindowsToaster;

if ( isWindowsNotifier ) {
	const originalNotify = notifier.notify;

	notifier._notify = notifierOptions => originalNotify( {
		...notifierOptions,
		appID: 'karma-notify-reporter'
	} );
}
```

przemyslaw-zan:
☝️ To avoid workarounds, we've decided to implement a custom notifier instead.

przemyslaw-zan:
Right now notifications use default icon. Do we want to use custom icons, and if so, where do we take them from? Should we have different icons for success and failure, or perhaps simple CKE5 logo for all notifications would be sufficient?

image

pomek (approved):
LGTM.

PR #881, opened 2023-05-17
Added script to validate and publish packages to npm for the new release process

psmyrek:

Feature (release-tools): Added script to validate and publish packages to npm. Closes ckeditor/ckeditor5#13957.

---

### Additional information


packages/ckeditor5-dev-release-tools/lib/tasks/publishpackages.js
```diff
@@ -0,0 +1,67 @@
...
+ * @param {String} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
+ * @returns {Promise}
+ */
+module.exports = async function publishPackages( options ) {
+	const {
+		packagesDirectory,
+		npmOwner,
+		npmTag = 'staging',
+		optionalEntries = null,
+		confirmationCallback = null,
```
martnpaneq, line 47: I have a question about `confirmationCallback`. If I understand correctly, it is supposed to work as a dry run? Or do we have other use case for it?

martnpaneq (approved):
I really like that utils are separated from the `publishPackages` function. It's very easy to read :)

packages/ckeditor5-dev-release-tools/lib/tasks/publishpackages.js
psmyrek: The purpose of this callback is to give user the final opportunity to confirm or cancel the publishing on npm.

For manually invoked scripts, I would expect a prompt here asking the user whether to continue (and that there will be no turning back after confirmation). For a script invoked on CI that should publish nightly versions, the callback will probably always return `true`.

pomek (approved):
LGTM.

PR #885, opened 2023-05-23
Don't require file extension in import paths

filipsobol:

Other (tests): A file extension is not required when resolving modules within a package marked as `type: module`. See ckeditor/ckeditor5#13673.

---

### Additional information

Adding `type:module` to packages as described in https://github.com/ckeditor/ckeditor5/issues/13673 changes the way webpack works. With this value added, it requires import paths to include file extensions. Unfortunately, this would be a massive change to our codebase, so to work around this, I suggest using [`resolve.fullySpecified`](https://webpack.js.org/configuration/module/#resolvefullyspecified) to disable this behavior from automated and manual tests.


arkflpc (approved):

pomek (approved):
LGTM.

Changes do not break existing scripts. I checked building docs and running tests (+with CC).

PR #886, opened 2023-05-26
Added support for running manual and automated tests written in TS.

martnpaneq:

Other (tests): Support for running manual and automated tests written in TypeScript. Remember that test runners do not validate types as we use `esbuild` which ignores it. Closes ckeditor/ckeditor5#14170, ckeditor/ckeditor5#14171.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
+++ b/packages/ckeditor5-dev-tests/lib/tasks/runautomatedtests.js
@@ -20,9 +20,9 @@ const transformFileOptionToTestGlob = require( '../utils/transformfileoptiontote
 // that matches to these patterns, the file will be skipped.
 const IGNORE_GLOBS = [
 	// Ignore files which are saved in `manual/` directory. There are manual tests.
-	path.join( '**', 'tests', '**', 'manual', '**', '*.js' ),
+	path.join( '**', 'tests', '**', 'manual', '**', '*.{js,ts}' ),
 	// Ignore `_utils` directory as well because there are saved utils for tests.
-	path.join( '**', 'tests', '**', '_utils', '**', '*.js' )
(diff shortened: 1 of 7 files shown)


psmyrek (commented):
Manual and automated tests in TypeScript work fine 🎉 

Remarks:
1. I found that the build directory for manual tests contains `.ts` files, which are not needed there (they are not imported by `.html` files):

   

   Would it be possible to not copy TypeScript files?

2. In `lib/tasks/runautomatedtests.js` there is `IGNORE_GLOBS` that also should skip TypeScript files.

3. I think that a few sentences in the CKE5 docs (contributing section) about the fact that manual and automated tests can be written now in both JavaScript and TypeScript would be useful.

martnpaneq:
> Manual and automated tests in TypeScript work fine 🎉
> 
> Remarks:
> 
> 1. I found that the build directory for manual tests contains `.ts` files, which are not needed there (they are not imported by `.html` files):
>    
>    Would it be possible to not copy TypeScript files?
> 2. In `lib/tasks/runautomatedtests.js` there is `IGNORE_GLOBS` that also should skip TypeScript files.
> 3. I think that a few sentences in the CKE5 docs (contributing section) about the fact that manual and automated tests can be written now in both JavaScript and TypeScript would be useful.

@psmyrek 
1. Fixed.
2. Good catch. I did not test running all tests and without this change the tests would fail.
3. I added a PR for `ckeditor5` with some info about TS support in the `testing-environment` section. I'm not sure if it's the best place, so if you have a better idea please let me know https://github.com/ckeditor/ckeditor5/pull/14285

psmyrek (approved):
LGTM 🎉

pomek (approved):
LGTM.

It works as expected, tests are compiled, but we lost type checking due to esbuild. It's fine when prototyping things, tho. I feel that at some point, we will bring back `ts-loader` behind some flag.

PR #887, opened 2023-05-29
Print raw strings if formatted markup is equal

psmyrek:

Fix (tests): Fixed `equalMarkup` assertion to print raw (unformatted) strings if formatted markup is equal. Closes ckeditor/ckeditor5#14175.

---

### Additional information

HTML beautification tools automatically remove extra whitespace characters inside tags (e.g. between attributes, after `<`, before `>`, etc.). It seems that this behavior cannot be adjusted. Apart from the `js-beautify` which we currently use, I checked also `diffable-html` (another popular tool for beautifying HTML) and there is also no control over whitespace inside tags.

For this reason, in case the formatted strings are the same, let's just display the raw (unformatted) strings.


pomek (approved):
LGTM.

PR #888, opened 2023-05-30
Treat warnings as errors in the "production" mode

martnpaneq:

Fix (tests): When running automated tests in the "production" mode, warnings will be treated as errors. It allows the detection of webpack messages that may impact a build but does not break the process.

Fix (tests): The manual and automated test environments do not report a warning regarding the missing `timers` module.

---

Closes cksource/ckeditor5-internal#3299.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-tests/bin/testautomated.js
+++ b/packages/ckeditor5-dev-tests/bin/testautomated.js
@@ -13,6 +13,14 @@ const tests = require( '../lib/index' );
 
 const options = tests.parseArguments( process.argv.slice( 2 ) );
 
+// We want to mark tests as failed if any warnings are logged. See https://github.com/cksource/ckeditor5-internal/issues/3299
+if ( options.production ) {
+	console.warn = errors => {
+		console.log( errors );
+		throw new Error( 'Warnings detected - terminating tests.' );
+	};
+}
+
 if ( options.files.length === 0 ) {
 	if ( options.cwd.endsWith( 'ckeditor5' ) ) {
 		options.files = [ '*' ];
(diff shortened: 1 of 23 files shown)


packages/ckeditor5-dev-tests/bin/testautomated.js
```diff
@@ -13,6 +13,14 @@ const tests = require( '../lib/index' );
 
 const options = tests.parseArguments( process.argv.slice( 2 ) );
 
+// We want to mark tests as failed if any warnings are logged. See https://github.com/cksource/ckeditor5-internal/issues/3299
+if ( options.production ) {
+	console.warn = errors => {
+		console.log( errors );
```
martnpaneq, line 19: I tried calling `console.error` instead, but it does not break the execution.
The question is should we also throw error on `console.error`?

psmyrek: I think that both `console.warn` and `console.error` should break the execution and throw an error.

pomek: During a call, we created a custom plugin. Could you guys verify if it works on your machines too?

Webpack config for automated tests:

```js
if ( options.production ) {
	config.plugins.push(
		new class TreatWarningsAsErrorsWebpackPlugin {
			apply( compiler ) {
				compiler.hooks.shouldEmit.tap( 'TreatWarningsAsErrorsWebpackPlugin', compilation => {
					const stats = compilation.getStats();

					compilation.errors = [
						...compilation.errors,
						...compilation.warnings
					];

					compilation.warnings = [];

					if ( compilation.errors.length ) {
						return false;
					}
				} );
			}
		}
	)
}
```

psmyrek: It works. Test run is terminated.

pomek:
The same error occurs for manual tests.

image

pomek:
I checked if the polyfill is needed at all. It turns out that everything works as expected without it.

PR #891, opened 2023-06-05
getDllPluginWebpackConfig() loads JS entry point if exists

pomek:

Fix (utils): When preparing a configuration for a DLL build, a JavaScript file should take precedence over the TypeScript version as it might differ from the source code. Closes ckeditor/ckeditor5#14335.


psmyrek (approved):
LGTM 🎉

PR #894, opened 2023-06-14
Release tools: Added support for publishing packages to npm in parallel

psmyrek:

Other (release-tools): Added support for publishing packages to npm in parallel. By default, 4 packages are published simultaneously. Closes ckeditor/ckeditor5#14333.

Other (release-tools): The `AbortSignal` parameter is no longer required for the `executeInParallel()` util. If `AbortSignal` is not passed, a default `AbortController` will be created automatically.

---

### Additional information

There are two main changes introduced by this PR:
- From now on publishing packages to npm uses `executeInParallel()` util with default concurrency set to 4. According to the npm support, this should be safe.
- Also, I get rid of the `AbortSignal` from the required arguments from the `executeInParallel()` util. If `AbortSignal` is not passed, a default `AbortController` will be created automatically. `AbortSignal` is needed to terminate asynchronous scripts executed in threads by the `executeInParallel()` util.

PRs in 3 other repositories that switch from custom `AbortController` to the default one are also created and linked with this one.


martnpaneq (approved):
It works as expected :)

How about removing the `abortController` that we had to add to the `ckeditor5` and `collaboration-features` release process? Maybe it would be a good idea to remove it right now, so that we don't forget about it it the future?

psmyrek:
PRs that switch from custom `AbortController` to the default one are created.

pomek (approved):
LGTM.

PR #906, opened 2023-08-01
Extracted `checkVersionAvailability()` in release tools to a separate util

psmyrek:

Other (release-tools): Extracted `checkVersionAvailability()` function in release tools to a separate util that checks if provided version for a package already exists in npm.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/utils/checkversionavailability.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/checkversionavailability.js
@@ -1,0 +1,27 @@
+/**
+ * @license Copyright (c) 2003-2023, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+'use strict';
+
+const { tools } = require( '@ckeditor/ckeditor5-dev-utils' );
+
+/**
+ * Checks if the provided version is not used in npm and there will be no errors when calling publish.
+ *
+ * @param {String} version
+ * @param {String} packageName
+ * @returns {Promise}
+ */
+module.exports = async function checkVersionAvailability( version, packageName ) {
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-dev-release-tools/lib/utils/checkversionavailability.js
```diff
@@ -0,0 +1,35 @@
...
+				return;
+			}
+
+			throw new Error( `The "${ packageName }@${ version }" already exists in npm.` );
+		} )
+		.catch( error => {
+			if ( !error.toString().includes( 'npm ERR! code E404' ) ) {
+				throw error;
+			}
+		} );
```
pomek, line 34: TBH, I do not understand why we didn't simplify the entire function for checking if a given version is available.

```suggestion
	return tools.shExec( `npm show ${ packageName }@${ version } version`, { verbosity: 'silent', async: true } )
		.then( result => {
			// Explicit check for npm < 8.13.0, which does not return anything and it exits with a zero status code when the version for the
			// provided package does not exist in the npm registry.
			if ( !result ) {
				// Is it available? Yes.
				return true;
			}

			// Is it available? No.
			return false;
		} )
		.catch( error => {
			if ( !error.toString().includes( 'npm ERR! code E404' ) ) {
				throw error;
			}
			
			// Npm 8.13+ throws if a package does not exist, so:
			// Is it available? Yes.
			return true;	
		} );
```

I believe it would require more changes so feel free to reject the idea if you do not agree.

psmyrek: Yeah, since we are making this util public, it also makes sense to improve it a bit. I will prepare the changes as you proposed, thanks.

pomek (approved):
LGTM.

PR #912, opened 2023-08-14
Refactored `reassignNpmTags()` to be called asynchronously to improve…

martnpaneq:
… performance.

Other (release-tools): The `reassignNpmTags()` task processes all packages asynchronously to improve the performance results. Closes ckeditor/ckeditor5#14769.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/tasks/reassignnpmtags.js
+++ b/packages/ckeditor5-dev-release-tools/lib/tasks/reassignnpmtags.js
@@ -16,6 +16,7 @@ const assertNpmAuthorization = require( '../utils/assertnpmauthorization' );
 
 /**
  * Used to switch the tags from `staging` to `latest` for specified array of packages.
+ * Each operation will be retried up to 3 times in case of failure.
  *
  * @param {Object} options
  * @param {String} options.npmOwner User that is authorized to release packages.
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-dev-release-tools/lib/tasks/reassignnpmtags.js
```diff
@@ -43,10 +45,11 @@ module.exports = async function reassignNpmTags( { npmOwner, version, packages }
 		if ( latestVersion === version ) {
```
martnpaneq, line 45: In my opinion `npm show` call is not needed. It increases the execution time, and decreases the code readability.

martnpaneq: After discussing this topic on devops daily, we decided to remove `npm show` call, because:
1. This logic is built in in the second call: `npm dist-tag add`. If a tag already exists it is not updated.
2. Using both `npm show` and `npm dist-tag add` increases the execution time. (2 calls instead of 1).
3. Error handling could be confusing with 2 calls. We would not be sure if the displayed error refers to `npm show` or `npm dist-tag add`.

pomek (approved):
LGTM.

```bash
$ time node scripts/release/switchlatest.js

real    0m2.412s
user    0m24.812s
sys     0m4.074s
```

Instead of ~5 min. Good job, good results.

PR #914, opened 2023-08-18
Introduced new scripts to configure CI.

pomek:

Other (ci): Introduced the `ckeditor5-dev-ci-allocate-swap-memory` binary script for allocating the SWAP memory to avoid issues with running out of RAM.

Other (ci): Introduced the `ckeditor5-dev-ci-install-latest-chrome` binary script for installing the latest Chrome browser.

---

### Additional information


pomek:
Self-merge (without a review) because these scripts were already reviewed in another repository. They have just moved from place A to B.

PR #916, opened 2023-08-28
Use process.env.CI when defining browsers available on CI

pomek:

Internal: Use `process.env.CI` to determine if a process is executed on CI when preparing a list of available browsers.

---

### Additional information


psmyrek (approved):
LGTM 🎉

PR #920, opened 2023-09-13
Support for ignoring some packages in the release process

pomek:

Feature (release-tools): The `updateVersions` and `updateDependencies` tasks accept a new option called `packagesDirectoryFilter`. It is a callback allowing filtering out directories/packages that the task should not touch. It receives an absolute path to a `package.json` and should return a boolean value.

---

### Additional information

See cksource/ckeditor5-commercial#5501.

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js
+++ b/packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js
@@ -24,24 +24,30 @@ const checkVersionAvailability = require( '../utils/checkversionavailability' );
  *
  * @param {Object} options
  * @param {String} options.version Version to store in a `package.json` file under the `version` key.
+ * @param {UpdateVersionsPackagesDirectoryFilter|null} [options.packagesDirectoryFilter=null] An optional callback allowing filtering out
+ * directories/packages that should not be touched by the task.
  * @param {String} [options.packagesDirectory] Relative path to a location of packages to update. If not specified,
  * only the root package is checked.
  * @param {String} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
  * @returns {Promise}
  */
-module.exports = async function updateVersions( { packagesDirectory, version, cwd = process.cwd() } ) {
+module.exports = async function updateVersions( options ) {
+	const {
+		packagesDirectory,
+		version,
+		packagesDirectoryFilter = null,
+		cwd = process.cwd()
+	} = options;
 	const normalizedCwd = toUnix( cwd );
 	const normalizedPackagesDir = packagesDirectory ? normalizeTrim( packagesDirectory ) : null;
 
 	const globPatterns = getGlobPatterns( normalizedPackagesDir );
-	const pkgJsonPaths = await glob( globPatterns, { cwd: normalizedCwd, absolute: true, nodir: true } );
+	const pkgJsonPaths = await getPackageJsonPaths( cwd, globPatterns, packagesDirectoryFilter );
+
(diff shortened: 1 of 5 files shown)


packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js
```diff
@@ -59,6 +65,36 @@ module.exports = async function updateVersions( { packagesDirectory, version, cw
...
+	if ( !packagesDirectoryFilter ) {
+		return pkgJsonPaths;
+	}
+
+	return pkgJsonPaths.filter( packagesDirectoryFilter );
+}
+
+/**
+ * @param {String} packagesDirectory
+ * @returns {Object}
```
pomek, line 90:
```suggestion
 * @returns {Promise.<Object>}
```

psmyrek (approved):
LGTM 🎉

packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js
```diff
@@ -59,6 +65,36 @@ module.exports = async function updateVersions( { packagesDirectory, version, cw
 	}
 };
 
+/**
+ * @param {String} cwd
+ * @param {Array.<String>} globPatterns
+ * @param {UpdateVersionsPackagesDirectoryFilter|null} packagesDirectoryFilter
+ * @returns {Promise.<Array.<String>>}
+ */
+async function getPackageJsonPaths( cwd, globPatterns, packagesDirectoryFilter ) {
```
martnpaneq, line 74: I see this function is copied in here:
https://github.com/ckeditor/ckeditor5-dev/pull/920/files#diff-e8b7e66eee579967e36ebc8f45594ff4b8eeacb021f4a8397cb8a02716aaef56R89

Maybe it's a good idea to extract a util?

packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js
```diff
@@ -59,6 +65,36 @@ module.exports = async function updateVersions( { packagesDirectory, version, cw
...
+	const pkgJsonPaths = await glob( globPatterns, {
+		cwd,
+		absolute: true,
+		nodir: true
+	} );
+
+	if ( !packagesDirectoryFilter ) {
```
martnpaneq, line 81: *minor*

this if could be removed if the default value for `packagesDirectoryFilter` is `(val) => val`.

packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js
pomek: I prefer creating a new util when a thing is repeated thrice and more. So far, I would KISS.

packages/ckeditor5-dev-release-tools/lib/tasks/updateversions.js
pomek: The question is whether we accept no value passed as a callback. The `packagesDirectoryFilter ` variable is always passed, so it overrides the default value, always.

pomek:
```
globalThis.foo = ( text, displayFn = console.log ) => {
    displayFn( text )
}

foo('Test') // OK
foo('Test', null) // displayFn is not a function
```

martnpaneq: Let't not make any changes.

martnpaneq (approved):

PR #923, opened 2023-09-26
Do not crash when creating inherited events when an interface extends a type

psmyrek:

Fix (typedoc-plugins): Fixed `typedoc-plugin-event-inheritance-fixer` plugin to not crash when creating inherited events if an interface extends a type. Closes ckeditor/ckeditor5#15063.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,9 @@
     "mocha": "^7.1.2",
     "nyc": "^15.1.0"
   },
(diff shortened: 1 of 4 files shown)


pomek (approved):
LGTM.

`yarn docs` does not crash anymore.

Unfortunately, CI results sound weird.

PR #926, opened 2023-10-17
Release tools - an entry point validator

pomek:

Feature (release-tools): The `publishPackages()` task accepts new options to ensure that packages define an entry point before publishing a new release. Closes ckeditor/ckeditor5#15127.

By default, the validator is disabled, but it can be changed by passing the `requireEntryPoint: true` value in the options object. Additionally, if any package does not define an entry point, the validator can skip such package by specifying its full name in the `optionalEntryPointPackages` array.

---

### Additional information


psmyrek (approved):
LGTM 🎉

PR #928, opened 2023-11-07
Typedoc linker does not emit a warning when linking to type's children

pomek:

Fix (docs): Typedoc linker validator should not emit a warning when a link points to a property defined as children of the `type` declaration. Closes ckeditor/ckeditor5#15321.

---

### Additional information


psmyrek (approved):
LGTM 🎉

PR #933, opened 2023-12-12
Custom "stale" bot

psmyrek:

Feature (stale-bot): Introduced a "stale" bot, which can be executed using the `yarn run ckeditor5-dev-stale-bot [<args>...]` command. It searches for issues and pull requests that are not active enough and then mark them as stale. Stale issues and pull requests are closed after a defined period of time or, if they become active again, they are unstaled by the bot. See ckeditor/ckeditor5#15399.

Other (utils):  Added support in the `Logger#error()` method to print in console an error instance, if provided as the second argument.

---

### Additional information


packages/ckeditor5-dev-stale-bot/bin/stale-bot.js
```diff
@@ -0,0 +1,277 @@
...
+
+	validateConfig( config );
+
+	printWelcomeMessage( dryRun );
+
+	const githubRepository = new GitHubRepository( config.GITHUB_TOKEN );
+
+	const viewerLogin = await githubRepository.getViewerLogin();
+
+	const options = parseConfig( viewerLogin, config );
```
pomek, line 46:
```suggestion
	const githubRepository = new GitHubRepository( config.GITHUB_TOKEN );
	const viewerLogin = await githubRepository.getViewerLogin();
	const options = parseConfig( viewerLogin, config );
```

packages/ckeditor5-dev-stale-bot/bin/stale-bot.js
```diff
@@ -0,0 +1,277 @@
...
+
+	const spinner = createSpinner();
+	spinner.instance.start();
+
+	const searchResult = await search( githubRepository, options, spinner );
+	const {
+		issuesOrPullRequestsToStale,
+		issuesOrPullRequestsToClose,
+		issuesOrPullRequestsToUnstale
+	} = searchResult;
```
pomek, line 56:
```suggestion
	const {
		issuesOrPullRequestsToStale,
		issuesOrPullRequestsToClose,
		issuesOrPullRequestsToUnstale
	} = await search( githubRepository, options, spinner );
```

pomek (approved):
LGTM.

I proposed a few cosmetic updates that do not have to be added.

packages/ckeditor5-dev-stale-bot/bin/stale-bot.js
psmyrek: This will not work as expected, because `searchResult` constant is also used at the end of `main()` function.

PR #939, opened 2024-02-09
Rollup CSS plugin implementation.

pszczesniak:

Internal: `Rollup` plugin for managing `CSS` bundling. Closes https://github.com/ckeditor/ckeditor5/issues/15711.

---

### Additional information


packages/ckeditor5-dev-build-tools/src/plugins/cssStyles.ts
```diff
@@ -0,0 +1,305 @@
...
+import { createFilter } from '@rollup/pluginutils';
+import { parse, type Rule, type Declaration, type Stylesheet } from 'css';
+import type { Plugin, GetModuleInfo, OutputBundle, OutputChunk, NormalizedOutputOptions } from 'rollup';
+
+/**
+ * Filter files only with `css` extension.
+ */
+const filter = createFilter( [ '**/*.css' ] );
+
+export function cssStyles(): Plugin {
```
filipsobol, line 15: The name of this plugin could a little bit more descriptive. What do you think about `splitCss` or something similar?

packages/ckeditor5-dev-build-tools/src/plugins/cssStyles.ts
```diff
@@ -0,0 +1,305 @@
...
+			// Determine import order of files
+			for ( const file in bundle ) {
+				const outputChunk = bundle[ file ]! as OutputChunk;
+				const root = outputChunk.facadeModuleId!;
+				const modules = getCSSModules( root, this.getModuleInfo );
+
+				modules.forEach( id => ids.add( id as string ) );
+			}
+
+			const cssStylesheet = combineStylesheetsIntoOne( ids, styles );
```
filipsobol, line 48: This is just an idea, but we might be able to use `MagicString.Bundle' to concatenate multiple files. This would not only simplify our implementation, but we would also be able to get generate proper source maps, which are invaluable when debugging the code.

I also used this library in other plugin (still WIP), so we'll use is anyway.

https://github.com/rich-harris/magic-string?tab=readme-ov-file#bundling

packages/ckeditor5-dev-build-tools/src/plugins/cssStyles.ts
```diff
@@ -0,0 +1,305 @@
...
+		}
+	};
+}
+
+/**
+ * @param content is a CSS file content.
+ * @param banner
+ */
+function unifyFileContentOutput( content: string | undefined, banner: string ): string {
+	return `${ banner }\n${ content ? content : '' }\n`;
```
filipsobol, line 77: We can use `MagicString.prepend` to achieve this while still maintaining proper source maps (since string concatenation breaks them).

packages/ckeditor5-dev-build-tools/src/plugins/cssStyles.ts
```diff
@@ -0,0 +1,305 @@
...
+ * Get `banner` from the `Rollup` configuration object.
+ */
+function getBanner( output: NormalizedOutputOptions, bundle: OutputBundle ): Promise<string> | string {
+	const mainChunk = Object
+		.values( bundle )
+		.filter( ( output ): output is OutputChunk => output.type === 'chunk' )
+		.find( chunk => chunk.isEntry )!;
+
+	return output.banner( mainChunk );
+}
```
filipsobol, line 90: Move this to a common util, as this code is also used in translations plugin.

packages/ckeditor5-dev-build-tools/src/plugins/cssStyles.ts
pszczesniak: Adresed in 4cb66cf.

packages/ckeditor5-dev-build-tools/package.json
```diff
@@ -24,14 +24,21 @@
...
+    "@rollup/plugin-json": "^6.1.0",
     "@rollup/plugin-node-resolve": "^15.2.3",
     "@rollup/plugin-typescript": "^11.1.6",
+    "@rollup/pluginutils": "^5.1.0",
+    "css": "^3.0.0",
     "glob": "^10.3.10",
+    "magic-string": "^0.30.7",
     "pofile": "^1.1.4",
     "rollup": "^4.9.5",
+    "rollup-plugin-delete": "^2.0.0",
```
filipsobol, line 36: This plugin will no be used

packages/ckeditor5-dev-build-tools/tests/plugins/splitCss/splitCss.test.ts
```diff
@@ -0,0 +1,263 @@
...
+		]
+	} );
+
+	const { output } = await bundle.generate( { format: 'esm', banner } );
+
+	return output;
+}
+
+test( 'Import of single `CSS` file', async () => {
+	const output = await generateBundle( join( import.meta.dirname, './fixtures/single-import/input.ts' ) );
```
filipsobol, line 39: AFAIK you should be able to use `./fixtures/single-import/body.css` as input, which would allow you to delete the `input.ts` file.

If this works, you can do the same for other tests.

packages/ckeditor5-dev-build-tools/tests/plugins/splitCss/splitCss.test.ts
```diff
@@ -0,0 +1,263 @@
...
+	const output = await generateBundle( join( import.meta.dirname, './fixtures/multiple-import/input.ts' ) );
+
+	const expectedResult = `
+body {
+color: '#000';
+}
+div {
+display: grid;
+}
+`;
```
filipsobol, line 76: (Nitpick, can be ignored if you don't like it) You can create a test util that will access a string and remove all whitespaces after new line. This would all you to properly format the code in tests while still making the tests pass.

```ts
function removeWhitespace( text: string ): string {
	return text.replaceAll( /\n\s+/gm, '\n' );
}
```

Then you can call it like so:

const test2 = removeWhitespace( `
	body {
		color: '#000';
	}
	div {
		display: grid;
	}
` );

packages/ckeditor5-dev-build-tools/tests/plugins/splitCss/splitCss.test.ts
```diff
@@ -0,0 +1,263 @@
...
+
+	const expectedEmptyResult = `${ TEST_BANNER }\n`;
+
+	verifyDividedStyleSheet( output, 'styles.css', expectedResult );
+	verifyDividedStyleSheet( output, 'editor-styles.css', expectedResult );
+	verifyDividedStyleSheet( output, 'content-styles.css', expectedEmptyResult );
+} );
+
+test( 'Import multiple `CSS` files', async () => {
+	const output = await generateBundle( join( import.meta.dirname, './fixtures/multiple-import/input.ts' ) );
```
filipsobol, line 67: Since all calls to the `generateBundle` start with `join( import.meta.dirname, ... )`, consider moving this part to that function instead.

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
```diff
@@ -0,0 +1,318 @@
...
+
+import { getBanner } from './utils.js';
+
+/**
+ * Filter files only with `css` extension.
+ */
+const filter = createFilter( [ '**/*.css' ] );
+
+export function splitCss(): Plugin {
+	const styles: Record<string, string> = {};
```
filipsobol, line 19: Consider renaming this to `files` or `stylesheets` so it's easier to remember what this variable holds.

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
```diff
@@ -0,0 +1,318 @@
...
+
+	const variablesUsedInEditorStylesContent: Set<string> = new Set(
+		dividedStylesheets.editorStylesContent!.match( VARIABLE_DEFINITION_REGEXP ) );
+	const variablesUsedInEditingViewStylesContent: Set<string> = new Set(
+		dividedStylesheets.editingViewStylesContent!.match( VARIABLE_DEFINITION_REGEXP ) );
+
+	const rootDeclarationForEditorStyles = createRootDeclarationOfUsedVariables(
+		rootDefinitions, variablesUsedInEditorStylesContent );
+	const rootDeclarationForEditingViewStyles = createRootDeclarationOfUsedVariables(
+		rootDefinitions, variablesUsedInEditingViewStylesContent );
```
filipsobol, line 208:
```suggestion
	const variablesUsedInEditorStylesContent: Set<string> = new Set(
		dividedStylesheets.editorStylesContent!.match( VARIABLE_DEFINITION_REGEXP )
	);

	const variablesUsedInEditingViewStylesContent: Set<string> = new Set(
		dividedStylesheets.editingViewStylesContent!.match( VARIABLE_DEFINITION_REGEXP )
	);

	const rootDeclarationForEditorStyles = createRootDeclarationOfUsedVariables(
		rootDefinitions,
		variablesUsedInEditorStylesContent
	);

	const rootDeclarationForEditingViewStyles = createRootDeclarationOfUsedVariables(
		rootDefinitions,
		variablesUsedInEditingViewStylesContent
	);
```

filipsobol: Nitpick

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
pszczesniak: I chosen `stylesheetsFiles`.

packages/ckeditor5-dev-build-tools/tests/plugins/splitCss/splitCss.test.ts
pszczesniak: @filipsobol Yes, it can be done like you suggest but it will be ok just for two or three testing cases.

From my point of view it can be missunderstood to someone who will look into tests, when we put a `css` file as in input when we default will point to a`ts` file.

Tests that will point to `ts` or `css` wont be groupped by input extension but spreaded across whole test suite.

Just thinking out loud to get all pros and cons for this suggested change.

Maybe we can do a sibgle test to show that we can make an input as a `css` file, WDYT?

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
```diff
@@ -0,0 +1,267 @@
...
+		.values( bundle )
+		.find( chunk => chunk.fileName && ( chunk.fileName.endsWith( '.css' ) ) );
+
+	if ( !cssStylesheetChunk ) {
+		return '';
+	}
+
+	const cssStylesheet = ( cssStylesheetChunk as EmittedAsset ).source as string;
+
+	return cssStylesheet;
```
filipsobol, line 70:
```suggestion
	return ( cssStylesheetChunk as EmittedAsset ).source.toString();
```

arkflpc (approved):

PR #940, opened 2024-02-15
CLI for building packages for new installation methods

filipsobol:

Internal: Add a `ckeditor5-build-package` CLI command for building packages for new installation methods. See [#15708](https://github.com/ckeditor/ckeditor5/issues/15708).

---

### Additional information


packages/ckeditor5-dev-build-tools/src/config.ts
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * If build speed becomes an issue, we can replace typescript with swc. However, this
+ * would put the responsibility of checking the types and generating `.d.ts` files on
+ * the package author.
+ *
+ * In the below configuration, swc is also used for minification and mangling, but it
+ * doesn't do as well as terser. Both options should be checked before deciding to use
+ * one or the other.
+ *
+ * swc( {
```
arkflpc, line 43: ` ```js ` is missing (perhaps).

filipsobol: This comment is internal and will not be used for documentation, so there is no need for that :wink:

pszczesniak (approved):

PR #943, opened 2024-02-23
Do not duplicate `CSS` variables declaration in `:root` selector (if so make a warning).

pszczesniak:

Type: Message. Closes #000.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
+++ b/packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
@@ -7,6 +7,7 @@ import { createFilter } from '@rollup/pluginutils';
 import { parse, type Rule, type Declaration, type Stylesheet } from 'css';
 import type { Plugin, OutputBundle, NormalizedOutputOptions, EmittedAsset } from 'rollup';
 import type { Processor } from 'postcss';
+import chalk from 'chalk';
 import cssnano from 'cssnano';
 
 export interface RollupSplitCssOptions {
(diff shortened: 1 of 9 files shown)


packages/ckeditor5-dev-build-tools/tests/plugins/splitCss/fixtures/duplicated-definitions/file2.css
```diff
@@ -0,0 +1,8 @@
+:root {
+	--variable1: blue;
```
filipsobol, line 2: I'd be good if this had different value from the `--variable1` in the `file1.css`. This way we could check if last definition of a given variable is used (the same as in CSS).

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
```diff
@@ -119,11 +120,21 @@ function getDividedStyleSheetsDependingOnItsPurpose( rules: Array<Rule> ) {
 		editingViewStylesContent += objectWithDividedStyles.editingViewStyles;
 
 		if ( objectWithDividedStyles.rootDefinitions.length ) {
-			rootDefinitionsList.push( ...objectWithDividedStyles.rootDefinitions );
+			objectWithDividedStyles.rootDefinitions.forEach( definitionText => {
+				const listOfDefinitions = definitionText.split( '\n' );
+
+				listOfDefinitions.forEach( definition => {
+					if ( definition && rootDefinitionsSet.has( definition ) ) {
+						showInfoOfDuplicatedVariableDefinition( definition );
```
filipsobol, line 128: Variable defined later should override previous value (the same as in CSS).

filipsobol (changes requested):

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
pszczesniak: TBH whole method should be modified. I choose `Set` to quickly eliminate the duplicates, but each `rootDefinitionsSet` helds a string like `--variable: red;` and it compares it with same one not just a variable name.

To make it work as expected `Set` should be changed into `Map` and store values as key/value: `'--variable'` => `'red'`. Check if the key already exists and if yes, replace just the value (and can make some warning that value `X` of variable `FOO` was changed to `Y` ).

pszczesniak:
Decision was made to not do it now and wait when we wants to add/modified existing source maps to splitted `CSS` files.

PR #946, opened 2024-03-22
Update CLI to allow bundling commercial packages

filipsobol:

Type: Message. Closes #000.

---

### Additional information


packages/ckeditor5-dev-build-tools/src/build.ts
```diff
@@ -5,20 +5,20 @@
 
 import fs from 'fs';
```
pszczesniak, line 6:
```suggestion
import { rmSync } from 'fs';
```

packages/ckeditor5-dev-build-tools/src/build.ts
```diff
@@ -93,32 +106,51 @@ async function normalizeOptions( options: Partial<BuildOptions> ): Promise<Build
...
+		/**
+		 * Create Rollup configuration based on provided arguments.
+		 */
+		const config = await getRollupConfig( args );
+
+		/**
+		 * Remove old build directory.
+		 */
+		if ( args.clean ) {
+			fs.rmSync( getCwdPath( 'dist' ), { recursive: true, force: true } );
```
pszczesniak, line 121:
```suggestion
			rmSync( getCwdPath( 'dist' ), { recursive: true, force: true } );
```

packages/ckeditor5-dev-build-tools/src/build.ts
filipsobol: This is needed for mocking in tests.

packages/ckeditor5-dev-build-tools/src/plugins/replace.ts
```diff
@@ -22,26 +24,56 @@ export interface RollupReplaceOptions {
...
+		return [
+			'ImportDeclaration',
+			'ExportAllDeclaration',
+			'ExportNamedDeclaration'
+		].includes( node.type );
+	}
+
 	return {
-		name: 'cke5-replace',
+		name: 'cke5-import-replace',
```
pszczesniak, line 42:
```suggestion
		name: 'cke5-replace-import',
```

To be consistent with the function name `replaceImports`.

pszczesniak (approved):
Found one tiny thing to consider, other than that looks :+1:

PR #955, opened 2024-05-22
Added ignore option.

przemyslaw-zan:

Feature (ci): Added `--ignore` option to the `ckeditor5-dev-ci-circle-workflow-notifier` task that allows ignoring jobs while calculating whether the notification should be sent. Closes ckeditor/ckeditor5#16403.

---

### Additional information


pomek:
Could you rephrase the comments and the proposed commit message to avoid the word "fail?" We don't know the status of ignored jobs. Just don't care if they are finished, errors, not started, or whatever.

packages/ckeditor5-dev-ci/bin/circle-workflow-notifier.js
```diff
@@ -97,7 +98,8 @@ async function getOtherJobsData() {
 function parseArguments( args ) {
 	const config = {
 		string: [
-			'task'
+			'task',
+			'ignore'
```
pomek, line 102: A potential user can use:

* `--ignore job` to ignore a single job,
* `--ignore job1 --ignore job2` to ignore many jobs,
* `--ignore job1,job2` (as above, but isn't aware of _array_ syntax).

Let's check if the `--ignore` value is a string and normalize it as an array.

Also, make sure to update JSDoc.

PR #964, opened 2024-07-02
Treat dependencies in the `dist` folder as production dependencies.

filipsobol:

Fix (dependency-checker): Treat dependencies in the `dist` folder as production dependencies. See ckeditor/ckeditor5#16646.

Other (build-tools): The `.d.ts` files for translations should import directly from `@ckeditor/ckeditor5-utils` instead of `ckeditor5`. See ckeditor/ckeditor5#16646.

MINOR BREAKING CHANGE: Treat dependencies in the `dist` folder as production `dependencies`.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-build-tools/src/plugins/translations.ts
+++ b/packages/ckeditor5-dev-build-tools/src/plugins/translations.ts
@@ -14,7 +14,7 @@ import type { Plugin } from 'rollup';
 import { removeWhitespace } from '../utils';
 
 const TYPINGS = removeWhitespace( `
-	import type { Translations } from 'ckeditor5';
+	import type { Translations } from '@ckeditor/ckeditor5-utils';
 
 	declare const translations: Translations;
 	export default translations;
--- a/packages/ckeditor5-dev-dependency-checker/lib/checkdependencies.js
+++ b/packages/ckeditor5-dev-dependency-checker/lib/checkdependencies.js
@@ -69,7 +69,7 @@ async function checkDependenciesInPackage( packagePath, options ) {
 			'**/*.ts': depCheck.parser.typescript,
 			'**/*.vue': depCheck.parser.vue
 		},
-		ignorePatterns: [ 'docs', 'build', 'dist' ],
+		ignorePatterns: [ 'docs', 'build' ],
 		ignoreMatches: [ 'eslint*', 'webpack*', 'husky', 'lint-staged' ]
 	};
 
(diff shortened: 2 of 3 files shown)


packages/ckeditor5-dev-build-tools/src/plugins/translations.ts
```diff
@@ -14,7 +14,7 @@ import type { Plugin } from 'rollup';
 import { removeWhitespace } from '../utils';
 
 const TYPINGS = removeWhitespace( `
-	import type { Translations } from 'ckeditor5';
+	import type { Translations } from '@ckeditor/ckeditor5-utils';
```
pomek, line 17: As it's a bug fix, we should have a test for it.

filipsobol: The `Fix` was the best commit type I could think of, but maybe `Other` is actually better. There's no bug in the current implementation - this change is just so that packages have `@ckeditor/ckeditor5-utils` as a dependency, not the entire `ckeditor5'.

IMO testing this doesn't make much sense, as the test would have to have a copy of this string and check if it matches.

pomek: If so, I would check if types are not imported from `ckeditor5`, and that would be enough in case somebody wants to update it in the future.

filipsobol: I added a new test for translations typings. :heavy_check_mark:

packages/ckeditor5-dev-dependency-checker/lib/checkdependencies.js
```diff
@@ -381,9 +381,30 @@ async function isDevDependency( packageName, absolutePaths ) {
...
+
+		/**
+		 * This folder contains the compiled code of the packages. Most of this code is the same
+		 * as the source, but during the build process some of the imports are replaced with those
+		 * compatible with the "new installation methods", which may use different dependencies.
+		 *
+		 * For example, the `ckeditor5/src/core.js` import is replaced with `@ckeditor/ckeditor5-core/dist/index.js`.
+		 *                   ^^^^^^^^^                                       ^^^^^^^^^^^^^^^^^^^^^^^^
+		 */
+		/[/\\]dist[/\\]/
```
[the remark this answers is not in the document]
filipsobol: Done :heavy_check_mark:

pomek (approved):

PR #966, opened 2024-07-04
Fix for not adding a new line while parsing data:image CSS definitions.

pszczesniak:

Fix: Should not corrupt the CSS code during splitting CSS into editor and content files. See https://github.com/ckeditor/ckeditor5/issues/16670.

---

### Additional information


packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
```diff
@@ -269,5 +270,13 @@ async function unifyFileContentOutput( content: string = '', minimize: boolean )
  * Wraps `declarations` list into passed `selector`;
  */
 function wrapDefinitionsIntoSelector( selector: string, definitions: string ): string {
+	// When definition contains `data:image` it should be in one following the specification.
```
filipsobol, line 273:
```suggestion
	// When definition contains `data:image` it should be in one line following the specification.
```

?

filipsobol (approved):

PR #970, opened 2024-07-09
Fix (build-tools): Keep nested CSS variables when splitting stylesheets into editor- or content-only stylesheets.

filipsobol:

Fix (build-tools): Preserve CSS variables used by other CSS variables when splitting styles into editor- or content-only stylesheets. Fixes [ckeditor/ckeditor5#16689](https://github.com/ckeditor/ckeditor5/issues/16689).

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-build-tools/tests/plugins/splitCss/fixtures/nested-css-variables/file1.css
+++ b/packages/ckeditor5-dev-build-tools/tests/plugins/splitCss/fixtures/nested-css-variables/file1.css
@@ -1,10 +1,14 @@
 :root {
+	/* Test CSS variables use by other CSS variables, including fallback values. */
 	--ck-spacing-unit: var(--ck-variable-1);
 	--ck-variable-1: var(--ck-variable-2);
-	--ck-variable-2: var(--ck-variable-3);
-	--ck-variable-3: var(--ck-variable-4);
-	--ck-variable-4: var(--ck-variable-5);
-	--ck-variable-5: 0.6em;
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-dev-build-tools/tests/plugins/splitCss/fixtures/nested-css-variables/file1.css
```diff
@@ -0,0 +1,12 @@
+:root {
+	--ck-spacing-unit: var(--ck-variable-1);
+	--ck-variable-1: var(--ck-variable-2);
+	--ck-variable-2: var(--ck-variable-3);
+	--ck-variable-3: var(--ck-variable-4);
+	--ck-variable-4: var(--ck-variable-5);
+	--ck-variable-5: 0.6em;
+}
```
pomek, line 8: I would add a few more assertions:

1. `--ck-variable-x` that uses `--ck-variable-y` but both aren't used anywhere.
2. A variable that is a default value if another variable does not exist: `var( --ck-variable-a, var( --ck-variable-b ) )`.

filipsobol: Done :heavy_check_mark:

pomek (approved):
I briefly reviewed the proposed changes in the source code, so the second pair of eyes should also look here, but it looks good. As you could notice, I paid more attention to tests :D

pszczesniak (approved):
I've tested the solution and it works as expected. I did also some side check quests and also all works.

PR #971, opened 2024-07-10
Internal: Update to Vitest@2

filipsobol:

Internal: Update to Vitest@2.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-build-tools/package.json
+++ b/packages/ckeditor5-dev-build-tools/package.json
@@ -56,10 +56,10 @@
     "@types/lodash-es": "^4.17.12",
     "@types/node": "^20.11.5",
     "@types/postcss-mixins": "^9.0.5",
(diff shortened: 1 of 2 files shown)


pomek:
Could I get short highlights why do we want to update and what it offers?

filipsobol:
In general, I created this PR because the new version works without major changes to the config or the tests themselves.

> [...] why we want to update

Just to be on the latest version and to be able to refer to the documentation without having to check if the given API is supported in 1.x.

> [...] what it offers?

[Here's the changelog](https://github.com/vitest-dev/vitest/releases/tag/v2.0.0), but TLDR is that it introduces a bunch of general bug fixes. It also adds a lot of useful browser testing features which we won't use in this repo, but we do/will do in the Vue and React repos, and it would be nice to be on the same version of Vitest in all of them.

pomek (approved):

PR #972, opened 2024-07-10
Preserve all selectors in CSS files during split into editor and content stylesheets.

pszczesniak:

Fix (build-tools): Should preserve all selectors and at-rules in CSS files during split into `editor` and `content` stylesheets. Closes https://github.com/ckeditor/ckeditor5/issues/16703.

---

### Additional information

The main problem was when we were splitting main css file into `editor` and `content`. Mainly it was done using `.ck-content` CSS class as a main target what should be put into `content` file and all the rest should land into `editor`. But this condition didn't include neither `@media` queries or  animations with `@keyframes` that class contains.

Currently all `@media` queries are splitted by class name mention above.

The `@keyframes` are gathered from whole main css file; also all selectors that are containing `animation` or `animation-name` are gathered and compared in against selector that are used into. By this they can be correctly divided and added to desired css stylesheets.

Also some CSS classes were omitted when there were more than one selector in definition. It's also fixed in this PR.

### DIFF
--- a/packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
+++ b/packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
@@ -394,14 +394,7 @@ function getSplittedStyleKeyframes( keyframesRules: Array<KeyFrames>, allRulesCo
 	keyframesRules.forEach( keyframeRule => {
 		const animationName = keyframeRule.name;
 
-		if ( !animationName ) {
-			return;
-		}
-
 		allRulesContainingAnimations.forEach( rule => {
-			if ( rule.type !== 'rule' ) {
-				return;
-			}
 			const declarations = rule.declarations as Array<Declaration>;
 			const selectorStartsWithCkContent = isSelectorStartsWithCkContent( rule );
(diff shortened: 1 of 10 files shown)


filipsobol (approved):

pomek:
Could I ask you to explain what the problem was?

pszczesniak:
> Could I ask you to explain what the problem was?

Will wrote an explanation when i finish the task, for now it's still WIP.

pszczesniak:
@pomek I've updated the `additional information` in PR description.

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
```diff
@@ -35,6 +43,11 @@ const filter = createFilter( [ '**/*.css' ] );
  */
 const VARIABLE_DEFINITION_REGEXP = /--([\w-]+)/gm;
 
+/**
+ * CSS rule type which are supported by this plugin.
+ */
+const SUPPORTED_RULE_TYPES = [ 'rule', 'media', 'keyframes' ];
```
filipsobol, line 49: This is fine for now, but I worry that long-term we might miss other types.

filipsobol:
Work will continue in #979.

PR #979, opened 2024-07-18
New approach to separating editor and content CSS

pszczesniak:

Fix (build-tools): Should preserved all needed selectors and at-rules in CSS files after split into editor and content stylesheets. Closes https://github.com/ckeditor/ckeditor5/issues/16703.

---

### Additional information

The main problem was when we were splitting main css file into `editor` and `content`. Mainly it was done using `.ck-content` CSS class as a main target what should be put into `content` file and all the rest should land into `editor`. But this condition didn't include neither `@media` queries or  animations with `@keyframes` that class contains.

First approach was to write it by us (PR: https://github.com/ckeditor/ckeditor5-dev/pull/972) but it was a lot of code and it wans't bullet proof for future. So decision was made to try something else: `PurgeCss` lib.

With using `PurgeCss` all at-rules are splitted correctly.

Also some CSS classes were omitted when there were more than one selector in definition. It's also fixed in this PR.

### DIFF
--- a/packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
+++ b/packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
@@ -29,7 +29,7 @@ const filter = createFilter( [ '**/*.css' ] );
 
 const REGEX_FOR_REMOVING_VAR_WHITESPACE = /(?<=var\()\s+|\s+(?=\))/g;
 
-// We must preserve all variables, keyframes and font faces.
+// We must preserve all variables, keyframes and font faces in splitted stylesheets.
 // For example this is caused by case when some of them can be defined in the `ckeditor5`
 // but used in `ckeditor5-premium-features` stylesheet and vice versa.
 const COMMON_PURGE_OPTIONS = {
--- a/packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
+++ b/packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
@@ -47,10 +47,24 @@ const purgeCssOptionsForEditor = {
 	...commonPurgeCssOptions,
 	// Pseudo class`:where` is preserved only if the appropriate html structure matches the CSS selector.
 	// It's a temporary solution to avoid removing selectors for Show blocks styles where `:where` occurs.
+	// For example this structure will be omitted without the HTML content:
+	//
+	// ```css
+	// .ck.ck-editor__editable.ck-editor__editable_inline.ck-show-blocks:not(.ck-widget)
+	//     :where(figure.image, figure.table) figcaption { /* ... */ }
+	// ```
+	//
 	// See: https://github.com/FullHuman/purgecss/issues/978
 	content: [ {
-		raw: '<html><body><div class="ck ck-editor__editable ck-editor__editable_inline ck-show-blocks">' +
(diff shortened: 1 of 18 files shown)


packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
```diff
@@ -30,10 +30,35 @@ export interface RollupSplitCssOptions {
...
+	content: [],
+	safelist: { deep: [ /^ck-content/ ] },
+	blocklist: []
+};
+
+const purgeCssOptionsForEditor = {
+	...commonPurgeCssOptions,
+	// Pseudo class`:where` is preserved only if the appropriate html structure matches the CSS selector.
+	// It's a temporary solution to avoid removing selectors for Show blocks styles where `:where` occurs.
+	// See: https://github.com/FullHuman/purgecss/issues/978
```
filipsobol, line 50: I'd be nice to know what **exact** issue this HTML solves.

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
```diff
@@ -30,10 +30,35 @@ export interface RollupSplitCssOptions {
...
+	content: [ {
+		raw: '<html><body><div class="ck ck-editor__editable ck-editor__editable_inline ck-show-blocks">' +
```
filipsobol, line 52: Would PurgeCSS handle this HTML if we used [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) and format this HTML nicely?

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
```diff
@@ -297,16 +170,21 @@ async function unifyFileContentOutput( content: string = '', minimize: boolean )
...
  */
-function wrapDefinitionsIntoSelector( selector: string, definitions: string ): string {
-	// When definition contains `data:image` it should be in one line following the specification.
-	// Currently used tool responsible for parsing definitions tries to split each definition into new line.
-	// When `data:image` contains `SVG` with style attribute which contains CSS definitions it splits it into new lines,
-	// which breaks the CSS.
-	if ( definitions.includes( 'data:image' ) ) {
-		definitions = removeNewline( definitions );
-	}
+async function normalizeStylesheet( content: string ): Promise<string> {
```
filipsobol, line 175: The name and description of this function don't tell me much about what it does.

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
pszczesniak: Addressed in 96d3bb7.

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
pszczesniak: Yes, it works with template literals. 96d3bb7

packages/ckeditor5-dev-build-tools/src/plugins/splitCss.ts
pszczesniak: I've change the description of the function to show exactly whit it does.

filipsobol (approved):

PR #981, opened 2024-07-25
Build tools extended with the ability to pass the globals parameter.

pszczesniak:

Fix (build-tools): Ability to pass `globals` parameter if necessary for external imports in `umd` bundles. See https://github.com/ckeditor/ckeditor5/issues/16798.

BREAKING CHANGE: Ability to pass `globals` parameter if necessary for external imports in `umd` bundles.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-build-tools/README.md
+++ b/packages/ckeditor5-dev-build-tools/README.md
@@ -148,9 +148,9 @@ This option is required if the `browser' option is enabled.
 
 **Default value:** `{}`
 
-A list of pairs necessary for external imports in `umd` bundles.
+Pairs of external package names and associated global variables used in the `umd` build.
 
-The list of `globals` already contains `ckeditor5` and `ckeditor5-premium-features` which are mapped to `CKEDITOR` and `CKEDITOR_PREMIUM_FEATURES`.
+The list already contains `ckeditor5` and `ckeditor5-premium-features` which are mapped to `CKEDITOR` and `CKEDITOR_PREMIUM_FEATURES`.
 
 When using the CLI, this option can be used multiple times.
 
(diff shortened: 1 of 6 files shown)


packages/ckeditor5-dev-build-tools/src/utils.ts
```diff
@@ -60,3 +61,11 @@ export function getUserDependency( name: string ): any {
...
 	return require( path );
 }
+
+/**
+ * `ckeditor5` and `ckeditor5-premium-features` globals.
+ */
+export const CKEDITOR_GLOBALS: GlobalsOption = {
+	ckeditor5: 'CKEDITOR',
+	'ckeditor5-premium-features': 'CKEDITOR_PREMIUM_FEATURES'
+};
```
filipsobol, line 71: Having this const in utils may be overkill, since we only use it in one place.

packages/ckeditor5-dev-build-tools/README.md
```diff
@@ -142,6 +142,30 @@ The name of the UMD bundle. This name will be used as the global variable name w
...
 
+#### `globals` / `--globals=[moduleID:Global]`
+
+**Type:** `{ [ name: string ]: string } | ( ( name: string ) => string )` | `string[]`
+
+**Default value:** `{}`
+
+A list of pairs necessary for external imports in `umd` bundles.
+
+The list of `globals` already contains `ckeditor5` and `ckeditor5-premium-features` which are mapped to `CKEDITOR` and `CKEDITOR_PREMIUM_FEATURES`.
```
filipsobol, line 153:
```suggestion
Pairs of external package names and associated global variables used in the `umd` build.

The list already contains `ckeditor5` and `ckeditor5-premium-features` which are mapped to `CKEDITOR` and `CKEDITOR_PREMIUM_FEATURES`.
```

filipsobol (approved):

packages/ckeditor5-dev-build-tools/src/utils.ts
pszczesniak: Yes, it makes sense. Fixed in a71ea6f.

filipsobol (approved):

PR #996, opened 2024-09-02
Add default license key to manual and automated tests.

mmotyczynska:

Internal: Adds default license key to manual and automated tests.

---

### Additional information


pomek (commented):
Could we add two tests (one per config type - automated and manual) that verify if we pass the `window?CKEDITOR_...` string to the `webpack.DefinePlugin` constructor?

Similar to: https://github.com/ckeditor/ckeditor5-dev/blob/e0775b7c213e68ed0eef9f007da1209f93a77757/packages/ckeditor5-dev-tests/tests/utils/automated-tests/getwebpackconfig.js#L126-L137

It should help us when we switch tooling used in the environment to ensure that this variable is injected properly.

martnpaneq (approved):
Looks good. The global const `CKEDITOR_GLOBAL_LICENSE_KEY` is added correctly to `window` both in manual and automated tests. I'm missing the context why it's needed though, as I could not find any link in the issue. Maybe it would be a good idea to add it for future reference.

PR #997, opened 2024-09-04
Converted the `ckeditor5-dev-ci` package to ESM

przemyslaw-zan:

Internal (ci): Converted the package to ESM.

---

### Additional information


packages/ckeditor5-dev-ci/lib/index.js
```diff
@@ -3,9 +3,7 @@
  * For licensing, see LICENSE.md.
  */
 
-'use strict';
-
-module.exports = {
+export default {
 	getJobApprover: require( './utils/get-job-approver' ),
 	members: require( './data/members.json' )
```
pomek, line 8: This needs to be changed. See: https://github.com/ckeditor/ckeditor5-dev/blob/ci/3759-release-tools/packages/ckeditor5-dev-release-tools/lib/index.js.

For the JSON file, we have the options `fs-extra` and `readJson()`, or we can use `import with`. Whatever we decide, we must be consistent across the repository.

@ckeditor/ckeditor-5-devops, WDYT?

packages/ckeditor5-dev-ci/lib/trigger-circle-build.js
```diff
@@ -5,7 +5,7 @@
 
 /* eslint-env node */
 
-const fetch = require( 'node-fetch' );
+import nodeFetch from 'node-fetch';
```
pomek, line 8: I think it's time to drop `node-fetch` in favor of the native library while we upgrade dependencies.

packages/ckeditor5-dev-ci/tests/format-message.js
```diff
@@ -181,7 +172,7 @@ describe( 'lib/format-message', () => {
 
 			expect( message ).to.be.an( 'object' );
 			expect( message ).to.have.property( 'text' );
-			expect( message.text ).to.equal( '<@slackId>, could you take a look?' );
+			expect( message.text ).toEqual( '<!channel> (Example Nick), could you take a look?' );
```
pomek, line 175: Why did you change the assertion? This test assumes that we find a Slack ID.

pomek (commented):
I left a few comments that should be addressed before merging.

packages/ckeditor5-dev-ci/lib/index.js
psmyrek: If it is possible, let's use native solutions wherever possible.

packages/ckeditor5-dev-ci/vitest.config.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @license Copyright (c) 2003-2024, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig( {
+	test: {
```
martnpaneq, line 9: For now this is fine, but if we are going to rewrite other packages, we need to store the config in a common place.

pomek (changes requested):
ESM requires extensions in `import`/`export` when operating on local files. You can fix it by updating ESLint rules.

https://github.com/ckeditor/ckeditor5-dev/blob/281aaeaf5495ed231373bf646cccb116d2ccc73e/.eslintrc.js#L28-L33

pomek (approved):

PR #1005, opened 2024-09-11
Convert `ckeditor5-dev-release-tools` to ESM

pomek:

Other (release-tools): Convert `ckeditor5-dev-release-tools` to ESM.

---

### Additional information


packages/ckeditor5-dev-release-tools/vitest.config.js
```diff
@@ -0,0 +1,54 @@
...
+			include: [
+				'lib/**'
+			],
+
+			reporter: [ 'text', 'json', 'html', 'lcov' ]
+		}
+	}
+} );
+
+function executeInParallelVirtualModulePlugin() {
```
martnpaneq, line 37: Could you explain how does it work and what are the benefits? Could it be used in other modules?

psmyrek (approved):
LGTM 🎉

martnpaneq (approved):

packages/ckeditor5-dev-release-tools/vitest.config.js
pomek: A Vitest environment provides a fake module that can be imported from a `virtual:parallelworker-integration-module` path.

As a worker script (`parallelworker.js`) loads a callback to execute via an import (previously via `require()`), we need to provide something that would not throw an error. Due to different `cwd` depending on a trigger of tests (an npm script, an IDE integration, etc.), we need to predict an absolute path. While it's doable, it complicates things that do not have to.

Hence, I decided to create a fake module that is not used in source code but only in tests.

PR #1015, opened 2024-09-26
Fixed `executeInParallel()` function throwing error when the number o…

martnpaneq:

Fix (release-tools): Passing an odd number as a value of the `concurrency` option will not crash the `executeInParallel()` function. Closes ckeditor/ckeditor5#17025.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
@@ -48,6 +48,7 @@ export default async function executeInParallel( options ) {
 		concurrency = os.cpus().length / 2
 	} = options;
 
+	const concurrencyAsInteger = Math.floor( concurrency ) || 1;
 	const normalizedCwd = upath.toUnix( cwd );
 	const packages = ( await glob( `${ packagesDirectory }/*/`, {
 		cwd: normalizedCwd,
@@ -58,7 +59,7 @@ export default async function executeInParallel( options ) {
 		packages.filter( packagesDirectoryFilter ) :
 		packages;
 
-	const packagesInThreads = getPackagesGroupedByThreads( packagesToProcess, concurrency );
+	const packagesInThreads = getPackagesGroupedByThreads( packagesToProcess, concurrencyAsInteger );
 
 	const callbackModule = upath.join( cwd, crypto.randomUUID() + '.mjs' );
 	await fs.writeFile( callbackModule, `export default ${ taskToExecute };`, 'utf-8' );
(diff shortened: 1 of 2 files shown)


pomek:
Please rebase your branch with the base one. It looks like you started from master.

packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
```diff
@@ -48,6 +48,7 @@ export default async function executeInParallel( options ) {
 		concurrency = os.cpus().length / 2
 	} = options;
 
+	const concurrencyAsInteger = Math.floor( concurrency ) || 1;
```
pomek, line 51: We need a few tests:
* `concurrency ` is a float number (specified as an argument)
* `concurrency ` is a float number (defined as a default value, `os.cpus()` returns an odd number)
* `concurrency ` is an integer equal to `0`

I would skip negative numbers.

packages/ckeditor5-dev-release-tools/tests/utils/executeinparallel.js
```diff
@@ -334,6 +334,56 @@ describe( 'executeInParallel()', () => {
...
+		for ( const worker of stubs.WorkerMock.instances ) {
+			getExitCallback( worker )( 0 );
+		}
+
+		await promise;
+	} );
+
+	it( 'should round down to the closest integer (`concurrency`)', async () => {
+		const options = Object.assign( {}, defaultOptions, {
+			concurrency: 3.5
```
pomek, line 353: We need to duplicate this test by calling the `executeInParallel` function with the default value of `concurrency` (which is mocked: `os().cpus().length` returns `7`)

martnpaneq: I think if we test `concurrency: 3.5`, there is no need to test the behaviour for default value, because it behaves exactly the same as the value passed by us.

I think we just need to check if the default value is actually used.

This simplifies the tests, because it's hard to mock consts.

pomek: The original issue was reported by using `os.cpus()` construction and I want to make sure (even if it duplicated a case when passing `concurrency` manually) it works as expected.

packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
martnpaneq: Added in [c6ffb35](https://github.com/ckeditor/ckeditor5-dev/pull/1015/commits/c6ffb35970cbdb01d04dff7f40665990f72fe160)

pomek (approved):

PR #1019, opened 2024-10-07
Move the version checks to changelog tasks

pomek:

Feature (ci): Created a new binary script called `ckeditor5-dev-ci-is-workflow-restarted` that returns with a non-zero exit code if a given workflow is executed for the first time. The restarted workflows exit with a zero exit code.

Feature (release-tools): A user-provided version will be checked against npm availability while generating a changelog. If it is already taken, the tools will not allow it to be used.

Other (release-tools): The `updateVersions()` task will no longer verify if the specified `version` is available on npm.

Other (release-tools): The `publishPackages()` task filters out already published packages to avoid pushing the same archive twice. Thanks to that, it can be a part of a process that would be restarted.

Other (release-tools): The `publishPackages()` task tries to publish the package once again when it fails independently from the returned error code. Previously, it was scheduled only when the `E409` error occurred.

Other (release-tools): The `verifyPackagesPublishedCorrectly()` task is no longer available as its responsibility has been merged into the `publishPackages()` task.

Other (release-tools): The `commitAndTag()` task does not commit files if a tag for the specified version is already created.

---

Closes cksource/ckeditor5-internal#3828.

---

### Additional information


psmyrek (dismissed):
LGTM 🎉

psmyrek (approved):
LGTM 🎉

martnpaneq (approved):

PR #1025, opened 2024-10-16
Introduced script for moving translations between packages

psmyrek:

Feature (translations): Introduced `moveTranslations()` function to move requested translations between packages. It removes contexts and translated messages from language files ("*.po" files) from the source package and adds (or overwrites) them in the destination package.

---

### Additional information

Closes cksource/ckeditor5-internal#3833.

### DIFF
--- a/packages/ckeditor5-dev-translations/lib/utils/getpackagecontext.js
+++ b/packages/ckeditor5-dev-translations/lib/utils/getpackagecontext.js
@@ -1,0 +1,32 @@
+/**
+ * @license Copyright (c) 2003-2024, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import upath from 'upath';
+import fs from 'fs-extra';
+import { CONTEXT_FILE_PATH } from './constants.js';
+
+/**
+ * @param {object} options
+ * @param {string} options.packagePath Path to the package containing the context.
+ * @returns {Context}
+ */
+export default function getPackageContext( { packagePath } ) {
+	const contextFilePath = upath.join( packagePath, CONTEXT_FILE_PATH );
+	const contextContent = fs.readJsonSync( contextFilePath, { throws: false } ) || {};
+
+	return {
+		contextContent,
+		contextFilePath,
+		packagePath
+	};
+}
+
+/**
+ * @typedef {object} Context
+ *
+ * @property {string} contextFilePath
+ * @property {object} contextContent
+ * @property {string} packagePath
+ */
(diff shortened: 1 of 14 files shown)


pomek (approved):

pomek:
While everything works as expected, I found a case that crashes the script. Execute the script with the following content:

``json
[
  {
    "source": "./packages/ckeditor5-core",
    "destination": "./packages/ckeditor5-autoformat",
    "messageId": "Cancel"
  },
  {
    "source": "./packages/ckeditor5-core",
    "destination": "./packages/ckeditor5-autoformat",
    "messageId": "Cancel"
  }
]
```

I agree that developers should not pass invalid entries. However, this might happen due to copying. When moving many entries, we can omit them by mistake, which isn't our intention.

Let's prevent it from crashing in such scenarios. We could:

1. filter out duplicated items,
2. avoid processing `messageId` when it has already been updated,
3. validate the configuration and exit when invalid entries are found,
4. ...?

Up to you what you decide :D

File: packages/ckeditor5-dev-translations/lib/utils/getpackagecontext.js [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+
+	return {
+		contextContent,
+		contextFilePath,
+		packagePath
+	};
+}
+
+/**
+ * @typedef {object} Context
```
pomek, line 27:
```suggestion
 * @typedef {object} TranslationsContext
```
WDYT?

packages/ckeditor5-dev-translations/lib/utils/getsourcemessages.js
```diff
@@ -30,3 +30,14 @@ export default function getSourceMessages( { packagePaths, sourceFiles, onErrorC
 			return sourceMessages;
 		} );
 }
+
+/**
+ * @typedef {object} Message
```
pomek, line 35:
```suggestion
 * @typedef {object} TranslatableEntry
```

WDYT?

packages/ckeditor5-dev-translations/lib/utils/getpackagecontext.js
psmyrek: Done in https://github.com/ckeditor/ckeditor5-dev/pull/1025/commits/5eb675f3874e4aefc83807afcfe12c182521eccd.

packages/ckeditor5-dev-translations/lib/utils/getsourcemessages.js
psmyrek: Done in https://github.com/ckeditor/ckeditor5-dev/pull/1025/commits/5eb675f3874e4aefc83807afcfe12c182521eccd.

pomek (approved):
LGTM.

PR #1032, opened 2024-10-22
Added skip ci to the `commitAndTag()` function.

martnpaneq:

Other (release-tools): The `commitAndTag()` function understands a new parameter called `skipCi`. By default, release commits will not trigger a new workflow on CI. This behavior can be disabled when passing the `false` value.

---

### Additional information


packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
```diff
@@ -16,9 +16,10 @@ const { toUnix } = upath;
  * @param {string} options.version The commit will contain this param in its message and the tag will have a `v` prefix.
  * @param {Array.<string>} options.files Array of glob patterns for files to be added to the release commit.
  * @param {string} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
+ * @param {boolean} [options.skipCi=true] Whether to add [skip ci] to the commit.
```
pomek, line 19:
```suggestion
 * @param {boolean} [options.skipCi=true] Whether to add the "[skip ci]" suffix to the commit message.
```

PR #1042, opened 2024-10-28
Added support for providing `cwd` for translations scripts

psmyrek:

Fix (translations): Added support for providing `options.cwd` parameter to `synchronizeTranslations()` and `moveTranslations()` functions. This `options.cwd` (or `process.cwd()` if not set) is used to resolve paths to packages.

---

### Additional information

Closes cksource/ckeditor5-internal#3849.

### DIFF
--- a/packages/ckeditor5-dev-translations/lib/movetranslations.js
+++ b/packages/ckeditor5-dev-translations/lib/movetranslations.js
@@ -3,6 +3,7 @@
  * For licensing, see LICENSE.md.
  */
 
+import upath from 'upath';
 import fs from 'fs-extra';
 import { logger } from '@ckeditor/ckeditor5-dev-utils';
 import getPackageContext from './utils/getpackagecontext.js';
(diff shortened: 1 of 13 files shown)


packages/ckeditor5-dev-translations/lib/movetranslations.js
```diff
@@ -112,10 +113,38 @@ function assertContextsExist( { packageContexts, config } ) {
...
 }
 
+/**
+ * @param {MoveTranslationsOptions} options
+ */
+function normalizeOptions( options ) {
+	const { config } = options;
+
+	const cwd = options.cwd || process.cwd();
+	const toAbsolute = path => upath.resolve( cwd, path );
```
pomek, line 123:
```suggestion
	const toAbsolute = path => upath.join( cwd, path );
```

I think it fits our needs better as we want to merge two paths (an absolute and a relative). `resolve()` merges the specified arguments with `process.cwd()` if specified paths are relative. I know, it's an edge case here.

psmyrek: We do not want just merge two paths. We want to calculate an absolute path from two arguments. The output from `toAbsolute()` must be an absolute path. The result from `upath.join()` is not always and absolute path (i.e. when both arguments are relative).

```
upath.join( '.', 'x/y/z' ) !== upath.resolve( '.', 'x/y/z' )
```

Using `process.cwd()` with `upath.resolve()` in case integrator did not specify own absolute `cwd` path is fine.

pomek: Good point, I didn't think about passing a relative `cwd`.

pomek (approved):
LGTM.

PR #1048, opened 2024-11-21
Pass path to a dynamically imported file as `file://` protocol URL to make it work in Windows

psmyrek:

Fix (release-tools): Fixed path calculation in `executeInParallel()` util that prepares a path to a dynamically imported worker script. Previously, it did not work in Windows, because there is a specific requirement that absolute path must be valid `file://` protocol URL. Now, the worker script is always imported using `file://` protocol URL.

---

### Additional information

Closes cksource/ckeditor5-internal#3882.

I'm not using `url.pathToFileURL()` to convert the path to a `file://` protocol URL, because it produces unexpected results in tests where we mock `process.cwd()` to be Windows-style.

Example: when mocked `process.cwd()` is `C:\Users\ckeditor`, then `url.pathToFileURL( 'C:/Users/ckeditor/uuid-4.mjs' )` internally tries to create an URL from `file://C:%5CUsers%5Cckeditor/C:/Users/ckeditor/uuid-4.mjs` and throws `ERR_INVALID_URL`.

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
@@ -70,7 +70,13 @@ export default async function executeInParallel( options ) {
 		return createWorker( {
 			signal: signal || defaultAbortController.signal,
 			onPackageDone,
-			workerData: { packages, callbackModule, taskOptions }
+			workerData: {
+				packages,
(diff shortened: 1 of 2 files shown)


pomek (approved):
I would like to use `pathToFileURL()`, but I assume you tried it, and it did not work.

So, let's use the proposed solution. LGTM.

PR #1051, opened 2024-12-17
Fix (build-tools): Always enable TypeScript plugin, even when `declarations` is set to `false`.

filipsobol:

Fix (build-tools): Always enable the TypeScript plugin, even when `declarations` is set to `false`.

---

TBD: Explain why / add ticket number.


pomek (approved):
LGTM.

PR #1064, opened 2025-02-06
Improved calculating pre-release versions that start with the same base name

psmyrek:

Other (release-tools): Exported the `getDateIdentifier()` util that converts the current date into `YYYYMMDD` format.

Other (release-tools): Improved the `getLastPreRelease()` util to distinguish pre-release versions starting with the same base name. Previously, `getLastPreRelease( '0.0.0-nightly' )` matched both `0.0.0-nightly` and `0.0.0-nightly-next` versions, but now only the first one is matched.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/utils/versions.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/versions.js
@@ -50,7 +50,7 @@ export function getLastPreRelease( releaseIdentifier, cwd = process.cwd() ) {
 			const lastVersion = Object.keys( result.versions )
 				.filter( version => {
 					const optionalDateIdentifier = '(-[0-9]{8})?';
-					const optionalSequenceNumber = '(.[0-9]+)?';
+					const optionalSequenceNumber = '(\\.[0-9]+)?';
 					const versionRegExp = new RegExp( `^${ releaseIdentifier }${ optionalDateIdentifier }${ optionalSequenceNumber }$` );
 
 					return versionRegExp.test( version );
--- a/packages/ckeditor5-dev-release-tools/lib/utils/versions.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/versions.js
@@ -29,7 +29,14 @@ export function getLastFromChangelog( cwd = process.cwd() ) {
 
 /**
  * Returns the current (latest) pre-release version that matches the provided release identifier.
+ * It takes into account and distinguishes pre-release tags with different names but starting with the same base name.
  * If the package does not have any pre-releases with the provided identifier yet, `null` is returned.
+ *
+ * Examples:
+ * 	* "0.0.0-nightly" - Matches the last "nightly" version regardless of the publication date.
+ *	  It does not match other nightly tags that start with the same "nightly" base name, e.g. "0.0.0-nightly-next-YYYYMMDD.X".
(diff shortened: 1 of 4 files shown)


packages/ckeditor5-dev-release-tools/lib/utils/versions.js
```diff
@@ -41,7 +48,13 @@ export function getLastPreRelease( releaseIdentifier, cwd = process.cwd() ) {
 	return packument( packageName )
 		.then( result => {
 			const lastVersion = Object.keys( result.versions )
-				.filter( version => version.startsWith( releaseIdentifier ) )
+				.filter( version => {
+					const optionalDateIdentifier = '(-[0-9]{8})?';
+					const optionalSequenceNumber = '(.[0-9]+)?';
```
filipsobol, line 53:
```suggestion
					const optionalSequenceNumber = '(\.[0-9]+)?';
```

Match literal `.`

packages/ckeditor5-dev-release-tools/lib/utils/versions.js
```diff
@@ -41,7 +48,13 @@ export function getLastPreRelease( releaseIdentifier, cwd = process.cwd() ) {
...
+					const versionRegExp = new RegExp( `^${ releaseIdentifier }${ optionalDateIdentifier }${ optionalSequenceNumber }$` );
+
+					return versionRegExp.test( version );
```
filipsobol, line 56: Just a random idea, but maybe we should separate all bits that are important to us by `.` instead of a mix of `.` and `-`?

Given this example:

```text
0.0.0-nightly-next-20230615.0
                  ^        ^
```

This version number consists of:

```text
MAJOR.MINOR.PATCH-channel_and_date.version
```

However, [according to spec](https://semver.org/#spec-item-9), each identifier in the pre-release portion (after first `-`) should be dot-separated, so the correct version should be:

```text
0.0.0-nightly-next.20230615.0
                  ^ `.` instead of `-`
```
This version number consists of:

```text
MAJOR.MINOR.PATCH-channel.date.version
```

With this approach, we could use `semver` (which we already have) to parse version numbers instead of using Regexp.

```js
import semver from 'semver';

const data = semver.parse( '0.0.0-nightly-next.20230615.0' );

/**
 * data.raw = '0.0.0-nightly-next.20230615'
 * data.major = 0
 * data.minor = 0
 * data.patch = 0
 * data.prerelease = [ 'nightly-next', '20230615', 0 ]
 */
```

If this is the only place we do it, then it shouldn't be an issue, but if we have more such places, when relying on official npm tooling would be nice.

pomek: I am fine with it, but I wonder if it's a breaking change. Or, whether it should be marked as.

psmyrek: After discussing this within the team, we decided not to change the nightly versioning format.

packages/ckeditor5-dev-release-tools/lib/utils/versions.js
psmyrek: Addressed in https://github.com/ckeditor/ckeditor5-dev/pull/1064/commits/b90cd7269511ff2273115547c3de40842cf8071d.

PR #1077, opened 2025-03-04
Added `dryRun` option to the `commitAndTag` function in `preparepackages`.

martnpaneq:

Feature: Added `dryRun` option to the `commitAndTag` function in `preparepackages`. Closes ckeditor/ckeditor5#17967.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
+++ b/packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
@@ -6,6 +6,7 @@
 import upath from 'upath';
 import { glob } from 'glob';
 import { simpleGit } from 'simple-git';
+import { tools } from '@ckeditor/ckeditor5-dev-utils';
 
 const { toUnix } = upath;
 
@@ -17,9 +18,10 @@ const { toUnix } = upath;
  * @param {Array.<string>} options.files Array of glob patterns for files to be added to the release commit.
  * @param {string} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
  * @param {boolean} [options.skipCi=true] Whether to add the "[skip ci]" suffix to the commit message.
+ * @param {boolean} [options.dryRun=false] In order to run pre commit checks in a dry run mode.
  * @returns {Promise}
  */
-export default async function commitAndTag( { version, files, cwd = process.cwd(), skipCi = true } ) {
+export default async function commitAndTag( { version, files, cwd = process.cwd(), skipCi = true, dryRun = false } ) {
 	const normalizedCwd = toUnix( cwd );
 	const filePathsToAdd = await glob( files, { cwd: normalizedCwd, absolute: true, nodir: true } );
 
@@ -36,6 +38,19 @@ export default async function commitAndTag( { version, files, cwd = process.cwd(
 
 	// Do not commit and create tags if a tag is already taken. It might happen when a release job is restarted.
 	if ( tagForVersion ) {
+		return;
+	}
+
+	if ( dryRun ) {
+		try {
+			await git.add( filePathsToAdd );
+			await tools.shExec( 'yarn lint-staged', { cwd, verbosity: 'error', async: true } );
+		} catch ( e ) {
+			throw e.message;
+		} finally {
+			await git.reset( filePathsToAdd );
+		}
+
 		return;
 	}
 
--- a/packages/ckeditor5-dev-release-tools/tests/tasks/commitandtag.js
+++ b/packages/ckeditor5-dev-release-tools/tests/tasks/commitandtag.js
@@ -7,9 +7,11 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { glob } from 'glob';
 import commitAndTag from '../../lib/tasks/commitandtag.js';
 import { simpleGit } from 'simple-git';
+import { tools } from '@ckeditor/ckeditor5-dev-utils';
 
 vi.mock( 'simple-git' );
 vi.mock( 'glob' );
+vi.mock( '@ckeditor/ckeditor5-dev-utils' );
 
 describe( 'commitAndTag()', () => {
 	let stubs;
@@ -19,6 +21,8 @@ describe( 'commitAndTag()', () => {
 			git: {
 				tags: vi.fn(),
 				commit: vi.fn(),
+				reset: vi.fn(),
+				add: vi.fn(),
 				addTag: vi.fn()
 			}
 		};
@@ -122,5 +126,20 @@ describe( 'commitAndTag()', () => {
 		await commitAndTag( { version: '1.0.0', packagesDirectory: 'packages' } );
 
 		expect( stubs.git.addTag ).toHaveBeenCalledExactlyOnceWith( 'v1.0.0' );
+	} );
+
+	it( 'should run in dry run mode without creating a commit and a tag', async () => {
+		vi.mocked( glob ).mockResolvedValue( [ 'package.json', 'packages/ckeditor5-foo/package.json' ] );
+
+		await commitAndTag( { version: '1.0.0', packagesDirectory: 'packages', files: [ '**/package.json' ], dryRun: true } );
+
+		expect( stubs.git.add ).toHaveBeenCalledWith( [
+			'package.json',
+			'packages/ckeditor5-foo/package.json'
+		] );
+		expect( tools.shExec ).toHaveBeenCalledWith( 'yarn lint-staged', expect.objectContaining( { verbosity: 'error', async: true } ) );
+
+		expect( stubs.git.commit ).not.toHaveBeenCalled();
+		expect( stubs.git.addTag ).not.toHaveBeenCalled();
 	} );
 } );


martnpaneq:
This is how the example output looks in case of errors:
image

martnpaneq:
I'm going to use the `silent` option in `shExec` for a better text formatting:
image

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
```diff
@@ -39,6 +41,19 @@ export default async function commitAndTag( { version, files, cwd = process.cwd(
...
+			await git.add( filePathsToAdd );
+			await tools.shExec( 'yarn lint-staged', { cwd, verbosity: 'silent', async: true } );
+		} catch ( e ) {
+			throw e.message;
+		} finally {
+			await git.reset( filePathsToAdd );
+		}
+
+		return;
+	}
```
pomek, line 55:
```js
if( dryRun ) {
  // ...
} else {
  // ...
}
```

While both solutions do the same, I would like to keep `if/else` because the part after if applies the opposition behavior to the checked instruction.

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
```diff
@@ -39,6 +41,19 @@ export default async function commitAndTag( { version, files, cwd = process.cwd(
 		return;
 	}
 
+	if ( dryRun ) {
+		try {
+			await git.add( filePathsToAdd );
+			await tools.shExec( 'yarn lint-staged', { cwd, verbosity: 'silent', async: true } );
```
pomek, line 47: The executed task must be customizable. While it can use _our_ defaults, integrations should be able to define their tasks.

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
```diff
@@ -39,6 +41,19 @@ export default async function commitAndTag( { version, files, cwd = process.cwd(
 		return;
 	}
 
+	if ( dryRun ) {
+		try {
```
pomek, line 45: Could we move the entire try/catch/finally section to a dedicated function?

martnpaneq: Adressed in [7859810](https://github.com/ckeditor/ckeditor5-dev/pull/1077/commits/7859810ad6c8a1b9659aa2cc2171824b42507622).

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
martnpaneq: Adressed in [7859810](https://github.com/ckeditor/ckeditor5-dev/pull/1077/commits/7859810ad6c8a1b9659aa2cc2171824b42507622).

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
martnpaneq: Adressed in [7859810](https://github.com/ckeditor/ckeditor5-dev/pull/1077/commits/7859810ad6c8a1b9659aa2cc2171824b42507622).

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
```diff
@@ -39,6 +51,33 @@ export default async function commitAndTag( { version, files, cwd = process.cwd(
...
+
+/**
+ * Function that imitates running dry run for the `git commit` command.
+ *
+ * @param {Array.<string>} filePathsToAdd File paths to run the dry run validations for.
+ * @param {string} preCommitCommand Command to run before creating a commit for the dry run mode.
+ * @param {string} cwd Current working directory from which all paths will be resolved.
+ * @returns {Promise<void>}
+ */
+async function dryRunGitCommit( filePathsToAdd, preCommitCommand, cwd ) {
```
pomek, line 71:
```suggestion
async function executePreCommitHook( { cwd, git, hookCommand, filesToCommit } ) {
```
I propose something like this. I am unsure if we can get `cwd` from the `git` instance. I checked docs, but it allows changing it only.

packages/ckeditor5-dev-release-tools/tests/tasks/commitandtag.js
```diff
@@ -123,4 +127,19 @@ describe( 'commitAndTag()', () => {
 
 		expect( stubs.git.addTag ).toHaveBeenCalledExactlyOnceWith( 'v1.0.0' );
 	} );
+
+	it( 'should run in dry run mode without creating a commit and a tag', async () => {
```
pomek, line 131: Could you add one more test to verify if the specified `options.preCommitCommand` value is passed to `shExec()`?

pomek: Also, I do not see anything regarding to `listrTask.output` in tests.

martnpaneq: Adressed in [d47c4a6](https://github.com/ckeditor/ckeditor5-dev/pull/1077/commits/d47c4a68ce779b98208229a04d93400ea3c3dba3).

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
martnpaneq: It looks like it's not possible to get `cwd` from the `simple-git` instance.

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
```diff
@@ -39,6 +51,34 @@ export default async function commitAndTag( { version, files, cwd = process.cwd(
...
+	const git = simpleGit( { baseDir: normalizedCwd } );
+
+	try {
+		await git.add( filePathsToAdd );
+		await tools.shExec( preCommitCommand, { cwd, verbosity: 'silent', async: true } );
+	} catch ( e ) {
+		throw e.message;
+	} finally {
+		await git.reset( filePathsToAdd );
+	}
```
psmyrek, line 83: Two things to consider:

- Instead of executing `yarn lint-staged` command, maybe we could import the `lint-staged` package and call it in Node? Example: https://github.com/lint-staged/lint-staged?tab=readme-ov-file#can-i-use-lint-staged-via-node.
- There is `--diff` option to allow overriding how `lint-staged` gets changed files. If we could call `lint-staged` without adding the files to staging, we could get rid of `git reset` later. See https://github.com/lint-staged/lint-staged?tab=readme-ov-file#command-line-flags.

pomek: Therefore, we can avoid changes in `ckeditor5-dev`. Instead, we will modify the prepare-release scripts only.

pomek: Long story short, the `lint-staged` package always adds files to staging, so we cannot avoid executing the `git reset` command.

Let's improve our flow since we cannot find a proper flow using the existing packages, and I'd like to avoid using more than necessary.

When executing `commitAndTag( { dryRun: true } )`, let's do the following:

* Store the latest SHA-1 commit.
* Make a new commit.
* Ensure the tag is not created.
* Execute `git reset --soft {SHA-1}` to remove the committed changes (a revert without a trace in the history).
* Store again the latest SHA-1 commit. Compare both values. If they aren't equal, throw an error. It means that the `dryRun=true` mode is messed up.

pomek: According to https://stackoverflow.com/a/24569155/3388924, perhaps we should use the `mixed` mode.

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
```diff
@@ -39,6 +48,40 @@ export default async function commitAndTag( { version, files, cwd = process.cwd(
...
+async function executePreCommitHook( { filePathsToAdd, cwd } ) {
+	const normalizedCwd = toUnix( cwd );
+	const git = simpleGit( { baseDir: normalizedCwd } );
+
+	try {
+		const preDryRunCommit = await git.log( [ '-1' ] );
+		await git.commit( '[dry run] Release.', filePathsToAdd );
+
+		try {
+			await git.reset( [ preDryRunCommit.latest.hash ] );
```
martnpaneq, line 77: I decided to throw an error with a message informing about a potential potential problem with continuing the release in case `git reset` fails during a dry run.

I'm not sure if it's needed, as it probably will not ever happen.

pomek:
I would simplify the implementation:

```diff
diff --git a/packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js b/packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
index 1779897b..d2f39737 100644
--- a/packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
+++ b/packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
@@ -18,7 +18,6 @@ const { toUnix } = upath;
  * @param {string} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
  * @param {boolean} [options.skipCi=true] Whether to add the "[skip ci]" suffix to the commit message.
  * @param {boolean} [options.dryRun=false] In order to run pre commit checks in a dry run mode.
- * @param {ListrTaskObject} [options.listrTask={}] An instance of `ListrTask`.
  * @returns {Promise}
  */
 export default async function commitAndTag( {
@@ -26,8 +25,7 @@ export default async function commitAndTag( {
        files,
        cwd = process.cwd(),
        skipCi = true,
-       dryRun = false,
-       listrTask = {}
+       dryRun = false
 } ) {
        const normalizedCwd = toUnix( cwd );
        const filePathsToAdd = await glob( files, { cwd: normalizedCwd, absolute: true, nodir: true } );
@@ -48,38 +46,15 @@ export default async function commitAndTag( {
                return;
        }

-       if ( dryRun ) {
-               await executePreCommitHook( { filePathsToAdd, cwd } );
-               listrTask.output = `[dry run] Creating git tag v${ tagForVersion }.`;
-       } else {
-               await git.commit( `Release: v${ version }.${ skipCi ? ' [skip ci]' : '' }`, filePathsToAdd );
-               await git.addTag( `v${ version }` );
-       }
-}
-
-/**
- * Function that imitates running dry run for the `git commit` command.
- *
- * @param {object} options
- * @param {Array.<string>} options.filePathsToAdd File paths to run the dry run validations for.
- * @param {string} options.cwd Current working directory from which all paths will be resolved.
- * @returns {Promise<void>}
- */
-async function executePreCommitHook( { filePathsToAdd, cwd } ) {
-       const normalizedCwd = toUnix( cwd );
-       const git = simpleGit( { baseDir: normalizedCwd } );
+       await git.commit( `Release: v${ version }.${ skipCi ? ' [skip ci]' : '' }`, filePathsToAdd );

        try {
-               const preDryRunCommit = await git.log( [ '-1' ] );
-               await git.commit( '[dry run] Release.', filePathsToAdd );
+               if ( dryRun ) {
+                       const lastCommit = await git.log( [ '-1' ] );

-               try {
-                       await git.reset( [ preDryRunCommit.latest.hash ] );
-               } catch {
-                       throw new Error(
-                               'Running `git reset` failed during `git commit` dry run. The release might be in a broken state. ' +
-                               'Either fix it manually or prepare the release again.'
-                       );
+                       await git.reset( [ lastCommit.latest.hash ] );
+               } else {
+                       await git.addTag( `v${ version }` );
                }
        } catch ( error ) {
                throw error.message;
```

pomek:
The `try/catch` block isn't needed at all:

```
yarn run release:prepare-packages --compile-only --verbose
yarn run v1.22.19
$ node ./scripts/preparepackages.js --compile-only --verbose
[STARTED] Verifying the repository.
[SKIPPED] Verifying the repository.
[STARTED] Check the release directory.
[PROMPT] ? The release directory must be empty. Continue and remove all files? (Y/n)
[PROMPT] ✔ The release directory must be empty. Continue and remove all files? yes
[COMPLETED] Check the release directory.
[STARTED] Updating the `#version` field.
[SKIPPED] Updating the `#version` field.
[STARTED] Updating dependencies.
[SKIPPED] Updating dependencies.
[STARTED] Run the "build" command in `ckeditor5-*` packages.
(node:133492) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 abort listeners added to [AbortSignal]. MaxListeners is 10. Use events.setMaxListeners() to increase limit
(Use `node --trace-warnings ...` to show where the warning was created)
[OUTPUT] Status: 1/12.
[OUTPUT] Status: 2/12.
[OUTPUT] Status: 3/12.
[OUTPUT] Status: 4/12.
[OUTPUT] Status: 5/12.
[OUTPUT] Status: 6/12.
[OUTPUT] Status: 7/12.
[OUTPUT] Status: 8/12.
[OUTPUT] Status: 9/12.
[OUTPUT] Status: 10/12.
[OUTPUT] Status: 11/12.
[OUTPUT] Status: 12/12.
[COMPLETED] Run the "build" command in `ckeditor5-*` packages.
[STARTED] Copying `ckeditor5-dev-*` packages to the release directory.
[COMPLETED] Copying `ckeditor5-dev-*` packages to the release directory.
[STARTED] Cleaning-up.
[COMPLETED] Cleaning-up.
[STARTED] Commit & tag.
[FAILED] Just an error

Error: Just an error
    at Module.commitAndTag (file:///home/pomek/Projects/ckeditor/ckeditor5/external/ckeditor5-dev/packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js:52:9)
    at async Task.run (file:///home/pomek/Projects/ckeditor/ckeditor5/node_modules/listr2/dist/index.js:2058:11)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

pomek (changes requested):
Please review the following points:

* https://github.com/ckeditor/ckeditor5-dev/pull/1077#issuecomment-2703396190
	* https://github.com/ckeditor/ckeditor5-dev/pull/1077#issuecomment-2703545073
* Update the suggested commit message after the recent changes.

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
```diff
@@ -17,9 +17,16 @@ const { toUnix } = upath;
  * @param {Array.<string>} options.files Array of glob patterns for files to be added to the release commit.
  * @param {string} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
  * @param {boolean} [options.skipCi=true] Whether to add the "[skip ci]" suffix to the commit message.
+ * @param {boolean} [options.dryRun=false] In order to run pre commit checks in a dry run mode.
```
pomek, line 20:
```suggestion
 * @param {boolean} [options.dryRun=false] When enabled, the function creates a commit to allow executing a recommit hook, then removes it using the `git reset` command.
```

pomek (approved):
LGTM.

pomek (approved):

martnpaneq:
Feature (release-tools): Added the options.dryRun parameter to the commitAndTag() function to verify if a release commit passes validation, so releasing a project will not fail due to issues while committing.

PR #1087, opened 2025-03-25
Changed loaded files order

przemyslaw-zan:

Fix (build-tools): Redefined the order of loaded files so the TypeScript sources will take precedence over JavaScript files.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-build-tools/src/config.ts
+++ b/packages/ckeditor5-dev-build-tools/src/config.ts
@@ -85,7 +85,7 @@ export async function getRollupConfig( options: BuildOptions ) {
 	/**
 	 * Valid extensions for JavaScript and TypeScript files.
 	 */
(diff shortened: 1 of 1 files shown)


pomek (approved):
LGTM.

pomek (changes requested):
I check this PR carefully, and it does not work at all. What's even worse, it will never work because (my assumption is here) the resolver mechanism is not used. ESM requires passing an extension when importing a module, so it's always a valid path. Hence, the resolver does not have to check which extension should be used because a JavaScript file exists and can be used.

To prove it, I compiled the `ckeditor5-font` package and disabled minifying the output.

In the source, I modified it like this:

image

In the compiled JavaScript, I added a similar text:

image

Then, let's build the package.

image

PS: I executed `yarn postinstall` when switching to the branch, so I am using the built package with the proposed fix.

PR #1091, opened 2025-03-31
Support for typedoc@0.28

pomek:

Other (typedoc-plugins): Introduced the `validate()` function that checks if the converted documentation is error-free.

The following cases are verified:

* An event in the `@fires` tag exists.
* An identifier in the `@link` tag points to an existing doclet.
* A module name matches the path to the file where the module is defined.
* Overloaded methods and functions are described with the mandatory and unique `@label` tag.
* An identifier in the `@see` tag points to an existing doclet.

Feature (docs): Support for `typedoc@0.28`. The package uses the latest (`0.28.4`) TypeDoc release when introducing this change.

Other (typedoc-plugins): Improved processing parameters of the CKEditor 5 error definitions.

MAJOR BREAKING CHANGE (docs): The API docs builder requires `typedoc` in version 0.28 or higher.

MAJOR BREAKING CHANGE (typedoc-plugins): The TypeDoc plugins require the `typedoc` dependency in a version `0.28` or higher.

MAJOR BREAKING CHANGE (typedoc-plugins): CKEditor 5 events are stored in the `#ckeditor5Events` property on a class or interface declaration.

MAJOR BREAKING CHANGE (typedoc-plugins): CKEditor 5 errors can be recognized by the `#isCKEditor5Error` flag on a declaration.

MINOR BREAKING CHANGE (docs): The `strict` option passed as a modifier to the `build()` function should be wrapped into the `validatorOptions` object.

MINOR BREAKING CHANGE (typedoc-plugins): Parameters of CKEditor 5 events and errors are now stored in the `#parameters` property (instead of `#typeParameters`).

---

### Additional information


psmyrek (approved):
LGTM 🎉

PR #1092, opened 2025-04-01
Fix (build-tools): Prioritize `.ts` over `.js` files

filipsobol:

Fix (build-tools): Prioritize `.ts` over `.js` files.

---

This PR causes the `.ts` file to take precedence if both `.js` and `.ts` files are present on the file system. This prevents potentially old and obsolete build files from overriding current source files.

It also reverts the changes made in #1051 so that the TypeScript plugin is only used if the `declaration: true` option is passed. This improves the build time of:

* `ckeditor5-premium-features` by ~24s (from 65.5s down to 41.5s),
* documentation (`yarn docs --skip-validation --skip-api`) by ~20s (from 111s down to 89.5s).

### DIFF
--- a/packages/ckeditor5-dev-build-tools/src/config.ts
+++ b/packages/ckeditor5-dev-build-tools/src/config.ts
@@ -25,6 +25,7 @@ import { addBanner } from './plugins/banner.js';
 import { emitCss } from './plugins/emitCss.js';
 import { replaceImports } from './plugins/replace.js';
 import { splitCss } from './plugins/splitCss.js';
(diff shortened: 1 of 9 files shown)


psmyrek (approved):
I can't reproduce the original problem so not sure if this PR really fixes the issue.

Test case I checked: I built some packages in the Commercial repository (`ckeditor5-comments`, `ckeditor5-collaboration` and `ckeditor5-collaboration-core`) and then I built the documentation (`yarn run docs --production`). Everything is fine. Also, the build time is shorter.

LGTM 🎉

PR #1108, opened 2025-04-11
Feature (web-crawler): Improve performance.

filipsobol:

Internal (web-crawler): Increase the number of concurrent workers to equal the number of CPU threads, but no more than 16.

Internal (web-crawler): Properly read console errors.

Feature (web-crawler): Improve performance.

MINOR BREAKING CHANGE (web-crawler): Rename `noSpinner` to `silent`.

---

While the old implementation took about **45 to 50 minutes** to validate all uber docs, the new one does it in **less than 4 minutes** on my laptop.

```text
== Start:     2025-04-15 09:39:05.055
== Now:       2025-04-15 09:42:52.254 (running for 3.8 minutes)
== Progress:  3828 / 3828 (100.00%), errors: 0 (0.00%)
== Remaining: 0.0 ms (@ 16.85 pages/second)
== Sys. load: 92.1% CPU / 41.8% memory
== Workers:   16
```

Here's the breakdown of the changes and how they reduced the build time:

1. Base: **50 minutes**
2. Using `document.links` instead of `page.$$eval( ... )`: **13 minutes**
3. Increasing the worker count: **8.5 minutes**
4. Adding Chrome flags, blocking `api.json`, reading `x-cke-crawler-ignore-patterns` only when there's an error on the page: **6.5 minutes**
5. Changing `waitUntil` to `load` instead of `[ 'load', 'networkidle0' ]`: **3.8 minutes**

### DIFF
--- a/packages/ckeditor5-dev-web-crawler/src/getlinks.ts
+++ b/packages/ckeditor5-dev-web-crawler/src/getlinks.ts
@@ -1,0 +1,34 @@
+/**
+ * @license Copyright (c) 2003-2025, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* eslint-env browser */
+// eslint-disable-next-line spaced-comment
+/// <reference lib="dom" />
+
+/**
+ * Returns all links on the page.
+ *
+ * This function is in a separate file to allow using DOM APIs thanks to TypeScript
+ * and eslint magic comments above.
+ */
+export function getAllLinks( ignoreAttribute: string ): Array<string> {
+	return Array
+		// Get all links from the document.
+		.from( document.links )
+
+		// Filter out links with the given attribute, downloads and protocols other than http(s).
+		.filter( link =>
+			link.href &&
+			!link.hasAttribute( ignoreAttribute ) &&
+			!link.hasAttribute( 'download' ) &&
+			/^https?:$/.test( link.protocol )
+		)
+
+		// Convert to absolute URL.
+		.map( link => link.origin + link.pathname )
+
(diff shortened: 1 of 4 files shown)


filipsobol:
Now that we ignore errors from most external services we can also double the `DEFAULT_CONCURRENCY` (to equal the number of CPU threads) for even better performance:

```text
== Start:     2025-04-11 17:19:46.197
== Now:       2025-04-11 17:27:42.121 (running for 7.9 minutes)
== Progress:  3825 / 3825 (100.00%), errors: 0 (0.00%)
== Remaining: 0.0 ms (@ 8.04 pages/second)
== Sys. load: 29.8% CPU / 40.8% memory
== Workers:   16
```

psmyrek (dismissed):
LGTM 🎉

packages/ckeditor5-dev-web-crawler/src/getlinks.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * @license Copyright (c) 2003-2025, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/* eslint-env browser */
+// eslint-disable-next-line spaced-comment
+/// <reference lib="dom" />
```
pomek, line 8: TBD: To add `DOM` as a known library in tsconfig.

pomek: Self answer: It would allow the use of DOM API in a Node.js package. While it does not make sense for every file except this one, as `getlinks` is executed in the browser context.

PR #1111, opened 2025-04-17
Migrated TypeDoc validators to TypeScript and moved to `typedoc-plugins` package

psmyrek:

Other (typedoc-plugins): Introduced `validate()` function that checks if the converted documentation is error-free. Closes ckeditor/ckeditor5#18364.

The following cases are verified:

* An event in the `@fires` tag exists.
* An identifier in the `@link` tag points to an existing doclet.
* A module name matches the path to the file where the module is defined.
* Overloaded methods and functions are described with the mandatory and unique `@label` tag.
* Identifier in the `@see` tag points to an existing doclet.

Internal (docs): TypeDoc validators are no longer available. They have been moved to `@ckeditor/typedoc-plugins` package.

Internal (typedoc-plugins): All TypeDoc validators from `@ckeditor/ckeditor5-dev-docs` package have been converted to TypeScript and rewritten to be a full-fledged TypeDoc plugins.

---

### Additional information


pomek (approved):
LGTM.

PR #1112, opened 2025-04-17
Internal: Properly handle and filter `UNCAUGHT_EXCEPTION` errors.

filipsobol:

Internal: Properly handle and filter `UNCAUGHT_EXCEPTION` errors.

---

Errors detected on the page should be added to `pageErrors` array first, so they can be filtered out before being added to the root `errors` map. See changes made in #1108.

### DIFF
--- a/packages/ckeditor5-dev-web-crawler/src/runcrawler.ts
+++ b/packages/ckeditor5-dev-web-crawler/src/runcrawler.ts
@@ -194,7 +194,7 @@ export default async function runCrawler( options: CrawlerOptions ): Promise<voi
 			message: ( error as Error ).message || '(empty message)'
 		} ) );
 
-		page.on( ERROR_TYPES.UNCAUGHT_EXCEPTION.event, error => onError( {
+		page.on( ERROR_TYPES.UNCAUGHT_EXCEPTION.event, error => pageErrors.push( {
 			pageUrl: data.url,
 			type: ERROR_TYPES.UNCAUGHT_EXCEPTION,
 			message: error.message || '(empty message)'


packages/ckeditor5-dev-web-crawler/src/runcrawler.ts
```diff
@@ -194,7 +194,7 @@ export default async function runCrawler( options: CrawlerOptions ): Promise<voi
 			message: ( error as Error ).message || '(empty message)'
 		} ) );
 
-		page.on( ERROR_TYPES.UNCAUGHT_EXCEPTION.event, error => onError( {
+		page.on( ERROR_TYPES.UNCAUGHT_EXCEPTION.event, error => pageErrors.push( {
```
pomek, line 197: Am I right, the current production version is also affected?

https://github.com/ckeditor/ckeditor5-dev/blob/v47.1.1/packages/ckeditor5-dev-web-crawler/lib/runcrawler.js#L504-L508

If so, I would like to know if it wasn't done on purpose. Isn't it _uncaught_ meaning that it was unexpected?

@psmyrek, could you take a look?

filipsobol: No, it was just that we had two implementation of `onError` - one that added the errors to the global / shared `errors` map and another that added to the local `pageErrors` array.

I changed it in #1108 to make the distinction clear.

pomek: Hm. I think I got it.

https://github.com/ckeditor/ckeditor5-dev/blob/cde2cf9bf8754a3ad6be19849bb6b2b7c0200caa/packages/ckeditor5-dev-web-crawler/lib/runcrawler.js#L266-L271

pomek: How should we test this PR?

pomek:
How should we test this PR?

packages/ckeditor5-dev-web-crawler/src/runcrawler.ts
psmyrek: > @psmyrek, could you take a look?

This is a new implementation of error handling which differs from the first version of the crawler.

filipsobol:
> How should we test this PR?

Good question.

1. Checkout to this branch in `dev` repository.
2. Checkout to the `ck/18364-update-dependencies` in `core` and `commercial`.
3. Run `yarn docs`.
4. Run `yarn docs:serve`.
5. In separate terminal window run `yarn docs:verify`.

pomek (approved):
LGTM.

I used the following code in a built CKEditor snippet.

```js
var n = document.createElement( "meta" );
n.name = "x-cke-crawler-ignore-patterns";
n.content = JSON.stringify( {
	"uncaught-exception": [ "qwertyuiopasdfghjklzxcvbnm" ]
} );
document.head.appendChild( n );

throw new Error( 'qwertyuiopasdfghjklzxcvbnm' )
```

On `#master`:

image

On `#internal/4010-properly-filter-UNCAUGHT_EXCEPTION-errors`:

image

PR #1116, opened 2025-04-23
Added default manual tests preset

przemyslaw-zan:

Feature (tests): Manual tests script will now load a default predefined identity file, if it exists.

MAJOR BREAKING CHANGE (tests): Manual tests script will now load a default predefined identity file, if it exists.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/utils/automated-tests/parsearguments.js
+++ b/packages/ckeditor5-dev-tests/lib/utils/automated-tests/parsearguments.js
@@ -4,7 +4,7 @@
  */
 
 import fs from 'fs-extra';
-import path from 'path';
+import path from 'upath';
 import minimist from 'minimist';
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-dev-tests/bin/testmanual.js
```diff
@@ -9,7 +9,7 @@ import chalk from 'chalk';
 import path from 'path';
 import * as tests from '../lib/index.js';
 
-const options = tests.parseArguments( process.argv.slice( 2 ) );
+const options = tests.parseArguments( process.argv.slice( 2 ), { allowDefaultIdentityFile: true } );
```
filipsobol, line 12: Is there a specific reason to introduce a second parameter?

Couldn't we extend the `minimistConfig` in `parseArguments` itself to add this flag with some default value?

filipsobol (approved):

PR #1119, opened 2025-05-05
Fixed commits exceeding character limit

przemyslaw-zan:

Fix (release-tools): Fixed `commitAndTag()` throwing an `ENAMETOOLONG` error when trying to commit a large number of files on Windows.

---

### Additional information

### DIFF
--- a/packages/ckeditor5-dev-release-tools/tests/tasks/commitandtag.js
+++ b/packages/ckeditor5-dev-release-tools/tests/tasks/commitandtag.js
@@ -18,6 +18,7 @@ describe( 'commitAndTag()', () => {
 		stubs = {
 			git: {
 				tags: vi.fn(),
+				add: vi.fn(),
 				commit: vi.fn(),
 				reset: vi.fn(),
 				log: vi.fn(),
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-dev-release-tools/tests/tasks/commitandtag.js
```diff
@@ -169,12 +170,47 @@ describe( 'commitAndTag()', () => {
...
+		vi.mocked( glob ).mockResolvedValue(
+			new Array( numberOfFiles ).fill().map( ( _, i ) => `very/long/path/foo/bar/file_${ i }.js` )
+		);
+
+		await commitAndTag( { version: '1.0.0', packagesDirectory: 'packages', files: [ '**/package.json' ] } );
+
+		expect( vi.mocked( glob ) ).toHaveBeenCalledExactlyOnceWith( expect.anything(), expect.objectContaining( {
+			absolute: true,
+			nodir: true
+		} ) );
```
psmyrek, line 195: This assertion is not needed in this unit test, because (as the test name suggests), it verifies only if files are split into chunks.

psmyrek (approved):
Now it does not crash when committing the files:

```
✔ Clean up phase.
  ✔ Removing files that will not be published.
  ✔ Removing local typings.
✔ Commit & tag phase.
✔ Revert the update of release date (`--compile-only`).
```

LGTM 🎉

pomek (commented):
Why can't we use a more straightforward implementation that would add files to staging one by one and then commit them?

```js
for ( const file of files ) {
	git.add( file );
}

git.commit();
```

It does the same and does not complicate the simple process.

przemyslaw-zan:
> Why can't we use a more straightforward implementation that would add files to staging one by one and then commit them?

I ran each case twice for 992 files. Script with chunks set to 4000 characters runs for **~3.2s**, while one that adds each file separately runs for **~96s**. Separating files into chunks is a massive time save, its over 30 times faster.

pomek (approved):

pomek:
It's hard to imagine that a release commit updates 992 files (usually fewer than 100), but I am convinced I can handle the proposed solution.

PR #1122, opened 2025-05-14
Feature (web-crawler): Block requests to `if-cdn.com` domain.

filipsobol:

Feature (web-crawler): Block requests to `if-cdn.com` domain.

---

This domain seems to be a CDN for `iframe.ly` which we already block.

Tested on Uber docs.

### DIFF
--- a/packages/ckeditor5-dev-web-crawler/src/constants.ts
+++ b/packages/ckeditor5-dev-web-crawler/src/constants.ts
@@ -40,6 +40,7 @@ export const IGNORED_HOSTS = [
 	'challenges.cloudflare.com',
 	'binance.com',
 	'iframe.ly',
+	'if-cdn.com',
 	'jsfiddle.net',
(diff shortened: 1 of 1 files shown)


packages/ckeditor5-dev-web-crawler/src/constants.ts
```diff
@@ -40,6 +40,7 @@ export const IGNORED_HOSTS = [
 	'challenges.cloudflare.com',
 	'binance.com',
 	'iframe.ly',
+	'if-cdn.com',
```
pomek, line 43:
```suggestion
	// See: https://github.com/ckeditor/ckeditor5/issues/18519.
	'if-cdn.com',
```

To know why. That's why I added the new entry at the end.

pomek:
Closing in favor of https://github.com/ckeditor/ckeditor5-dev/pull/1123.

PR #1126, opened 2025-05-14
Enabled returning changelog instead of writing it

przemyslaw-zan:

Feature: Added `returnChangelog` option for the `generateChangelog()` function that makes the function return the changelog instead of writing it to a file. Closes ckeditor/ckeditor5#18064.

---

### Additional information

The option also prevents changesets from deletion, and handles version automatically instead of asking the user to pick one.

### DIFF
--- a/packages/ckeditor5-dev-changelog/src/generatechangelog.ts
+++ b/packages/ckeditor5-dev-changelog/src/generatechangelog.ts
@@ -26,11 +26,19 @@ import { getNewChangelog } from './utils/getnewchangelog.js';
 import { removeChangesetFiles } from './utils/removechangesetfiles.js';
 import { removeScope } from './utils/removescope.js';
 
+export async function generateChangelog(
+  config: RepositoryConfig & GenerateChangelog & { noWrite?: false | undefined }
+): Promise<void>;
+
+export async function generateChangelog(
+  config: RepositoryConfig & GenerateChangelog & { noWrite: true }
+): Promise<string>;
+
 /**
  * This function handles the entire changelog generation process including version management,
  * package information gathering, and changelog file updates.
  */
-export async function generateChangelog<T extends boolean | undefined>( {
+export async function generateChangelog( {
 	nextVersion,
 	cwd = process.cwd(),
 	packagesDirectory = PACKAGES_DIRECTORY_NAME,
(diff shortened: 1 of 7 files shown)


pomek:
I think this point from DoD is not implemented:

> A user can decide if the input files should be removed or should not be touched.

packages/ckeditor5-dev-changelog/src/generatechangelog.ts
```diff
@@ -40,12 +40,13 @@ export async function generateChangelog( {
 	date = format( new Date(), 'yyyy-MM-dd' ) as RawDateString,
 	changesetsDirectory = CHANGESET_DIRECTORY,
 	skipLinks = false,
-	singlePackage = false
-}: RepositoryConfig & GenerateChangelog ): Promise<void> {
+	singlePackage = false,
+	returnChangelog = false
```
pomek, line 44:
```suggestion
	noWrite = false
```

If `noWrite=true`, return instead of saving. Other proposal: `emitOnly`.

packages/ckeditor5-dev-changelog/src/generatechangelog.ts
```diff
@@ -40,12 +40,13 @@ export async function generateChangelog( {
...
+}: RepositoryConfig & GenerateChangelog ): Promise<string | undefined> {
```
pomek, line 45: It's a good example of using overloads:

```ts
type GenerateChangelogWithReturn = GenerateChangelogBase & {
  returnChangelog: true;
};

type GenerateChangelogWithoutReturn = GenerateChangelogBase & {
  returnChangelog?: false;
};


export function generateChangelog(
  config: RepositoryConfig & GenerateChangelogWithReturn
): Promise<string>;

export function generateChangelog(
  config: RepositoryConfig & GenerateChangelogWithoutReturn
): Promise<void>;
```

pomek (changes requested):
See above.

packages/ckeditor5-dev-changelog/src/utils/getnewversion.ts
```diff
@@ -43,6 +52,13 @@ export async function getNewVersion(
 		bumpType = 'major';
 	}
 
+	if ( noWrite ) {
+		return {
+			newVersion: semver.inc( oldVersion, bumpType ) || oldVersion,
+			isInternal: false
+		};
+	}
+
```
pomek, line 61: Why did you add these? The issue you work on does not say anything about updating versions. We aim to resolve a promise and prevent the removal of input files. Am I missing anything, or are the changes out of the scope?

przemyslaw-zan: Since we want to run those options in an automatic script, we need to prevent `inquirer` from running, since it would block the script.

pomek: For this purpose we have the `nextVersion` option.

pomek (approved):
LGTM.

PR #1130, opened 2025-05-16
Fix rendering values of the `@default` tag in API docs

psmyrek:

Fix (docs): Fixed rendering values of the `@default` tag in API docs. Previously, TypeDoc tried to guess whether the `@default` tag value should be treated as a code, which produced invalid outputs in some cases. Now it just takes the value from this tag and renders it as inline code. Closes ckeditor/ckeditor5#18526.

---

### Additional information

I checked a few `@default` tags and it seems they are now rendered correctly:
obraz
obraz
obraz
obraz
obraz
obraz
obraz

### DIFF
--- a/packages/ckeditor5-dev-docs/lib/build.js
+++ b/packages/ckeditor5-dev-docs/lib/build.js
@@ -36,6 +36,12 @@ export default async function build( config ) {
 		basePath: config.cwd,
 		readme: 'none',
 
+		jsDocCompatibility: {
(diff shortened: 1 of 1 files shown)


pomek (approved):
LGTM.

pomek:
> Is this targeted at `master` for a reason, or can we push it to `stable` safely?

`ckeditor5-dev` has no `#stable`. After the merge, we will push it to npm, and then we can rebuild the docs.

godai78:
pepeagent

PR #1132, opened 2025-05-20
Added validations for the changeset entries.

martnpaneq:

Feature (changelog): Added validations for the changeset entries: `type`, `scope`, `closes`, `see` and `breaking-change`. Closes ckeditor/ckeditor5#18326.

---

### Additional information


martnpaneq:
This is the current format of validations. Any formatting suggestions are welcome:
image

martnpaneq:
Final version of the formatting for the logs:
image

martnpaneq:
Example changesets for testing validations:

[changelog.zip](https://github.com/user-attachments/files/20361496/changelog.zip)

packages/ckeditor5-dev-changelog/src/generatechangelog.ts
```diff
@@ -25,20 +25,40 @@ import { getRepositoryUrl } from './utils/external/getrepositoryurl.js';
...
+	try {
+		return await main( options );
+	} catch ( error ) {
+		if ( error instanceof InternalError ) {
+			console.error( chalk.red( 'Error: ' + error.message ) );
+
+			process.exit( 1 );
+		} else {
+			throw error;
+		}
```
pomek, line 53:
```suggestion
		if ( !( error instanceof InternalError ) ) {
			throw error;
		}
	
		console.error( chalk.red( 'Error: ' + error.message ) );
		process.exit( 1 );
```

One fewer `{}` :D

packages/ckeditor5-dev-changelog/src/errors.ts
```diff
@@ -0,0 +1,22 @@
...
+			'No valid changesets found. Please verify that:\n' +
+			'1) Changesets are properly created in the "changesetsDirectory"\n' +
+			'2) The "cwd" parameter is set to the root of your project\n' +
+			'3) The "packagesDirectory" parameter correctly specifies the directory containing packages.\n' +
+			'If no errors are visible in the console, but changesets exist in your project, ' +
+			'this likely indicates the project is not configured correctly.\n' +
+			'If there are validation errors in the console, please fix them by following the validation details.\n';
+
+		super( message );
+	}
```
pomek, line 21:
```suggestion
```

This could be moved directly to the `catch()` part. Thanks to that, the `InternalError` can be reused.

packages/ckeditor5-dev-changelog/src/generatechangelog.ts
```diff
@@ -25,20 +25,40 @@ import { getRepositoryUrl } from './utils/external/getrepositoryurl.js';
...
 ): Promise<string>;
 
+/**
+ * Wrapper function that provides error handling for the changelog generation process.
+ */
+export async function generateChangelog(
+	options: RepositoryConfig & GenerateChangelog
+): Promise<string | void> { // eslint-disable-line @typescript-eslint/no-invalid-void-type
+	try {
+		return await main( options );
```
pomek, line 45:
```suggestion
		return main( options );
```

pomek: Sorry! The proposal changes how the `generateChangelog()` handles an error returned by the `main()` function. We must use `await`.

packages/ckeditor5-dev-changelog/src/generatechangelog.ts
filipsobol: Also consider borrowing the `removeWhitespace()` util from `build-tools`. With it, you can write this error message in string literal, and it will remove all whitespace characters after every new line. It looks much cleaner.

https://github.com/ckeditor/ckeditor5-dev/blob/a9bdc6572d0caa33b22608a722847f960e4267d1/packages/ckeditor5-dev-build-tools/src/plugins/translations.ts#L14-L21

packages/ckeditor5-dev-changelog/src/utils/getsectionswithentries.ts
```diff
@@ -76,86 +72,63 @@ function getIssuesLinks( issues: Array<string> | undefined, prefix: string, gitH
...
 	// If someone tries to use minor/major breaking change in a single package, we simply cast it to a generic breaking change.
-	if ( singlePackage && typeof breakingChange === 'string' ) {
-		return 'breaking';
-	}
-
-	if ( typeof breakingChange === 'string' ) {
-		return breakingChange;
-	}
+	if ( singlePackage ) {
+		if ( breakingChangeNormalized === 'minor' || breakingChangeNormalized === 'major' || breakingChangeNormalized === true ) {
```
filipsobol, line 110:
```suggestion
		if ( [ true, 'minor', 'major' ].includes( breakingChangeNormalized ) ) {
```

Nitpick, feel free to ignore.

packages/ckeditor5-dev-changelog/src/utils/getsectionswithentries.ts
```diff
@@ -76,86 +72,63 @@ function getIssuesLinks( issues: Array<string> | undefined, prefix: string, gitH
...
 
-	if ( breakingChange === true ) {
-		return 'breaking';
+		if ( breakingChangeNormalized === 'major' ) {
+			return 'major';
+		}
 	}
 
-	if ( type === 'Feature' ) {
+	if ( entry.data.typeNormalized === 'Feature' ) {
```
filipsobol, line 123: If I understand correctly, at this point we are already past validation, so we know the input is valid.

If so, we can replace these 3 `if` statements with `return entry.data.typeNormalized.toLowerCase()`.

filipsobol: Or maybe not because there's `return 'invalid';` after that. At this point, is it still possible for the `type` to be invalid?

pomek:
I found a serious issue with our current approach in the tool. Let's assume the following file as input:

```md
---
type: Feature
scope:
  - ckeditor5-utils
see:
  - ckeditor5#123								# a slug-like expression is on purpose
  - #123 										# `#` next to the issue number is on purpose
  - github.com/ckeditor/ckeditor5/issue/1132	# URL-like expression
---

Lorem Ipsum is simply dummy text of the printing and typesetting industry.
```

While the _see_ action includes invalid values, the remaining values are correct and can be easily transformed into a changelog entry. The previous approach (git based) would accept such a commit:

```
Feature (utils): Lorem Ipsum is simply dummy text of the printing and typesetting industry. See ckeditor5#123. See ##123. See github.com/ckeditor/ckeditor5/issue/1132.
```

The invalid parts would not be replaced with proper links. I must update them manually in the post-processing phase. The new approach allows me to fix the files and generate them again. 

**However, when preparing a nightly changelog** (https://github.com/ckeditor/ckeditor5/issues/14724), we take the input and do not update it. Hence, we need to improve the current solution:

* When a commit has an invalid type, treat the entire commit as invalid.
* When a commit has invalid references (see or close), ignore its value, but the entry should be processed.
* When a commit defines an invalid scope, ignore its value and process the entry.

This leads to modifying the output. We have three groups of entries now:

* Valid (correct type, references, and scope).
* Partially valid (correct type, references, and/or scope need to be fixed).
* Invalid (incorrect type).

I would suggest the following change to avoid redesigning the entire output (console). Add `+` or `!` based on status:

* Valid => green `+`
* Partially valid => yellow `!`

```
○ Listing the changes...
  ◌ Found Features:
  - + "Feature (ckeditor5-source-editing): The `formatHtml()` helper function is extracted to the `@ckeditor/ckeditor5-utils` package."

  ◌ Found Other changes:
  - ! "Other (ckeditor5-utils): The `formatHtml()` function is now exported by the `@ckeditor/ckeditor5-utils` package. It was previously a private function in the `ckeditor5-source-editing` package."

```

packages/ckeditor5-dev-changelog/src/generatechangelog.ts
martnpaneq: @filipsobol using `removeWhitespace` would be nice, but that would require to add `ckeditor5-dev-build-tools` as a dependency. Are we sure we want to do that just to use one util?

martnpaneq: @pomek I simplified the global error handling in [87252be](https://github.com/ckeditor/ckeditor5-dev/pull/1132/commits/87252beaf98bdd5010efcb11a5a346d38fc713d4). I left the message in `InternalError`, because I think this is not a bad solution. We could change the name of the error to e.g. `InvalidChangesets` or something similar. But in the current approach, we could add as many error types as we want, without increasing the size of the entry script `generateChangelog.ts`.

packages/ckeditor5-dev-changelog/src/utils/getsectionswithentries.ts
martnpaneq: At this point, we are sure, that the record is validated, and if it is invalid, it would fall in the first if:
image

So your suggestion makes sense.

martnpaneq:
How to test the change:

Add this code to the end of `generateChangelog.ts`:

```

const __filename = fileURLToPath( import.meta.url );
const __dirname = path.dirname( __filename );

generateChangelog( {
	cwd: upath.join( __dirname, '..', '..', '..', '..', '..' ),
	removeInputFiles: false,
	externalRepositories: [
		{
			cwd: upath.join( __dirname, '..', '..', '..', '..', '..', 'external', 'ckeditor5-commercial' )
		}
	]
} );

```

And run this function using anything that can understand TS, e.g.
```
npx tsx src/generatechangelog.ts
```

packages/ckeditor5-dev-changelog/src/errors.ts
pomek, line 1: The file name does not match the exported structure.

packages/ckeditor5-dev-changelog/src/utils/normalizeentry.ts
```diff
@@ -0,0 +1,61 @@
...
+
+	return {
+		...entry,
+		data: {
+			...entry.data,
+			typeNormalized,
+			breakingChangeNormalized,
+			scopeNormalized,
+			closesNormalized,
+			seeNormalized
```
pomek, line 52: Couldn't we drop the _normalized_ suffix from properties' names? It looks weird.

martnpaneq: Well, we need them to include original, user-provided input in the validation messages. Let's see this example:
```js
{
  type: 'FIXES',
  typeNormalized: 'Fix'
}
```
The validation message will look like this: 
```
 - Type "FIXES" should be one of: "Feature", "Other" or "Fix" ("Fixes" is allowed) (case insensitive).
```

In order to get rid of `*Normalized` suffix, we would have to override the original data (e.g. override `type` field). This is not a bad idea, but it could be confusing, to understand in the future. And we would loose the ability to reference the unmodified data in the future.

packages/ckeditor5-dev-changelog/src/utils/normalizeentry.ts
```diff
@@ -0,0 +1,61 @@
...
+		} else if ( breakingChangeLowerCase === 'major' ) {
+			breakingChangeNormalized = 'major';
+		}
+	} else {
+		if ( breakingChangeLowerCase === 'minor' ) {
+			breakingChangeNormalized = 'minor';
+		} else if ( breakingChangeLowerCase === 'major' ) {
+			breakingChangeNormalized = 'major';
+		}
+	}
```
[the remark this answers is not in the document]
martnpaneq: Actually this made me think, that normalization should not care about `singlePackage`. I'm making changes right now.

packages/ckeditor5-dev-changelog/src/errors/InternalError.ts
pomek, line 1: `internalerror.ts`

packages/ckeditor5-dev-changelog/src/utils/normalizeentry.ts
pomek: I am unsure if we need to display the invalid value. It's enough that we say "the provided one isn't supported". Update to xyz.

packages/ckeditor5-dev-changelog/src/utils/normalizeentry.ts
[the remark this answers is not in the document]
pomek: > We should decide if it is correct during a validation step.

It's the `type` validation, so we should check it. Also, we know how to map it:

* `singlePackage=true`
	* All kinds of breaking changes are just a breaking change.
* `singlePackage=false`
	* `'major'` => major breaking change
	* `'minor'` => minor breaking change
	* `true` => major breaking change

packages/ckeditor5-dev-changelog/src/utils/normalizeentry.ts
martnpaneq: Ok, I'm going to stop using `Normalized` and `Validated` suffixes, but I'm going to modify the original data.

packages/ckeditor5-dev-changelog/src/utils/normalizeentry.ts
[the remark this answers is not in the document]
martnpaneq: App was coded to behave in a way that is described above in [11aaf0f](https://github.com/ckeditor/ckeditor5-dev/pull/1132/commits/11aaf0f880ca4e141168d2eb6ffd4529e50968c8).

The logic is located in `validateentry.ts`.

PR #1137, opened 2025-05-22
Merged type and breaking change keys

przemyslaw-zan:

Other (changelog): The "breaking-change" key was removed from changeset entries. Now the "type" key also handles types "Minor" / "Major" / "Breaking". Closes ckeditor/ckeditor5#18550.

---

### Additional information


packages/ckeditor5-dev-changelog/src/utils/normalizeentry.ts
```diff
@@ -3,33 +3,24 @@
...
-		} else if ( breakingChangeLowerCase === 'minor' ) {
-			breakingChangeNormalized = 'minor';
-		} else if ( breakingChangeLowerCase === 'major' ) {
-			breakingChangeNormalized = 'major';
+import { TYPES } from '../constants.js';
+import type { NormalizedType, ParsedFile } from '../types.js';
+
+export function normalizeEntry( entry: ParsedFile ): ParsedFile {
+	// Normalize Type.
+	const typeCapitalized = capitalize( entry.data.type ) as NormalizedType;
```
martnpaneq, line 11: This type should not be cast to `NormalizedType`, because user can provide an invalid value, e.x. `Fixed`.

packages/ckeditor5-dev-changelog/src/constants.ts
```diff
@@ -31,3 +31,12 @@ export const ISSUE_SLUG_PATTERN = /^(?<owner>[a-z0-9.-]+)\/(?<repository>[a-z0-9
 export const ISSUE_PATTERN = /^\d+$/;
 export const ISSUE_URL_PATTERN =
 	/^(?<base>https:\/\/github\.com)\/(?<owner>[a-z0-9.-]+)\/(?<repository>[a-z0-9.-]+)\/issues\/(?<number>\d+)$/;
+
+export const TYPES = [
+	{ name: 'Feature' },
+	{ name: 'Other' },
+	{ name: 'Fix', aliases: [ 'Fixes' ] },
+	{ name: 'Major', aliases: [ 'Major breaking change' ] },
```
martnpaneq, line 39: I don't know if displaying `Major` is the best solution. `Major` is not a `type`, and we display it in the logs:
image

martnpaneq: Previously it looked like this. We were able to provide a `type`:
image

packages/ckeditor5-dev-changelog/src/constants.ts
```diff
@@ -31,3 +31,12 @@ export const ISSUE_SLUG_PATTERN = /^(?<owner>[a-z0-9.-]+)\/(?<repository>[a-z0-9
...
 	/^(?<base>https:\/\/github\.com)\/(?<owner>[a-z0-9.-]+)\/(?<repository>[a-z0-9.-]+)\/issues\/(?<number>\d+)$/;
+
+export const TYPES = [
+	{ name: 'Feature' },
+	{ name: 'Other' },
+	{ name: 'Fix', aliases: [ 'Fixes' ] },
+	{ name: 'Major breaking change', aliases: [ 'Major' ] },
+	{ name: 'Minor breaking change', aliases: [ 'Minor' ] },
+	{ name: 'Breaking change', aliases: [ 'Breaking' ] }
+] as const satisfies ReadonlyArray<{ name: string; aliases?: ReadonlyArray<string> }>;
```
pomek, line 42: Is there any reason it is so complicated? You could just add an empty `aliases` array to the remaining types. However, I would drop the concept of alias. What is supposed to do that?

pomek:
Could you resolve conflicts, please?

image

PR #1138, opened 2025-05-23
Commit the generated changelog when `noWrite=false`

pomek:

Feature (changelog): The generated changelog entries will be committed automatically, including the external repositories. Closes ckeditor/ckeditor5#18051.

Feature (utils): Added a new function `tools.commit()`.

---

### Additional information


pomek:
Tests for changes in `ckeditor5-dev-release-tools`.

- **Commercial:**

	```
	✔ Commit & tag phase.
	✔ Revert the update of release date (`--compile-only`).
	```

- **CKEditor 5 dev:**

	image
	image

packages/ckeditor5-dev-utils/src/git/commit.ts
```diff
@@ -0,0 +1,50 @@
...
+			await git.add( chunk );
+		}
+
+		return git.commit( message );
+	};
+
+	if ( dryRun ) {
+		const lastCommit = await git.log( [ '-1' ] );
+
+		await makeCommit();
```
martnpaneq, line 30: I don't remember why we want to commit the changes in dry run mode. Isn't it enough to just log the to be committed files?

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
```diff
@@ -6,13 +6,10 @@
 import upath from 'upath';
 import { glob } from 'glob';
 import { simpleGit } from 'simple-git';
-
-const { toUnix } = upath;
-
-const CHUNK_LENGTH_LIMIT = 4000;
+import { git } from '@ckeditor/ckeditor5-dev-utils';
```
martnpaneq, line 9: I'm loudly thinking, because this is quite a big change.

This looks tricky that we have a naming conflict between `simpleGit` and `git`. Maybe we should create a full wrapper for `simple-git` in `@ckeditor/ckeditor5-dev-utils`. This way:
1. We would not have a naming conflict.
2. We would have `simple-git` dependency defined in just one place.
3. We would be able to switch `simple-git` for any other package in the future.

packages/ckeditor5-dev-changelog/src/generatechangelog.ts
```diff
@@ -43,20 +44,36 @@ export async function generateChangelog( {
...
+
+	// TODO: This should be built-in `getExternalRepositoriesWithDefaults`.
 	const gitHubUrl = await workspaces.getRepositoryUrl( cwd, { async: true } );
+
 	const { version: oldVersion, name: packageName } = await workspaces.getPackageJson( cwd, { async: true } );
+
+	// TODO: It's an internal of `getNewChangelog()`.
 	const dateFormatted = getDateFormatted( date );
+
+	// TODO It should accept a single parameter: the normalized repositories array.
```
martnpaneq, line 73: I agree with this TODO, that `getChangesetFilePaths` should accept a simple array of prepared files.

packages/ckeditor5-dev-changelog/src/generatechangelog.ts
```diff
@@ -43,20 +44,36 @@ export async function generateChangelog( {
...
 	date = format( new Date(), 'yyyy-MM-dd' ) as RawDateString,
 	changesetsDirectory = CHANGESET_DIRECTORY,
 	skipLinks = false,
 	singlePackage = false,
+
+	// TODO: Merge `removeInputFiles` and `noWrite` options.
 	noWrite = false,
 	removeInputFiles = true
 }: RepositoryConfig & GenerateChangelog ): Promise<string | void> { // eslint-disable-line @typescript-eslint/no-invalid-void-type
+	// TODO: getExternalRepositoriesWithDefaults => `normalizeRepositories`.
```
martnpaneq, line 59: I agree with all other TODOs, but I'm not sure about this one. Providing default values is different than normallization, and this util only provides default values.

packages/ckeditor5-dev-changelog/src/utils/commitchanges.ts
```diff
@@ -0,0 +1,32 @@
...
+		// Copy to avoid modifying the original array.
+		const files = changesetPaths.slice( 0 );
+
+		if ( isRoot ) {
+			files.unshift( upath.join( cwd, CHANGELOG_FILE ) );
+		}
+
+		logInfo( `◌ Processing "${ cwd }".`, { indent: 1 } );
+
+		await git.commit( { cwd, message, files } );
```
martnpaneq, line 30: I got this error when trying to commit changes. Is it expected?
image

packages/ckeditor5-dev-release-tools/lib/tasks/commitandtag.js
pomek: I should add the `commit()` function to the `tools` namespace. This will clarify that the function comes from our packages instead of external (`simple-git`).

As for an abstract over Git, this would be a massive topic, as we still have many calls like `tools.shExec( 'git ...' )` in various places in our code. It looks like an unspecified future topic.

packages/ckeditor5-dev-changelog/src/utils/commitchanges.ts
pomek: Improvements in 0c0b4a6.

PR #1146, opened 2025-05-30
Added checkVersionMatch() script

przemyslaw-zan:

Feature (dependency-checker): Added `checkVersionMatch()` script which verifies if all dependencies versions defined in `package.json` are consistent. Closes ckeditor/ckeditor5#18579.

Internal: Updated dependencies across dev packages to use consistent version definitions. See ckeditor/ckeditor5#18579.

---

### Additional information


packages/ckeditor5-dev-dependency-checker/lib/checkversionmatch.js
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * @param {Object} options
+ * @param {Object.<String, String>} options.expectedDependencies
+ * @param {Array.<Object>} options.packageJsons
+ * @param {Object.<String, String>} options.pathMappings
+ * @param {Function} options.isCkeditor5Package
+ */
+function fixDependenciesVersions( { expectedDependencies, packageJsons, pathMappings, isCkeditor5Package } ) {
+	packageJsons
+		.filter( packageJson => packageJson.dependencies )
```
psmyrek, line 59: When a `package.json` contains only `devDependencies`, they will not be fixed, because there is `.filter( packageJson => packageJson.dependencies )`.

I know this function is moved as is from CKEditor 5 repository, but maybe it would be nice to fix it.

psmyrek (commented):
If I understood the issue correctly, we would like to introduce a new bin script for this new util. 

Also, I missed the requirement to use strict dependencies versions in the `ckeditor5-dev` packages, but this PR removed the caret ranges for each dependency. Is it confirmed that this is the way to go?

packages/ckeditor5-dev-dependency-checker/lib/checkversionmatch.js
```diff
@@ -0,0 +1,252 @@
...
+	console.log( chalk.green( '✅  All dependencies fixed!' ) );
+}
+
+/**
+ * @param {Object} options
+ * @param {Object.<String, String>} options.expectedDependencies
+ * @param {Function} options.devDependenciesFilter
+ * @param {Array.<Object>} options.packageJsons
+ */
+function checkDependenciesMatch( { expectedDependencies, packageJsons, devDependenciesFilter } ) {
```
psmyrek, line 103: Pedantic case: the order of `@param` do not match the order of arguments.

psmyrek:
The script seems to work as expected. However the CI is red. Could you align the TypeScript version to make it green again?

PR #1154, opened 2025-06-11
Fix changelog entry generation

filipsobol:
### 🚀 Summary

The template file for the changelog entries generator is included in the published package to avoid issues when using the `ckeditor5-dev-changelog-create-entry` binary script.

---

### 📌 Related issues

* Closes ckeditor/ckeditor5#18665.

### DIFF
--- a/.changelog/20250611114035_ck_18665_fix_changelog_generation.md
+++ b/.changelog/20250611114035_ck_18665_fix_changelog_generation.md
@@ -26,4 +26,4 @@ closes:
   - https://github.com/ckeditor/ckeditor5/issues/18665
 ---
 
-Fix changelog entry generation on published version of `dev-changelog` package.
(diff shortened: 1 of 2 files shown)


.changelog/20250611114035_ck_18665_fix_changelog_generation.md
```diff
@@ -0,0 +1,29 @@
...
+# Optional: Issues this change closes.
+# Format:
+# - {issue-number}
+# - {repo-owner}/{repo-name}#{issue-number}
+# - Full GitHub URL
+closes:
+  - https://github.com/ckeditor/ckeditor5/issues/18665
+---
+
+Fix changelog entry generation on published version of `dev-changelog` package.
```
pomek, line 29: I would go into the details of what was wrong:

> The template file for the changelog entries generator is included in the published package to avoid issues while using the `ckeditor5-dev-changelog-create-entry` binary script.

Other than that, in CKeditor we use _past_ form, so I would stick to that convention even if we disagree with it.

PR #1163, opened 2025-06-23
Ensure to exit the build process once Rollup is done

psmyrek:
### 🚀 Summary

Introduced a fix in all TypeScript packages to ensure to exit the build process once Rollup is done. Previously, building two dependent packages could end up hanging the build process.

---

### 📌 Related issues

* Closes ckeditor/ckeditor5#18668

---

### 💡 Additional information

I checked the generated `dist` directories in all TypeScript packages before and after this change. They all are the same so I assume `--forceExit` does not break or change the builds.

### DIFF
--- a/packages/ckeditor5-dev-build-tools/package.json
+++ b/packages/ckeditor5-dev-build-tools/package.json
@@ -63,7 +63,7 @@
     "vitest": "^3.1.1"
   },
   "scripts": {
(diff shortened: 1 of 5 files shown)


pomek (approved):
LGTM. I wasn't able to break the process after adding the new flag.

PR #1165, opened 2025-06-25
Add support for importing raw file content by appending `?raw` query parameter to the import path.

filipsobol:
### 🚀 Summary

Add support for importing raw file content by appending `?raw` query parameter to the import path.

---

### 📌 Related issues

See https://github.com/ckeditor/ckeditor5/issues/18745


pomek (approved):
LGTM.

PR #1166, opened 2025-06-25
[Backport to v43] Add support for importing raw file content by appending `?raw` query parameter to the import path.

filipsobol:
### 🚀 Summary

Add support for importing raw file content by appending `?raw` query parameter to import path.

---

### 📌 Related issues

See https://github.com/ckeditor/ckeditor5/issues/18745

---

### 💡 Additional information

:warning: :warning: :warning: 

This is a backport of #1165 to `#master-v43` branch, which still uses conventional commits. Remember to paste the following commit message when merging:

```markdown
Feature (build-tools): Add support for importing raw file content by appending ?raw query parameter to the import path. See ckeditor/ckeditor5#18745.
```

:warning: :warning: :warning:

### DIFF
--- a/packages/ckeditor5-dev-build-tools/src/config.ts
+++ b/packages/ckeditor5-dev-build-tools/src/config.ts
@@ -23,6 +23,7 @@ import typescriptPlugin from '@rollup/plugin-typescript';
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 import { addBanner } from './plugins/banner.js';
 import { emitCss } from './plugins/emitCss.js';
+import { rawImport } from './plugins/rawImport.js';
 import { replaceImports } from './plugins/replace.js';
 import { splitCss } from './plugins/splitCss.js';
 import { translations as translationsPlugin } from './plugins/translations.js';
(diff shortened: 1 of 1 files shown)


pomek (changes requested):
While the plugin works as expected, it's required to modify the `getRollupConfig()` function.

https://github.com/ckeditor/ckeditor5-dev/blob/0af452ed6e7fb5706431f2f1442ce7f0cf8c0522/packages/ckeditor5-dev-build-tools/src/config.ts#L117

I believe we should enable it by default.

pomek:
My test:

* Input: `import DARK_MODE_STYLES from '../theme/dark-mode.css?raw';`
* Output: `var DARK_MODE_STYLES = ":root { ...`

pomek:
In my TypeScript project I had to add the following declaration:

```
declare module '*?raw' {
	const classes: string;
	export default classes;
}
```

I don't think we can fix it automatically, because `tsc` needs to understand the imports before we start preparing the `dist` files.

pomek (approved):
LGTM.

PR #1170, opened 2025-06-27
Add the `timeout` configuration option to the `runCrawler` function

filipsobol:
### 🚀 Summary

Add the `timeout` configuration option to the `runCrawler` function and export the `DEFAULT_TIMEOUT` variable.

---

### 💡 Additional information

I'm testing whether we can use Docker executors instead of Machine executors in CircleCI, but I get timeouts on one of the heavy manual tests focused on testing performance. Currently, there is no way to override the default 15s timeout.


pomek (approved):
LGTM.

PR #1183, opened 2025-07-17
Added support for non-stable releases.

martnpaneq:
### 🚀 Summary

Added support for non-stable releases.

---

### 📌 Related issues

* Closes ckeditor/ckeditor5#18842

---

### 💡 Additional information


.changelog/20250717110708_ck_18842_changelog_pre_release.md
```diff
@@ -0,0 +1,9 @@
+---
+type: Feature
+scope:
+  - ckeditor5-dev-changelog
+closes:
+  - https://github.com/ckeditor/ckeditor5/issues/18842
+---
+
+Introduced support for non-stable releases in the changelog generating tool.
```
pomek, line 9: In the issue (https://github.com/ckeditor/ckeditor5/issues/18842), we described the entire process of how it works. Could we be more specific and explain what "Introduced support" means?

packages/ckeditor5-dev-changelog/src/utils/removechangelogentryfiles.ts
```diff
@@ -15,9 +17,16 @@ import type { ChangesetPathsWithGithubUrl } from '../types.js';
...
 	logInfo( `○ ${ chalk.cyan( 'Removing the changeset files...' ) }` );
 
-	await Promise.all(
-		entryPaths
-			.flatMap( repo => repo.filePaths )
-			.map( file => fs.unlink( file ) )
-	);
+	for ( const repo of entryPaths ) {
+		const changesetDirectory = upath.join( repo.cwd, CHANGESET_DIRECTORY );
+		const entries = await readdir( changesetDirectory, { withFileTypes: true } );
```
pomek, line 22: Why do we read anything from `fs` if we have `entryPaths`?

packages/ckeditor5-dev-changelog/src/utils/promptreleasetype.ts
```diff
@@ -0,0 +1,26 @@
...
+		{
+			type: 'list',
+			name: 'releaseType',
+			message: 'Select the release type?',
+			choices: [
+				{ name: 'Latest (stable) release', value: 'latest' },
+				{ name: 'Pre-release (alpha/beta/rc)', value: 'prerelease' }
+			]
+		}
+	] );
```
pomek, line 23: Make it a `y/n` question.

> Is it a pre-release? (y/N)

It would be enough to decide what to do.

packages/ckeditor5-dev-changelog/src/utils/generatechangelog.ts
```diff
@@ -64,11 +67,16 @@ const main: GenerateChangelogEntryPoint<GenerateChangelogConfig> = async options
 		shouldIgnoreRootPackage,
 		externalRepositories
 	} );
+
+	const releaseType = await promptReleaseType();
```
pomek, line 71: There is no need to ask when `nextVersion` is specified (as an option).

packages/ckeditor5-dev-changelog/src/utils/detectreleasechannel.ts
```diff
@@ -0,0 +1,30 @@
...
+ * For licensing, see LICENSE.md.
+ */
+
+import semver from 'semver';
+import { type ReleaseChannel } from '../types.js';
+
+/**
+ * Detects the release channel from a version string.
+ */
+export function detectReleaseChannel( version: string ): ReleaseChannel {
```
pomek, line 12: While the function supports four types of releases, it would be beneficial to include this limitation in the changelog entry.

pomek:
I have not run the changes yet, but based on my observation, they would fail anyway.

When moving files between directories, Git did not add the change automatically to the staged:

```
 D .circleci/config.yml
?? .circleci/newconfig.yml
```

This means that the tool will commit only the updated `CHANGELOG.md` file.

https://github.com/ckeditor/ckeditor5-dev/blob/41b1a47c75f2386dd13b489c0e281f34232bb2fd/packages/ckeditor5-dev-changelog/src/utils/generatechangelog.ts#L146-L149

`commitChanges()` receives the old paths (before they are moved).

.changelog/20250717110708_ck_18842_changelog_pre_release.md
martnpaneq: Updated description in [be7ac14](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/be7ac140d67cb518bc265fed7dd442aa69b93246).

packages/ckeditor5-dev-changelog/src/utils/promptreleasetype.ts
martnpaneq: Changed to `y/n` question in [e4f1e73](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/e4f1e73009f8cbfb20a274d322f2225d0c5bcd46).

packages/ckeditor5-dev-changelog/src/utils/removechangelogentryfiles.ts
martnpaneq: I added it to detect and remove directories, but I rolled it back to the original implementation, and directories are ignored. Fixed in [e4f1e73](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/e4f1e73009f8cbfb20a274d322f2225d0c5bcd46).

packages/ckeditor5-dev-changelog/src/utils/movechangelogentryfiles.ts
```diff
@@ -0,0 +1,39 @@
...
+		const changelogDir = upath.join( cwd, CHANGESET_DIRECTORY );
+		const targetPath = upath.join( changelogDir, targetDir );
+
+		await fs.ensureDir( targetPath );
+
+		for ( const filePath of filePaths ) {
+			const fileName = upath.basename( filePath );
+			const targetFilePath = upath.join( targetPath, fileName );
+
+			await fs.move( filePath, targetFilePath, { overwrite: true } );
```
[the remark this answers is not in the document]
martnpaneq: This is the most tricky thing in the whole PR. I had to use `simple-git` to add moved files to staging. It's ugly, but it works. We could think about a better solution if you want. Fixed in [e4f1e73](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/e4f1e73009f8cbfb20a274d322f2225d0c5bcd46).

packages/ckeditor5-dev-changelog/src/utils/findchangelogentrypaths.ts
```diff
@@ -14,18 +14,20 @@ type FindChangelogEntryPathsOptions = {
 	cwd: string;
 	externalRepositories: Array<RepositoryConfig>;
 	shouldSkipLinks: boolean;
+	includeAllChannels?: boolean;
```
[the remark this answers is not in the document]
martnpaneq: Fixed in [e4f1e73](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/e4f1e73009f8cbfb20a274d322f2225d0c5bcd46).

packages/ckeditor5-dev-changelog/src/utils/detectreleasechannel.ts
martnpaneq: Do you mean we should validate the input provided by the user?

packages/ckeditor5-dev-changelog/src/utils/movechangelogentryfiles.ts
```diff
@@ -9,31 +9,49 @@ import upath from 'upath';
...
 
+		const modifiedFilePaths: Array<string> = [];
+
 		for ( const filePath of filePaths ) {
 			const fileName = upath.basename( filePath );
 			const targetFilePath = upath.join( targetPath, fileName );
 
-			await fs.move( filePath, targetFilePath, { overwrite: true } );
+			await git.add( filePath );
+			fs.renameSync( filePath, targetFilePath );
```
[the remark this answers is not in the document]
martnpaneq: Changed in [560768b](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/560768bb6376f10cf258e127ab234e1e0b365889).

packages/ckeditor5-dev-changelog/src/utils/generatechangelog.ts
```diff
@@ -125,11 +135,20 @@ const main: GenerateChangelogEntryPoint<GenerateChangelogConfig> = async options
...
 
-	await removeChangelogEntryFiles( entryPaths );
+	let pathsToCommit = entryPaths;
+
+	// Handle changelog entry files based on release type.
+	if ( releaseType === 'latest' ) {
+		await removeChangelogEntryFiles( entryPaths );
+	} else {
+		const targetChannel = detectReleaseChannel( newVersion );
+		pathsToCommit = await moveChangelogEntryFiles( entryPaths, targetChannel );
```
pomek, line 145: To enable tracking all changes to files from the Git perspective, we should amend the `pathsToCommit` array instead of overriding it.

`pathsToCommit` must include the source files and the destination files.

While the results still operate on the same data, from the tool perspective, we have two collections:

1. Source files that have been moved to another directory,
2. The destination files required for incoming changelog generation.

It's Git's responsibility to commit files,

https://github.com/ckeditor/ckeditor5-dev/blob/79ed963ec28b28b17a08f01072d31449958ca52e/packages/ckeditor5-dev-utils/src/tools/commit.ts#L29-L35

martnpaneq:
> Tool again suggests 51.0.0-alpha.0 which is already taken.

@psmyrek Tool suggests the version read from main `package.json`, and generating a new changelog does not modify it (only release does it).

martnpaneq:
We talked with @psmyrek, and agreed that each pre-release bump, should include all of the changes from the previous pre-release.

pomek:
@martnpaneq, see DoD, please.

.changelog/20250717110708_ck_18842_changelog_pre_release.md
```diff
@@ -0,0 +1,18 @@
...
+Introduce support for non-stable releases in the changelog generating tool.
+
+The following flow has been implemented to ensure clarity during the preparation process:
+
+- A new interactive prompt now asks users to specify whether the release is a _pre-release_ (e.g., `alpha`, `beta`, `rc`), _pre-release_ promotion or a _stable_ release.
+- Changelog entries are **no longer deleted** during non-stable releases. Instead, they are **moved to a pre-release subdirectory** inside `.changelog/`.
+- When continuing within the same channel (e.g., multiple alphas), only top-level `.md` files are included.
+- When transitioning to a **new channel**, tool collects entries from **pre-release directory**, ensuring no loss entries.
+- Stable (latest) releases now aggregate all prior changelog entries across cycles, and remove all of them after generating the changelog to start fresh.
+  Thanks to that, the generated changelog provides **clear traceability** of what was shipped during each phase, and reduces the chance of human error through improved automation and prompts.
```
pomek, line 18:
```suggestion

Thanks to that, the generated changelog provides **clear traceability** of what was shipped during each phase, and reduces the chance of human error through improved automation and prompts.
```

packages/ckeditor5-dev-changelog/src/utils/getreleasetypefromversion.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * @license Copyright (c) 2003-2025, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import { type ChangelogReleaseType } from '../types.js';
+import semver from 'semver';
+
+export function getReleaseTypeFromVersion( currentVersion: string, nextVersion: string ): ChangelogReleaseType {
```
pomek, line 9: I would rename it to `getReleaseType()`. It would be enough.

packages/ckeditor5-dev-changelog/src/utils/getreleasetypefromversion.ts
```diff
@@ -0,0 +1,22 @@
...
+
+	if ( !nextVersionPrerelease ) {
+		return 'latest';
+	}
+
+	if ( nextVersionPrerelease[ 0 ] === currentVersionPrerelease?.[ 0 ] ) {
+		return 'prerelease';
+	}
+
+	return 'prerelease-promote';
```
pomek, line 21: To avoid optional chaining operators, which, in my opinion, can be challenging to understand, I would apply default values and reorganize the logic slightly.

```suggestion
	const [ currentChannel ] = semver.prerelease( currentVersion ) || [ 'latest' ];
	const [ nextChannel ] = semver.prerelease( nextVersion ) || [ 'latest' ];

	if ( nextChannel === 'latest' ) {
		return 'latest';
	}

	if ( nextChannel === currentChannel ) {
		return 'prerelease';
	}

	return 'prerelease-promote';
```

packages/ckeditor5-dev-changelog/src/utils/movechangelogentryfiles.ts
pomek, line 1: Once tests are corrected, we can do some magic using `AsyncArray`.

* https://github.com/ckeditor/ckeditor5-dev/pull/1183#discussion_r2223321476
* https://github.com/ckeditor/ckeditor5-dev/pull/1183#discussion_r2223323937

```ts
export async function moveChangelogEntryFiles(
	entryPaths: Array<ChangesetPathsWithGithubUrl>
): Promise<Array<ChangesetPathsWithGithubUrl>> {
	const targetDir = PRE_RELEASE_DIRECTORY;

	logInfo( `○ ${ chalk.cyan( `Moving changelog entries to ${ targetDir }/ directory...` ) }` );

	return new Promise( ( resolve, reject ) => {
		return new AsyncArray( Promise.resolve( entryPaths ) )
			.map( async repo => {
				const { cwd, filePaths } = repo;
				const changelogDir = upath.join( cwd, CHANGESET_DIRECTORY );
				const targetPath = upath.join( changelogDir, targetDir );

				await fs.ensureDir( targetPath )
					.catch( reject );

				return {
					...repo,
					filePaths: await new AsyncArray( Promise.resolve( filePaths ) )
						.map( async filePath => {
							const fileName = upath.basename( filePath );
							const targetFilePath = upath.join( targetPath, fileName );

							await fs.rename( filePath, targetFilePath )
								.catch( reject );

							return [ targetFilePath, filePath ];
						} )
						.flat()
				};
			} )
			.then( resolve );
	} );
}
```

The `Promise` wrapper is needed because `AsyncArray` does not handle errors well. Hence, when operations on `fs.*` fail, we catch an error and reject the main promise.

pomek:
@psmyrek, I think it's an important case. Good to catch it. The opening question defines the behavior for the next layers, so we must ensure the provided version meets the expectations.

When selecting:

* `latest` => validator must accept `X.Y.Z` versions only,
* `prerelease` / `prerelease-promote` => validator must accept a prerelease version, and `X.Y.Z` is invalid.

packages/ckeditor5-dev-changelog/src/utils/movechangelogentryfiles.ts
martnpaneq: As far as I can see, we could not add any `try/catch` for `fs` operations, and let the error thrown naturally to interrupt the script.

martnpaneq: I did not use AsyncArray, because simple double `for of` loop implementation is shorter and easier to read.

packages/ckeditor5-dev-changelog/tests/utils/movechangelogentryfiles.ts
```diff
@@ -0,0 +1,191 @@
...
+
+		await expect( moveChangelogEntryFiles( mockEntryPaths ) )
+			.rejects.toThrow( 'Directory creation failed' );
+	} );
+
+	it( 'should handle fs.renameSync errors', async () => {
+		const error = new Error( 'File rename failed' );
+		vi.mocked( fs.rename ).mockImplementationOnce( () => {
+			throw error;
+		} );
```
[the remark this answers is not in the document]
martnpaneq: Fixed in [dd6e559](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/dd6e559bd34fc6ff9a3a0e08222ca410cc60764f)

packages/ckeditor5-dev-changelog/tests/utils/movechangelogentryfiles.ts
```diff
@@ -0,0 +1,191 @@
...
+	];
+
+	const mockGit = {
+		add: vi.fn().mockResolvedValue( undefined )
+	};
+
+	beforeEach( () => {
+		vi.clearAllMocks();
+		vi.mocked( fs.ensureDir ).mockResolvedValue();
+		vi.mocked( fs.renameSync ).mockImplementation( () => {} );
```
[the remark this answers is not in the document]
martnpaneq: Fixed in [dd6e559](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/dd6e559bd34fc6ff9a3a0e08222ca410cc60764f)

packages/ckeditor5-dev-changelog/tests/utils/getreleasetypefromversion.ts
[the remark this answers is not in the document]
martnpaneq: Added in [dd6e559](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/dd6e559bd34fc6ff9a3a0e08222ca410cc60764f)

packages/ckeditor5-dev-changelog/src/utils/getreleasetypefromversion.ts
martnpaneq: Added in [dd6e559](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/dd6e559bd34fc6ff9a3a0e08222ca410cc60764f)

packages/ckeditor5-dev-changelog/tests/utils/movechangelogentryfiles.ts
pomek, line 1: I can still find `simple-git` occurrences even if we removed it from `package.json`.

psmyrek (commented):
Now it does not work when trying to create a pre-release when the previous one was the latest release.

1. Go to `ckeditor5-dev`. The version is `50.3.1`.
2. Create changelog: `yarn run release:prepare-changelog`.
3. Choose the "pre-release" release type.
4. Accept suggested version: `51.0.0-alpha.0`.

    ```
    file:///D:/Projects/ckeditor/ckeditor5/external/ckeditor5-dev/packages/ckeditor5-dev-changelog/dist/index.js:627
                    if (releaseType === 'prerelease' && currentPrerelease[0] !== prerelease[0]) {
                                                                         ^
    
    TypeError: Cannot read properties of null (reading '0')
        at validate (file:///D:/Projects/ckeditor/ckeditor5/external/ckeditor5-dev/packages/ckeditor5-dev-changelog/dist/index.js:627:70)
        at async Object.<anonymous> (file:///D:/Projects/ckeditor/ckeditor5/node_modules/@inquirer/input/dist/esm/index.js:21:80)
    ```

martnpaneq:
Internal should be supported only for the continuation of current release (in case of a pre-release it should be the same channel):
1. Latest internal release: latest -> latest.
2. Prerelease continuation internal release: alpha -> alpha.

packages/ckeditor5-dev-changelog/src/utils/providenewversion.ts
```diff
@@ -133,8 +136,40 @@ function createVersionQuestion( options: Options ): Array<Question> {
 				return 'Given version is already taken.';
 			}
 
+			const prerelease = semver.prerelease( newVersion )!;
+			const currentPrerelease = semver.prerelease( version )!;
+
+			if ( ( releaseType === 'prerelease-promote' || releaseType === 'prerelease' ) && !prerelease ) {
+				return 'You chose a prerelease release path. Please provide a version with a channel suffix.';
```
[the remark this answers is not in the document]
martnpaneq: Fixed in [f6f4644](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/f6f4644dcb5ef94cf5b563ea267a00f38b9880ae).

packages/ckeditor5-dev-changelog/tests/utils/movechangelogentryfiles.ts
martnpaneq: Removed in [f6f4644](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/f6f4644dcb5ef94cf5b563ea267a00f38b9880ae).

packages/ckeditor5-dev-changelog/src/utils/providenewversion.ts
```diff
@@ -111,11 +114,18 @@ function createVersionQuestion( options: Options ): Array<Question> {
 		message,
 		filter: ( newVersion: string ) => newVersion.trim(),
 		async validate( newVersion: string ): Promise<VersionValidationResult> {
+			const [ newChannel ] = semver.prerelease( newVersion ) || [ 'latest' ];
```
pomek, line 117:
```ts
		async validate( newVersion: string ): Promise<VersionValidationResult> {
			const isInternal = newVersion === 'internal';

			const [ newChannel ] = semver.prerelease( newVersion ) || [ 'latest' ];
			const [ currentChannel ] = semver.prerelease( version ) || [ 'latest' ];

			// Special‑case: internal releases.
			if ( isInternal ) {
				if ( releaseType === 'latest' && currentChannel === 'latest' ) {
					return true;
				}
				return 'Internal release is only allowed on the latest channel.';
			}

			// Generic semantic‑version checks.
			if ( !semver.valid( newVersion ) ) {
				return 'Please provide a valid version or "internal" for internal changes.';
			}

			if ( !semver.gt( newVersion, version ) ) {
				return `Provided version must be higher than "${ version }".`;
			}

			if ( !( await npm.checkVersionAvailability( newVersion, packageName ) ) ) {
				return 'Given version is already taken.';
			}

			// Rules that depend on release type.
			const isPrerelease = releaseType === 'prerelease';
			const isPrereleasePromote = releaseType === 'prerelease-promote';
			const isLatest = releaseType === 'latest';

			// Pre‑release types must always include a channel suffix.
			if ( ( isPrerelease || isPrereleasePromote ) && newChannel === 'latest' ) {
				return 'You chose the "pre-release" release type. Please provide a version with a channel suffix.';
			}

			// Promoting a pre‑release: new version ≥ suggested version.
			if ( isPrereleasePromote && !semver.gte( newVersion, suggestedVersion ) ) {
				return `Provided version must be higher or equal to ${ suggestedVersion }.`;
			}

			// Continuing a pre‑release stream: channel cannot change.
			if (
				isPrerelease &&
				currentChannel !== 'latest' &&
				currentChannel !== newChannel
			) {
				return `Provided channel must be the same existing channel ${ currentChannel }.`;
			}

			// Latest release must not carry a channel suffix.
			if ( isLatest && newChannel !== 'latest' ) {
				return 'You chose the "latest" release type. Please provide a version without a channel suffix.';
			}

			return true;
		},
```

This version simplifies the logic a bit. I am open to suggestions or tweaks.

martnpaneq: Extracted the file that you suggested to a separate function: [1c9ebbd](https://github.com/ckeditor/ckeditor5-dev/pull/1183/commits/1c9ebbd0b2910a886c77d748adc56de9246a94d4)

psmyrek:
It is not possible to prepare an internal release when there are no changelog entries in the `.changelog` directory. Empty `.changelog` directory is expected for an internal release, because it will never have changelog entries to publish.

Steps to reproduce:

1. Go to `ckeditor5-dev`. The version is `50.3.1`.
2. Remove all changelog entries from the `.changelog` directory.
3. Create changelog: `yarn run release:prepare-changelog`.
4. Choose the "latest" release type.
5. It is not possible to provide the `internal` version, because the tool detects that there are no changelog entries.

```
ℹ️  No changelog entries found, so there is nothing to prepare a changelog from.
```

psmyrek:
Another case: the tool suggests the `internal` version in cases when it is not supported.

Steps to reproduce:

1. Go to `ckeditor5-dev`. The version is `50.3.1`.
2. Create changelog: `yarn run release:prepare-changelog`.
3. Choose the "pre-release" release type.
4. The tool suggests the internal version (among others), but it is not supported in this scenario.

    ```
    ? Type the new version (current: "50.3.1", suggested: "51.0.0-alpha.0", or "internal" for internal changes): internal
    > Internal release is only allowed on the latest channel.
    ```

PR #1185, opened 2025-07-25
Removed internal release support.

martnpaneq:
### 🚀 Summary

Disallowed "internal" as a valid release type.

The "internal" version identifier is no longer accepted to enforce consistent and meaningful versioning.
Developers must now provide a clear reason for publishing a release. Attempts to use "internal" will result in a validation error.

---

### 📌 Related issues

* Closes ckeditor/ckeditor5#18894

---

### 💡 Additional information


martnpaneq:
Currently, we do not validate input provided by param nextVersion. So any version, including internal would be approved. Should we add validations to `nextVersion`? Recently we extracted a util `validateInputVersion`, so it looks like we should just call it on `nextVersion`.

martnpaneq:
Here is a followup ticket for `nextVersion` validation: https://github.com/ckeditor/ckeditor5/issues/18901

PR #1191, opened 2025-08-07
Don't hardcode the path to the `@ckeditor/ckeditor5-utils` package.

filipsobol:
### 🚀 Summary

Don't hardcode the path to the `@ckeditor/ckeditor5-utils` package.

---

### 📌 Related issues

See https://github.com/ckeditor/ckeditor5-internal/issues/4083

### DIFF
--- a/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
+++ b/packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
@@ -4,18 +4,20 @@
  */
 
 import path from 'path';
-import getWebpackConfigForAutomatedTests from './getwebpackconfig.js';
 import { createRequire } from 'module';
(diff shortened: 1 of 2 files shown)


packages/ckeditor5-dev-tests/lib/utils/automated-tests/getkarmaconfig.js
```diff
@@ -28,6 +26,10 @@ export default function getKarmaConfig( options ) {
 
 	const basePath = process.cwd();
 	const coverageDir = path.join( basePath, 'coverage' );
+	const utilsAssets = path.resolve(
+		resolvePath( '@ckeditor/ckeditor5-utils/package.json', { paths: [ basePath ] } ),
```
martnpaneq, line 30: Just out of curiosity. Why do we need to resolve path to `package.json`?

martnpaneq (approved):

PR #1197, opened 2025-08-14
Migrate to pnpm

filipsobol:
### 🚀 Summary

This PR migrates the repository to pnpm. Because some production code now runs `pnpm` instead of `yarn`, this is a **major breaking change**.

---

### 📌 Related issues

* See: https://github.com/ckeditor/ckeditor5-internal/issues/4083.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@
 .nyc_output/
 coverage/
 npm-debug.log
-yarn.lock
 executeinparallel-integration.log
 
 # Prepared files to publish to npm.


packages/ckeditor5-dev-tests/bin/testautomated.js
```diff
@@ -12,11 +12,7 @@ import * as tests from '../lib/index.js';
 const options = tests.parseArguments( process.argv.slice( 2 ) );
 
 if ( options.files.length === 0 ) {
-	if ( options.cwd.endsWith( 'ckeditor5' ) ) {
-		options.files = [ '*' ];
-	} else {
-		options.files = [ '/' ];
-	}
+	options.files = [ '*' ];
```
filipsobol, line 15: This is needed to be able to run `pnpm run test` without any parameters from a commercial repository and still run all tests.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,13 @@
+packages:
+  - /*
+  - packages/*
```
psmyrek, line 3: The `/*` path causes that pnpm scans all directories in the filesystem root. On Windows, it ends with `EPERM: operation not permitted, scandir 'D:\System Volume Information'` to which pnpm has no access.

It seems that the root package is always included, even when custom location wildcards are used.

```suggestion
  - packages/*
```

filipsobol: Fixed

psmyrek (dismissed):

PR #1214, opened 2025-10-03
Using `workspace:*` for dependencies in `checkVersionMatch` function.

martnpaneq:
### 🚀 Summary

Updated the `checkVersionMatch()` function to support the [`workspace:*`](https://pnpm.io/workspaces) protocol for dependencies.

Added a `workspacePackages` option that specify the list of packages that should use `workspace:*` instead of specific version numbers. Both `dependencies` and `devDependencies` are validated to ensure consistent workspace versioning with pnpm.

Switched to using `workspace:*` in monorepo workspace, so that the lock file is not broken after a release.

---

### 📌 Related issues

* See ckeditor/ckeditor5-commercial#8556

---

### 💡 Additional information

### DIFF
--- a/packages/ckeditor5-dev-dependency-checker/lib/checkversionmatch.js
+++ b/packages/ckeditor5-dev-dependency-checker/lib/checkversionmatch.js
@@ -11,6 +11,8 @@ import semver from 'semver';
 import { globSync } from 'glob';
 import { execSync } from 'child_process';
 
+const PNPM_WORKSPACE_VERSION = 'workspace:*';
+
 const DEPENDENCY_TYPES = [
 	'dependencies',
 	'devDependencies'
@@ -160,6 +162,12 @@ function getExpectedDepsVersions( { packageJsons, devDependenciesFilter, version
 
 			Object.entries( packageJson[ dependencyType ] ).forEach( ( [ dependencyName, version ] ) => {
 				if ( dependencyType === 'devDependencies' && !devDependenciesFilter( dependencyName ) ) {
+					return;
+				}
+
+				if ( devDependenciesFilter( dependencyName ) ) {
+					expectedDependencies[ dependencyName ] = PNPM_WORKSPACE_VERSION;
+
 					return;
 				}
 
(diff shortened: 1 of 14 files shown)


martnpaneq:
The CI validations need to be fixed, and probably a new property is needed to select what packages need to be marked as part of the workspace. `devDependenciesFilter` should probably not be reused.

packages/ckeditor5-dev-dependency-checker/lib/checkversionmatch.js
```diff
@@ -149,16 +155,23 @@ function checkDependenciesMatch( { packageJsons, devDependenciesFilter, expected
...
 			if ( !packageJson[ dependencyType ] ) {
 				return;
 			}
 
 			Object.entries( packageJson[ dependencyType ] ).forEach( ( [ dependencyName, version ] ) => {
+				if ( useWorkspace && devDependenciesFilter( dependencyName ) ) {
+					expectedDependencies[ dependencyName ] = PNPM_WORKSPACE_VERSION;
+
+					return;
+				}
```
psmyrek, line 173: I think we cannot use `devDependenciesFilter()` callback to decide which packages require `workspace:*` version. This callback is designed to work only with dev dependencies, but when we set `useWorkspace: true`, we also want to update normal dependencies as well. It works currently, but the naming does not match the purpose.

martnpaneq: Updated here [992e410](https://github.com/ckeditor/ckeditor5-dev/pull/1214/commits/992e410fd63581b30387b9970031c8187dcc66cb).

Last thought to consider. Should `workspacePackages` be an array of strings or a callback similar to `devDependenciesFilter`?

It is an array right now, because I went with a simpler implementation.

filipsobol:
I haven't checked all the code yet, but I noticed an issue in the screenshot you shared above.

The `workspace:*` value was replaced with `^54.0.0`, but according to the [pnpm docs](https://pnpm.io/workspaces#publishing-workspace-packages), it should be replaced with `54.0.0`, because `*` indicates a strict version rather than a range. This suggests that the current implementation doesn’t rely on pnpm to replace these values and that we’re handling it manually instead.

If possible, we should let pnpm update all these values during `pnpm publish`. This way, we wouldn’t be reinventing the wheel, and in the future, we could also take advantage of [`publishConfig`](https://pnpm.io/package_json#publishconfig) to update the `"main": "src/index.ts"` values automatically instead of handling it ourselves.

PR #1299, opened 2026-02-18
Migrate the repository tooling to Rolldown and isolated declarations.

filipsobol:
### 🚀 Summary

Migrate the repository tooling to Rolldown and isolated declarations.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-internal/issues/4293.

---

#### No changelog

There are no changelog entries, as this is an internal-only change and should not affect the consumers.

#### Explanation for number of changes

There are more changes than necessary, but that's because:
* I renamed numerous files, and Git didn't register it properly, so instead of one rename, there are two changes for some files (deletion + addition).
* Because some tests are written in TypeScript, I configured TS to typecheck them too, which previously didn't happen. This forced me to fix some typings in tests too.

#### Findings

* Similar migrations should be fairly easy to implement in CKE5 repositories.

#### How to test

In the repository root, run the following commands:

* `pnpm run -r build` — run `build` command in all packages that have this command.
* `pnpm run typecheck` — run a single TypeScript check for all `.ts` files in the repository.

In the CKE5 repository (on the `master-v48` branch), run the following commands:
* `pnpm run nice`
* `pnpm run docs`
* `pnpm run docs:verify`
* `pnpm run release:prepare-packages --compile-only`
* `pnpm run check-dependencies`
* `pnpm run manual`

---

> [!NOTE]
> **Medium Risk**
> Build pipeline and TypeScript configuration changes affect how packages are bundled and how declarations are produced, which can break publishing or CI if Rolldown/isolated declarations behave differently. Runtime logic changes are small but touch crawler error handling and should be validated with existing tests and builds.
> 
> **Overview**
> Switches multiple internal tooling packages from `rollup` to `rolldown` by replacing `rollup.config.js` with `rolldown.config.js`, updating package scripts/deps, and dropping per-package `tsconfig.json` files.
> 
> Adds `scripts/plugin-declarations.js` (tested in `scripts-tests/plugin-declarations.mjs`) to generate `.d.ts` assets via `oxc-transform` isolated declarations, and updates the root `tsconfig.json` to enforce repo-wide strict typechecking (`isolatedDeclarations`, `noEmit`, expanded includes/excludes). CI is updated to run `pnpm run typecheck`, and the dynamic CircleCI config generator no longer injects per-package `types` steps.
> 
> Includes follow-up typing fixes in tests and runtime code (e.g., web-crawler event handlers and error message extraction) to satisfy the stricter typechecking.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d080d01379164d567081d3452b7166315d6b3d06. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/ckeditor5-dev-build-tools/package.json
```diff
@@ -57,13 +57,13 @@
     "@types/node": "^22.14.1",
     "@types/postcss-mixins": "^9.0.5",
     "@vitest/coverage-v8": "^4.0.18",
+    "rolldown": "^1.0.0-rc.4",
```
pomek: While I can accept isolated declarations, I am opposed to using unstable dependencies.

pomek: I propose to merge isolated declarations if this PR focuses only on them.

Once `rolldown` officially releases v1 (https://voidzero.dev/posts/announcing-rolldown-rc#roadmap-to-1-0-and-vite-8), we can adopt it in this repository.

pomek:
Is this change a breaking change?

filipsobol:
> Is this change a breaking change?

I don't think so. While the bundles might be slightly different, the APIs we already expose should stay intact. We don't fundamentally change the APIs, so if something stops working, then it will be a regression needing fixing.

See the “Additional information” section above.

scripts/plugin-declarations.js
```diff
@@ -0,0 +1,61 @@
...
+export function declarationFilesPlugin() {
+	const sourceDirectoryPath = path.join( process.cwd(), 'src' );
+
+	return {
+		name: 'emit-declaration-files',
+
+		async generateBundle() {
+			const sourceFilePaths = getTypeScriptSourceFiles( sourceDirectoryPath );
+
+			await Promise.all( sourceFilePaths.map( async sourceFilePath => {
```
psmyrek: The plugin processes all source files in parallel. If one file has a declaration error, the build is not stopped: other files may still continue and call `this.emitFile()` before the build fully stops. In the end, build failure can still produce some generated `.d.ts` assets and that creates a "half-done" result.

WDYT about collecting all declaration results first and only emit files if no errors were found?

packages/ckeditor5-dev-changelog/rolldown.config.js
```diff
@@ -0,0 +1,31 @@
...
+const externals = [
+	...Object.keys( pkg.dependencies || {} ),
+	...Object.keys( pkg.peerDependencies || {} )
+];
+
+export default defineConfig( {
+	input: {
+		index: 'src/index.ts',
+		template: 'src/template.ts'
+	},
```
psmyrek: This package builds with multiple entry points and the build output includes files like `constants-xxxx.js` in addition to `index.js` and `template.js`. Any script/tool expecting only fixed filenames may break.

WDYT about disabling code splitting for this package?

scripts/plugin-declarations.js
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/83792da749ef7d97af11158270beb9c5560f6255. The plugin now collects all declaration results first and emits assets only when no declaration errors were found.

packages/ckeditor5-dev-changelog/rolldown.config.js
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/83792da749ef7d97af11158270beb9c5560f6255. The changelog package now builds the two entry points separately with fixed output names, so the build no longer creates a shared hashed chunk.

packages/ckeditor5-dev-changelog/rolldown.config.js
```diff
@@ -0,0 +1,31 @@
...
+		cleanDir: true,
+		format: 'esm',
+		dir: 'dist',
+		assetFileNames: '[name][extname]'
+	},
+	external: id => externals.some( name => id.startsWith( name ) ),
+	plugins: [
+		declarationFilesPlugin()
+	]
+} );
```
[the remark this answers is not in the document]
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/83792da749ef7d97af11158270beb9c5560f6255. The changelog package now builds the two entry points separately with fixed output names, so the build no longer creates a shared hashed chunk.

scripts/plugin-declarations.js
```diff
@@ -0,0 +1,61 @@
...
+
+					this.error( `Could not generate a declaration file for "${ filename }".\n${ errorMessage }` );
+				}
+
+				this.emitFile( {
+					type: 'asset',
+					fileName: getDeclarationFileName( filename ),
+					source: code
+				} );
+			} ) );
```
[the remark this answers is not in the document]
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/83792da749ef7d97af11158270beb9c5560f6255. Declaration generation now aggregates errors across all files before calling this.error(), and the test covers multiple failing files.

psmyrek (approved):

PR #1349, opened 2026-04-09
Improved argument parsing.

przemyslaw-zan:
### 🚀 Summary

`ckeditor5-dev-tests-run-automated` and `ckeditor5-dev-tests-run-manual` will now point the user to the `--help` argument when an unknown argument is parsed, then exit process with code 1.

`ckeditor5-dev`'s internal `test` script was also improved in the same way. Also added `--help` and `--files` argument which allows running only specified packages, instead of all of them.

---

### 💡 Testing

- Setup `ckeditor5-dev` within `ckeditor5` on the branch `ci/4355`. Attempt to run both manual and automatic tests with real and fake arguments.
- Run `pnpm run test` within `ckeditor5-dev` with real and fake arguments.


psmyrek (commented):
It's now possible to run tests with a flag that doesn't make sense, and validation won't catch it as an invalid call. For example, you can run manual tests with the coverage flag. This flag is silently ignored. This behavior made sense previously when invalid arguments were simply ignored, but now I think it's worth addressing.

psmyrek (approved):
It still allows a `--no-*` prefix for non-boolean options, leading to invalid runtime types. Example: `pnpm manual --files alignment --no-port` is accepted but it leads to a runtime error:

```
TypeError [ERR_INVALID_ARG_VALUE]: The argument 'options' is invalid. Received { port: false }
```

Fixing this might be tricky, because we want to support the `--no-*` prefix for non-boolean [`--identity-file`](https://github.com/ckeditor/ckeditor5-dev/blob/ad0cb18ea2ce4f420242052ad3fbc42d3e8df83c/packages/ckeditor5-dev-tests/lib/utils/parsearguments.js#L495-L498).

For this reason, LGTM.

PR #1353, opened 2026-04-14
Added typeOnlyExportValidator.

przemyslaw-zan:
### 🚀 Summary

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5/issues/18942.

---

### 💡 Additional information

Testing: link to `ckeditor5`. Run the `docs` script. Expect warnings such as:

```
./packages/ckeditor5-engine/src/index.ts:40:1 - [warning] Class "DowncastDispatcher" must not be exported as a type-only export.

40      DowncastDispatcher,

./packages/ckeditor5-engine/src/index.ts:55:1 - [warning] Class "UpcastDispatcher" must not be exported as a type-only export.

55      UpcastDispatcher,

./packages/ckeditor5-engine/src/index.ts:95:14 - [warning] Class "ModelConsumable" must not be exported as a type-only export.

95    export type { ModelConsumable } from './conversion/modelconsumable.js';

```

### DIFF
--- a/.changelog/20260414175411_ck_18942.md
+++ b/.changelog/20260414175411_ck_18942.md
@@ -1,0 +1,9 @@
+---
+type: Feature
+scope:
+  - typedoc-plugins
+closes:
+  - ckeditor/ckeditor5#18942
+---
+
+Added a new TypeDoc validator that detects classes incorrectly exported as type-only exports (`export type { Foo }` or `export { type Foo }`). Type-only exports strip classes to their structural type, causing TypeDoc to render them as interfaces instead of classes in API documentation. Ambient `declare class` declarations are excluded from this check.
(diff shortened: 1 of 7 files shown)


pomek:
Is there any reason for creating a typedoc plugin instead of an ESLint rule? In the past, to disallow particular imports, we used ESLint, and it still happens.

Using typedoc as the `export` validator seems to be overkill. Some of the developers even notice that they're doing something wrong when it's a warning, which most likely would be ignored.

There is already a rule [`ckeditor5-rules/no-default-export`](https://github.com/ckeditor/ckeditor5-linters-config/blob/3344a18aac36426f2f6213ee3106c50b776cd2ae/packages/eslint-plugin-ckeditor5-rules/lib/rules/no-default-export.js) that checks exports. Why wouldn't you create a new one based on this?

martnpaneq (changes requested):
This PR adds a third way of validating our code - using typedoc during the `pnpm run docs` command. This is not ideal, because no one would think about running docs to validate types.

We already use `eslint` and we have a special script for validating exports specifically - `validate-module-re-exports`.

I think that adding a typedoc plugin is not the right approach here, and we should update the `validate-module-re-exports` script.

przemyslaw-zan:
Moved functionality to https://github.com/ckeditor/ckeditor5/pull/20069

PR #1356, opened 2026-04-30
Bumped deps.

przemyslaw-zan:
### 🚀 Summary

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -12,8 +12,10 @@ overrides:
   'diff@^7': '^8.0.3'
   'lodash@^4': '^4.18.1'
   'minimatch@^9': '^9.0.7'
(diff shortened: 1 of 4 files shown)


pnpm-workspace.yaml
```diff
@@ -12,8 +12,10 @@ overrides:
   'diff@^7': '^8.0.3'
   'lodash@^4': '^4.18.1'
   'minimatch@^9': '^9.0.7'
+  'postcss@^8': '^8.5.12'
```
martnpaneq, line 15: Can we remove those entries after updating the lock file?

przemyslaw-zan: I've checked, and it is needed to prevent older versions from installing.

pomek (approved):

PR #1358, opened 2026-05-04
Removed module validator.

przemyslaw-zan:
### 🚀 Summary

Removed the `module-validator` from `typedoc-plugins`.


.changelog/20260504160308_ci_4389.md
```diff
@@ -0,0 +1,7 @@
+---
+type: Other
+scope:
+  - typedoc-plugins
+---
+
+Removed the `module-validator` from `typedoc-plugins`.
```
psmyrek: Would you consider making the change a bit more descriptive for anyone who potentially used it? I would briefly mention the reason for removing it: the validator became unnecessary because we have a dedicated ESLint rule `validate-module-tag` that does the same thing, etc.

psmyrek (approved):

PR #1371, opened 2026-05-26
Remove the `--disable-gpu` flag from the web crawler configuration

filipsobol:
### 🚀 Summary

Remove the `--disable-gpu` flag from the web crawler configuration, which causes performance issues on macOS.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-internal/issues/4486.

---

### 💡 Additional information

How to test in the CKE5 repository:

1. Run `pnpm run docs`.
2. Run `pnpm run docs:serve`.
3. In another terminal, run `pnpm run docs:verify`.

### DIFF
--- a/.changelog/20260526101915_ci_4486_improve_web_crawler_performance.md
+++ b/.changelog/20260526101915_ci_4486_improve_web_crawler_performance.md
@@ -1,0 +1,6 @@
+---
+type: Other
+scope: ckeditor5-dev-web-crawler
+---
+
+Remove the `--disable-gpu` flag from the web crawler configuration which causees performance issues on macOS.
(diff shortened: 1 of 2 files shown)


martnpaneq (approved):
Looks like the fix is working on Mac. Docs crawler works correctly on branch:

Image

But it takes ~6h and runs out of memory on `master`:

Image

przemyslaw-zan:
I've ran this on windows, [Intel® Core™ Ultra 7 Processor 165U with integrated graphics](https://www.intel.com/content/www/us/en/products/sku/237329/intel-core-ultra-7-processor-165u-12m-cache-up-to-4-90-ghz/specifications.html).

Under WSL, there is no difference, all runs with and without the fix average around 6.4 to 6.8 minutes.

Under native windows, the averages after 3 runs each are 2.93 minutes before, 2.83 minutes after, so negligible to no difference.

PR #1397, opened 2026-06-17
Bumped deps.

przemyslaw-zan:
### 🚀 Summary

---

### 📌 Related issues

* Closes #000

---

### 💡 Additional information

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -14,7 +14,7 @@ overrides:
   'serialize-javascript@6': '^7.0.3'
   'uuid@^8': '^14.0.0'
   'ws@^8': '^8.20.1'
-  'esbuild@^0.27': '^0.28.1'
+  'js-yaml@^3': '^4.2.0'
 
 minimumReleaseAge: 4320 # 3 days
 minimumReleaseAgeExclude:
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -15,7 +15,6 @@ overrides:
   'uuid@^8': '^14.0.0'
   'ws@^8': '^8.20.1'
   'esbuild@^0.27': '^0.28.1'
-  'js-yaml@^3': '^4.2.0'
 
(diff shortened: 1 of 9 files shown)


pnpm-workspace.yaml
```diff
@@ -11,21 +11,17 @@ onlyBuiltDependencies:
 overrides:
   'diff@^7': '^8.0.3'
   'lodash@^4': '^4.18.0'
-  'qs@^6': '^6.15.2'
   'serialize-javascript@6': '^7.0.3'
   'uuid@^8': '^14.0.0'
   'ws@^8': '^8.20.1'
+  'esbuild@^0.27': '^0.28.1'
```
psmyrek: Instead of adding another override, could you please check if bumping `esbuild-loader` to v4.5.0 in `ckeditor5-dev-utils` fixes the thing in `esbuild`?

przemyslaw-zan:
I've added a `'js-yaml@^3': '^4.2.0'` override. Two packages in our tree pulled `js-yaml` v3:

- **`@istanbuljs/load-nyc-config`** (via `babel-plugin-istanbul`) - calls only `.load()`,
  which exists in v4 (there it is the default, safe loader). Fully compatible, no change needed.
- **`depcheck`** (used by our dependency checker) - calls the legacy `.safeLoad()`, which v4
  removed (it was renamed to `.load()`).

## Why it's safe

`depcheck`'s `.safeLoad()` calls are only reached when it parses a `.travis.yml`, a
`serverless.yml`, or a YAML config file. Our codebase contains none of these, and one of those
call sites is additionally wrapped in a `try/catch`. I've confirmed it empirically: running the
full dependency check across the codebase on `js-yaml` v4 completes with no errors.

**Caveat:** if a `.travis.yml`, `serverless.yml`, or YAML `.depcheckrc` is ever added to a
package the checker scans, `depcheck` would throw on those paths (v4 has no `.safeLoad()`).
None exist today, so this is latent only.

psmyrek (approved):

PR #1407, opened 2026-06-26
Add the `ckDebugPlugin()` Vite plugin to the manual server package

filipsobol:
### 🚀 Summary

* Add the `ckDebugPlugin()` Vite plugin to the manual server package.
* Auto attach inspector in Vite-based manual test server.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-internal/issues/4536.

---

### 💡 Additional information

N/A

### DIFF
--- a/packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
+++ b/packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
@@ -1,0 +1,56 @@
+/**
+ * @license Copyright (c) 2003-2026, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import type { Plugin } from 'vite';
+
+const SOURCE_FILE_REGEXP = /\.[cm]?[jt]sx?$/;
+
+export function ckDebugPlugin(): Plugin {
+	const debugFlags = getDebugFlags( process.env.CK_DEBUG );
+
+	return {
+		name: 'ck-debug',
+		enforce: 'pre',
+		transform: {
+			filter: {
+				id: {
+					include: SOURCE_FILE_REGEXP
+				}
+			},
+			handler( code ) {
+				if ( !code.includes( '// @if ' ) ) {
+					return null;
+				}
+
+				const transformedCode = code.replace(
+					/(^[ \t]*)\/\/ @if (!?[\w]+) \/\/(.*)$/gm,
+					( match, indentation: string, flagName: string, body: string ) => {
+						if ( !debugFlags.has( flagName ) ) {
+							return match;
+						}
+
+						return `${ indentation }/* @if ${ flagName } */${ body }`;
+					}
+				);
+
+				return transformedCode === code ? null : transformedCode;
+			}
+		}
+	};
+}
+
+function getDebugFlags( debugOption?: string ): Set<string> {
+	if ( !debugOption || debugOption === 'false' ) {
+		return new Set();
+	}
+
+	return new Set( [
+		'CK_DEBUG',
+		...debugOption.split( ',' )
+			.map( flag => flag.trim() )
+			.filter( Boolean )
+			.map( flag => `CK_DEBUG_${ flag.toUpperCase() }` )
+	] );
+}
--- a/packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
+++ b/packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
@@ -49,16 +49,18 @@ export function ckDebugPlugin(): Plugin {
 	};
 }
 
-function getDebugFlags( debugOption?: string ): Set<string> {
-	if ( !debugOption || debugOption === 'false' ) {
+function getDebugFlags( debugOption: string = '' ): Set<string> {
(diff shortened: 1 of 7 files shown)


packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @license Copyright (c) 2003-2026, CKSource Holding sp. z o.o. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+import type { Plugin } from 'vite';
+
+const SOURCE_FILE_REGEXP = /\.[cm]?[jt]sx?$/;
+
+export function ckDebugPlugin(): Plugin {
+	const debugFlags = getDebugFlags( process.env.CK_DEBUG );
```
filipsobol, line 11: Maybe instead of relying on environment variables like `CK_DEBUG=mention,focustracker,typing pnpm manual:vite`, we allow passing an array to `ckDebugPlugin` itself like so:

```ts
ckDebugPlugin( [
  'mention',
  'focustracker',
  'typing'
] )
```

This way, developers don't have to remember the exact syntax, but only to pass args to the plugin?

**This would be more in line with how we allow changing loaded preset definition** by updating the following import path in `vite.manual.ts`:

```ts
import identityDefinitions from './scripts/presets/staging.json' with { type: 'json' };
```

packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
```diff
@@ -0,0 +1,64 @@
...
+				return transformedCode === code ? null : transformedCode;
+			}
+		}
+	};
+}
+
+function getDebugFlags( debugOption?: string ): Set<string> {
+	if ( !debugOption || debugOption === 'false' ) {
+		return new Set();
+	}
```
psmyrek: This changes the default behavior. Now, if `CK_DEBUG` is not set, debug is off. Previously, `CK_DEBUG` was enabled by default.

packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
psmyrek: Editing `vite.manual.mts` is already how we choose an identity file and the manual test include list, so putting debug flags in the config fits how the Vite manual server works today.

On the other hand, debug flags are usually quick, temporary switches. Making people edit the config for every debug run is awkward and can leave behind messy local changes. Passing them straight from the command line works much better from DX point of view.

In my personal opinion, the best solution would be a wrapper around Vite that supports the custom logic we had for a long time and that developers are already used to. Not all of the old options need to be supported, just some of them: debug flags, identity file, include list.

psmyrek (commented):
1. Current source has 32 debug-only `require()` lines (examples: [here](https://github.com/ckeditor/ckeditor5-commercial/blob/f3213838ba3256fe5f647349b60eb8a4193e4772/external/ckeditor5/packages/ckeditor5-utils/src/dom/position.ts#L15-L20), [here](https://github.com/ckeditor/ckeditor5-commercial/blob/f3213838ba3256fe5f647349b60eb8a4193e4772/external/ckeditor5/packages/ckeditor5-utils/src/dom/position.ts#L15-L20) and [here](https://github.com/ckeditor/ckeditor5-commercial/blob/bd30aef28f8455686790a8f91466893d758c4d7f/external/ckeditor5/packages/ckeditor5-engine/src/model/model.ts#L45-L46)). When `ckDebugPlugin()` uncomments them, Vite leaves `require()` calls in browser ESM output and that will fail. 
    ```
    CK_DEBUG=mention,engine pnpm manual:vite
    ```
    After opening http://localhost:8125/external/ckeditor5/packages/ckeditor5-mention/tests/manual/mention.html, error is thrown:

    ```
    Uncaught TypeError: (intermediate value)(...).default is undefined
        <anonymous> model.ts:46
    ```

2. The `ckDebugPlugin()` has been defined in `ckeditor5-dev-manual-server`. What about supporting debug flags in automated tests?

3. Related PR is missing in `ckeditor5-commercial` that will enable the new plugin in `vite.manual.mts`.

packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
filipsobol: The plugin will keep using CLI env vars for now. We will discuss wrapping Vite to allow custom CLI flags in the follow-ups.

filipsobol:
It looks like the original issue that forced us to use `require` is no longer valid, so I replaced all such calls with static `import`s:

* https://github.com/ckeditor/ckeditor5-commercial/pull/10748
* https://github.com/ckeditor/ckeditor5-linters-config/pull/127

Some imports also tried to pull default exports that were no longer valid (replaced by named exports), so I updated them too.

filipsobol:
> The `ckDebugPlugin()` has been defined in `ckeditor5-dev-manual-server`. What about supporting debug flags in automated tests?

Added in https://github.com/ckeditor/ckeditor5-commercial/pull/10748.

psmyrek (approved):

PR #1408, opened 2026-06-29
Handle CircleCI jobs without the "dependencies" property in the workflow notifier.

filipsobol:
### 🚀 Summary

Handle CircleCI jobs without the `dependencies` property in the workflow notifier. The CircleCI API may return job objects without that property, which caused `processJobStatuses()` to throw. Such jobs are now treated as jobs with no dependencies.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-internal/issues/4559

---

### 💡 Additional information

The fix is a minimal optional-chaining guard (`!job.dependencies?.length`); the later dependency lookup is already gated behind that check. Added test coverage for jobs with a missing `dependencies` property.

### DIFF
--- a/.changelog/20260629121027_ci_4559_handle_undefined_job_dependencies.md
+++ b/.changelog/20260629121027_ci_4559_handle_undefined_job_dependencies.md
@@ -1,0 +1,7 @@
+---
+type: Fix
+scope:
+  - ckeditor5-dev-ci
+---
+
+The workflow notifier no longer throws when the CircleCI API returns a job without the `dependencies` property. Such a job is now treated as a job with no dependencies.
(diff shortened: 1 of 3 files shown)


psmyrek (approved):

pomek:
From the issue DoD's:

> Consider adding retry logic to handle CircleCI API instability

Was this considered? And if so, why did you reject implementing it?

filipsobol:
> From the issue DoD's:
> 
> > Consider adding retry logic to handle CircleCI API instability
> 
> Was this considered? And if so, why did you reject implementing it?

Yes, but I generally avoid adding logic for blindly retrying the code because it can mask real issues. It would be overkill here because the fix for the actual root cause was small enough, and such errors are so uncommon.

IMO, retry logic should only be added as a last resort, when the provided code is fundamentally flawed and cannot be fixed or the expense is too great to justify.

PR #1410, opened 2026-07-01
Disable debug plugin unless `CK_DEBUG` env var is set

filipsobol:
### 🚀 Summary

* Disable the debug plugin transform when the `CK_DEBUG` environment variable is not set.
* Keep `CK_DEBUG=true` and namespace values such as `CK_DEBUG=engine` enabled.

---

### 📌 Related issues

* https://github.com/ckeditor/ckeditor5-internal/issues/4536

---

### 💡 Additional information

The previous version skipped registering the transform hook only when `CK_DEBUG=false` was set, but we want it to skip the transform when:

- `CK_DEBUG` is not provided at all,
- an empty value is provided, for example `CK_DEBUG= pnpm run ...` (then `CK_DEBUG` equals `''`),
- `CK_DEBUG=false pnpm run ...` is provided.

### DIFF
--- a/.changelog/20260701145301_ci_4536_disable_ck_debug_by_default.md
+++ b/.changelog/20260701145301_ci_4536_disable_ck_debug_by_default.md
@@ -1,0 +1,7 @@
+---
+type: Fix
+scope:
+  - ckeditor5-dev-manual-server
+---
+
+The debug plugin no longer enables debug code unless the `CK_DEBUG` environment variable is set.
--- a/packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
+++ b/packages/ckeditor5-dev-manual-server/src/debug-plugin/plugin.ts
@@ -49,8 +49,8 @@ export function ckDebugPlugin(): Plugin {
 	};
 }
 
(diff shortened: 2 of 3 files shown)


psmyrek (commented):
Handling `CK_DEBUG` flag affects what gets tested, so why do we want to change the default behavior? Now, if we don't pass this flag, the code behind `CK_DEBUG` will no longer run in the tests. Previously this flag was always enabled by default.

Maybe it's enough to add `v8 ignore` (also behind `CK_DEBUG`) in a few places if the problem is that V8 measures coverage much more accurately than the old Istanbul.

filipsobol:
Adding support for the `CK_DEBUG=...` flag causes some tests to drop coverage and others to timeout.

After checking this again, I found that:
* the old webpack-based manual test server handled the `--debug` flag the same as `ckDebugPlugin()` that is on the `#master` branch (and released),
* the old Karma-based automated test server parsed the `--debug` flag but **ignored it**.

The root of the issue is that I was trying to add support for `CK_DEBUG=...` to the new Vitest-based server, which the old implementation didn't have.

There's no requirement to have support for this, so I'm closing the issue.

PR #1411, opened 2026-07-02
Introduce new manual tests format

filipsobol:
### 🚀 Summary

Introduce new manual tests format.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-internal/issues/4551.

---

### 💡 Additional information

N/A


packages/ckeditor5-dev-manual-server/src/static-assets-plugin/plugin.ts
```diff
@@ -0,0 +1,61 @@
...
+		includePackageNames
+	) );
+	const getManualStaticAssets = manualStaticAssetsCache.get;
+
+	return {
+		name: 'ckeditor5-manual-static-assets',
+
+		configResolved( config ) {
+			workspaceRoot = config.root;
+		},
```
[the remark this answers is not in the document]
filipsobol: Not a real issue. No change made.

The cache can only go stale if something reads it before `configResolved` sets the correct `workspaceRoot`. But this plugin has no `config()` hook — its cache is only read in `configureServer` (request time) and `generateBundle` (build time), both of which run *after* `configResolved`. So it's always computed with the right root. The sibling `manualTestsPlugin` needs `invalidate()` only because its `config()` hook reads the pages cache *before* `configResolved`; that early read doesn't exist here.

packages/ckeditor5-dev-manual-server/src/manual-test-plugin/plugin.ts
```diff
@@ -169,131 +130,179 @@ export function manualTestsPlugin( options: ManualTestsPluginOptions ): Plugin {
...
+function composeFreshManualHtml( builtHtml: string, sourceHtml: string ): string | null {
+	const builtHeadEnd = builtHtml.toLowerCase().indexOf( HEAD_CLOSE_TAG );
+	const sourceHeadEnd = sourceHtml.toLowerCase().indexOf( HEAD_CLOSE_TAG );
+
+	if ( builtHeadEnd == -1 || sourceHeadEnd == -1 ) {
+		return null;
 	}
 
-	const normalizedIncludePackageNames = new Set( includePackageNames.map( normalizePackageName ) );
+	return builtHtml.slice( 0, builtHeadEnd + HEAD_CLOSE_TAG.length ) + sourceHtml.slice( sourceHeadEnd + HEAD_CLOSE_TAG.length );
```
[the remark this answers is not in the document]
filipsobol: Real behavior, but not a fixable bug. Documented it instead (commit [3d53465](https://github.com/ckeditor/ckeditor5-dev/pull/1411/changes/3d534650dbabd02f31afacae6d0296d4778f8a89)).

Adding `<ck-manual-header>` to a page after the initial build genuinely leaves the head without the header `<meta>` + component script under `bundledDev`, so chrome stays broken until restart. A code fix isn't feasible — regenerating the head injection means re-running the exact build pipeline the dev engine refuses to run for HTML entries, which is why the splice workaround exists at all. The real defect was that the existing doc comment *understated* the limitation (it claimed body edits are always reflected and only head-confined edits need a restart). I corrected the comment to state that toggling `<ck-manual-header>` also requires a restart.

packages/ckeditor5-dev-manual-server/src/manual-test-plugin/plugin.ts
```diff
@@ -169,131 +130,182 @@ export function manualTestsPlugin( options: ManualTestsPluginOptions ): Plugin {
...
+		try {
+			const sourceFilePath = resolve( workspaceRoot, filePath );
+			const sourceMtime = statSync( sourceFilePath ).mtimeMs;
+			const builtSourceMtime = builtSourceMtimes.get( filePath );
 
-function getFileSource( file: { source: string | Uint8Array } ): string {
-	return typeof file.source == 'string' ? file.source : Buffer.from( file.source ).toString( 'utf8' );
-}
+			if ( builtSourceMtime == undefined ) {
+				builtSourceMtimes.set( filePath, sourceMtime );
```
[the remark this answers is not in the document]
filipsobol: Not a real issue. No change made.

The premise is that memory-file keys "use a leading slash (as before this refactor)." They don't — the keys are relative paths *without* a leading slash (`packages/ckeditor5-foo/tests/manual/foo.manual.html`), which the tests encode and the implementation was verified against for Vite 8.1.0. The two consumers need *different* formats and both are correct:

- `getManualPages().has( toPublicSpecifier( filePath ) )` — the page map is keyed by public specifier (`/packages/…`), so the lookup normalizes.
- `resolve( workspaceRoot, filePath )` — needs the **relative** form. Passing the `toPublicSpecifier` form here would make it absolute (`/packages/…`) and break `resolve`.

So using the raw key for `resolve` isn't an oversight — it's required. The old shell-based implementation resolved the disk path the exact same way (`resolve( workspaceRoot, filePath )`), so the "before this refactor keys had a slash" premise is false. And Vite passes one consistent key format, so the mtime map keyed by that same value is consistent.

packages/ckeditor5-dev-manual-server/src/manual-test-plugin/plugin.ts
```diff
@@ -169,131 +130,182 @@ export function manualTestsPlugin( options: ManualTestsPluginOptions ): Plugin {
...
+				return file;
+			}
+
+			if ( sourceMtime <= builtSourceMtime ) {
+				return file;
+			}
 
-	return `${ pattern }.{html,js,md,ts}`;
+			// HTML entry outputs are always emitted as strings.
+			const freshHtml = composeFreshManualHtml( file.source as string, readFileSync( sourceFilePath, 'utf8' ) );
```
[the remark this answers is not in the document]
filipsobol: Not a real issue. No change made.

HTML entry outputs are emitted as strings — asserted deliberately with the inline comment and `as string`. There was an earlier `getFileSource` helper decoding `Uint8Array`, but that helper served the *shell-composition* path that's since been removed; it was dropped on purpose because HTML entries are strings under bundledDev (verified against Vite 8.1.0). Non-page memory files (JS/binary assets, which *can* be `Uint8Array`) are already filtered out one check earlier (`getManualPages().has(...)`), so only string HTML reaches `composeFreshManualHtml`. And in the impossible-today bytes case, `toLowerCase` throwing is caught and the built output is served — graceful degradation, not breakage.

packages/ckeditor5-dev-manual-server/src/manual-test-plugin/plugin.ts
```diff
@@ -423,7 +339,3 @@ function parseRequestUrl( requestUrl: string | undefined ): URL | null {
 	// @ts-expect-error Remove when we upgrade TypeScript and bump `target`.
```
przemyslaw-zan, line 365: Unused '@ts-expect-error' directive.

packages/ckeditor5-dev-manual-server/src/manual-test-plugin/plugin.ts
```diff
@@ -221,79 +213,87 @@ function useBundledDevManualHtmlSource(
...
-			html: readFileSync( resolve( workspaceRoot, stripLeadingSlash( entry.htmlFilePath ) ), 'utf8' ),
-			shellScriptPublicPath,
-			shellTemplateFilePath: MANUAL_SHELL_TEMPLATE_FILE_PATH,
-			workspaceRoot
-		} );
-		const html = mergeBundledAssetTags( transformedSourceHtml, bundledHtml, entry, shellScriptPublicPath );
-
-		return { source: html };
+		try {
+			const sourceFilePath = resolve( workspaceRoot, filePath );
```
przemyslaw-zan: We could apply `stripLeadingSlash( filePath )` here just to be sure.

packages/ckeditor5-dev-manual-server/src/manual-test-plugin/plugin.ts
```diff
@@ -221,79 +213,87 @@ function useBundledDevManualHtmlSource(
...
-	return typeof file.source == 'string' ? file.source : Buffer.from( file.source ).toString( 'utf8' );
-}
+/**
+ * Combines the freshly built `<head>` (asset tags and injected manual chrome) with the current
+ * post-`</head>` markup from the source file. Returns `null` when either document is missing a
+ * `</head>`, so the caller can fall back to the built output.
+ */
+function composeFreshManualHtml( builtHtml: string, sourceHtml: string ): string | null {
+	const builtHeadEnd = builtHtml.toLowerCase().indexOf( HEAD_CLOSE_TAG );
+	const sourceHeadEnd = sourceHtml.toLowerCase().indexOf( HEAD_CLOSE_TAG );
```
przemyslaw-zan: `toLowerCase()` can change length for some Unicode chars, shifting the splice point. /<\/head>/i.exec() is the same line count.

packages/ckeditor5-dev-manual-server/src/static-assets-plugin/static-assets.ts
```diff
@@ -8,7 +8,11 @@ import { globSync, statSync, createReadStream } from 'node:fs';
 import type { IncomingMessage, ServerResponse } from 'node:http';
 import { toPublicSpecifier } from '../utils.js';
 
-const PROCESSED_MANUAL_TEST_EXTENSIONS = new Set( [ '.html', '.js', '.md', '.ts' ] );
+// Extensions handled by Vite's module/HTML pipeline, never served raw as static fixtures.
+// `.manual.html` entries go through the HTML pipeline too, but plain `.html` fixtures must be
+// served raw, so they are excluded by suffix in `isManualStaticAssetPath` rather than by extension.
+const PROCESSED_MANUAL_TEST_EXTENSIONS = new Set( [ '.js', '.md', '.ts' ] );
```
przemyslaw-zan: `.md` is stale.

przemyslaw-zan (changes requested):
Works well, I only have a few nitpicks.

---

The workarounds could use a reference to https://github.com/rolldown/rolldown/issues/10022#issuecomment-4893217167.

packages/ckeditor5-dev-manual-server/src/manual-test-plugin/plugin.ts
filipsobol: It's used, but IDEs report it as unused. It will sort itself out after we the editor to a newer TS version.

packages/ckeditor5-dev-manual-server/src/manual-test-plugin/plugin.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/pull/1411/changes/6958f53aa6f98649a3413ca2c0e0e57d183b24a6.

packages/ckeditor5-dev-manual-server/src/manual-test-plugin/plugin.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/pull/1411/changes/6958f53aa6f98649a3413ca2c0e0e57d183b24a6.

packages/ckeditor5-dev-manual-server/src/static-assets-plugin/static-assets.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/pull/1411/changes/6958f53aa6f98649a3413ca2c0e0e57d183b24a6.

przemyslaw-zan (approved):

PR #1412, opened 2026-07-06
Rename the equalMarkup Chai assertion to the toEqualMarkup Vitest matcher

pomek:
### 🚀 Summary

Renames the `equalMarkup` Chai assertion to the `toEqualMarkup()` Vitest matcher (matching the Vitest naming style) and removes the unused `attribute` Chai assertion.

The matcher is registered in a Vitest setup file via `expect.extend( { toEqualMarkup } )`, importing it from the main entry point of `@ckeditor/ckeditor5-dev-tests`. To make that import safe in a browser context, the main entry point now exposes only the matcher — the test runner tasks and `parseArguments()` are no longer exported (the binary scripts import them directly and remain the public interface of the runners). The package ships the `vitest` type augmentation (`lib/vitest/matchers.d.ts`), so consumers get `toEqualMarkup()` typings out of the box.

---

### 📌 Related issues

* Closes [ckeditor/ckeditor5-internal#4595](https://github.com/ckeditor/ckeditor5-internal/issues/4595)

---

### 💡 Additional information

* The `attribute` assertion had zero call sites in `ckeditor5` and `ckeditor5-commercial`, hence it is removed instead of migrated (a deviation from the issue's DoD).
* The old `equalMarkup` Chai factory stays in this PR — the Karma entry-file generator still registers it. It is removed together with the Karma runner in #1413.


pomek:
Superseded by #1413 — both changes were combined there (the matcher migration and the runner removal are tightly coupled), and the PR now targets `master` directly. Closing.

PR #1423, opened 2026-07-13
Support triggering CircleCI GitHub App pipeline definitions in the trigger-circle-build script

pomek:
### 🚀 Summary

Added support for triggering a specific CircleCI GitHub App pipeline definition using the new, optional `--pipeline-definition-id` option of the `ckeditor5-dev-ci-trigger-circle-build` command. When the option is provided, the script calls the `POST /project/gh/{slug}/pipeline/run` endpoint with the given `definition_id` instead of the project's default OAuth pipeline. The release flow in `.circleci/template.yml` now uses the option when triggering the release pipeline.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-internal/issues/4615.

---

### 💡 Additional information

* The pipeline definition ID is not a secret. It is a plain identifier that is useless without a valid CircleCI token, so it can live in the repository configuration.
* The new `/pipeline/run` endpoint accepts only the `gh` provider prefix in the project slug, while the legacy endpoint supports both `gh` and `github`. The URL for the new code path reflects that.
* Verified end-to-end: running the command from this branch with `--pipeline-definition-id` created pipelines `#3478`/`#3479` on `master` in `ckeditor/ckeditor5-dev`, and the `main` workflow executed its jobs as expected. The release step was not started because the release-readiness guard halted it.

### DIFF
--- a/.changelog/20260713112818_support_circleci_pipeline_definitions.md
+++ b/.changelog/20260713112818_support_circleci_pipeline_definitions.md
@@ -1,0 +1,7 @@
+---
+type: Feature
+scope:
+  - ckeditor5-dev-ci
+---
+
+Added support for triggering a specific CircleCI GitHub App pipeline definition using the `--pipeline-definition-id` option of the `ckeditor5-dev-ci-trigger-circle-build` command.
--- a/.circleci/template.yml
+++ b/.circleci/template.yml
@@ -170,7 +170,7 @@ jobs:
           command: pnpm ckeditor5-dev-ci-circle-disable-auto-cancel-builds --organization ckeditor --repository ckeditor5-dev
       - run:
           name: Trigger the release pipeline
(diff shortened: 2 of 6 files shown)


pomek:
- Example job: https://app.circleci.com/pipelines/github/ckeditor/ckeditor5-dev/3479/workflows/de00a0b2-ebfd-4b7d-bf0a-f9d897b2c061
- Triggered via:
	```bash
	CKE5_CIRCLE_TOKEN=<hidden> \
	CIRCLE_SHA1=b42ccd4553605df2f457113ece9446d4842ad1de \
	CIRCLE_BRANCH=master \
	node packages/ckeditor5-dev-ci/bin/trigger-circle-build.js \
	  --slug ckeditor/ckeditor5-dev \
	  --pipeline-definition-id 73d66b9b-677a-4100-bf5a-138833dbc4d5
	```

psmyrek (approved):

PR #1433, opened 2026-07-15
Limit manual test reloads to affected clients

filipsobol:
### 🚀 Summary

Scopes Vite bundled-dev HTML and CSS full reloads to the affected manual-test client instead of delegating them to the environment-wide reload path. JavaScript changes continue to use the existing manual refresh prompt.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/11045

---

### 💡 Additional information

The plugin waits for the latest bundled output before reloading and falls back to the best available output if refreshing fails. Regression coverage verifies HTML client isolation, CSS full reloads, and build-output failures.

### DIFF
--- a/packages/ckeditor5-dev-manual-server/src/refresh-plugin/plugin.ts
+++ b/packages/ckeditor5-dev-manual-server/src/refresh-plugin/plugin.ts
@@ -168,6 +168,10 @@ function ensureLatestBuildOutput( bundledDev: BundledDevInternals ): void {
 }
 
 function sendManualRefreshPayload( payload: HotPayload, send: ( payload: HotPayload ) => void ): void {
+	if ( payload.type == 'update' && !payload.updates.length ) {
+		return;
+	}
+
 	if ( shouldShowManualRefreshPrompt( payload ) ) {
 		send( {
 			type: 'custom',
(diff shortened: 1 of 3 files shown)


packages/ckeditor5-dev-manual-server/src/refresh-plugin/plugin.ts
```diff
@@ -104,21 +106,25 @@ function wrapBundledDevClientSend( clients: BundledDevInternals[ 'clients' ] ):
...
+		if ( hmrOutput.type == 'FullReload' ) {
+			if ( shouldShowManualRefreshPromptForFiles( files ) ) {
+				ensureLatestBuildOutput( bundledDev );
+
+				client.send( {
+					type: 'custom',
+					event: MANUAL_REFRESH_EVENT_NAME
+				} );
+			} else {
+				reloadClientAfterLatestBuildOutput( bundledDev, client, files, workspaceRoot );
```
martnpaneq: Looks like async function call is missing an `await`.

packages/ckeditor5-dev-manual-server/src/refresh-plugin/plugin.ts
```diff
@@ -104,21 +106,25 @@ function wrapBundledDevClientSend( clients: BundledDevInternals[ 'clients' ] ):
...
 	bundledDev.handleHmrOutput = ( client, files, hmrOutput, invalidateInformation ) => {
-		if ( hmrOutput.type == 'FullReload' && shouldShowManualRefreshPromptForFiles( files ) ) {
-			ensureLatestBuildOutput( bundledDev );
-
-			client.send( {
-				type: 'custom',
-				event: MANUAL_REFRESH_EVENT_NAME
-			} );
+		if ( hmrOutput.type == 'FullReload' ) {
+			if ( shouldShowManualRefreshPromptForFiles( files ) ) {
```
martnpaneq: *nitpick* 

We could use a guard clause instead of nested `if`s.

martnpaneq (approved):
Change works correctly - After editing HTML in one manual test no other manual tests are reloaded.

PR #1438, opened 2026-07-17
Emit separate editor and content stylesheets from the build tools

filipsobol:
### 🚀 Summary

Replaces the `splitCss` plugin with a reworked `bundleCss` plugin in `ckeditor5-dev-build-tools`. CSS must now be imported through explicit `index-editor.css` and `index-content.css` entry points, and the build emits three stylesheets: the combined bundle (preserving the JavaScript import order, where the editor and content styles of each feature sit next to each other), plus separate `-editor` and `-content` bundles.

The CSS ordering traversal now starts from the chunk facade module instead of every chunk module, which reduces the plugin's `generateBundle()` time on the full `ckeditor5` bundle from ~1.2 s to ~30 ms with byte-identical output.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-internal/issues/4268

---

### 💡 Additional information

The three stylesheets are bundled by Lightning CSS from virtual entry points, so `@import`s are flattened and warnings are forwarded to Rolldown with correct source locations.


packages/ckeditor5-dev-build-tools/tests/plugins/bundleCss/bundleCss.test.ts
```diff
@@ -49,123 +43,169 @@ function getAsset( output: RolldownOutput[ 'output' ], fileName: string ): Outpu
...
 
-	await bundle.generate( {
-		format: 'esm',
-		assetFileNames: '[name][extname]',
-		file: 'input.js'
-	} );
+	await bundle.generate( { format: 'esm', assetFileNames: '[name][extname]', file: 'input.js' } );
 
 	expect( warnings.some( warning => warning.includes( 'Lightning CSS warning in' ) ) ).toBe( true );
 	expect( warnings.some( warning => warning.includes( 'Unknown at rule: @unknown' ) ) ).toBe( true );
```
przemyslaw-zan: Current assertion essentially asserts any amount of occurrences of this text, we should assert that it occurs exactly once.

packages/ckeditor5-dev-build-tools/tests/plugins/bundleCss/bundleCss.test.ts
```diff
@@ -49,123 +43,169 @@ function getAsset( output: RolldownOutput[ 'output' ], fileName: string ): Outpu
...
-test( 'Allows minifying the generated bundle', async () => {
-	const unminifiedOutput = await generateBundle( {
-		fileName: 'styles.css'
-	} );
-
-	const minifiedOutput = await generateBundle( {
-		fileName: 'styles.css',
-		minify: true
-	} );
+test( 'throws when CSS is imported directly instead of through a theme entry point', async () => {
```
przemyslaw-zan, line 118: Add a test that lists multiple instances instead of just a single one.

packages/ckeditor5-dev-build-tools/tests/plugins/bundleCss/bundleCss.mocked.test.ts
przemyslaw-zan, line 310: This and similar tests should check whether `bundleAsyncMock` was called, and check number of calls.

packages/ckeditor5-dev-build-tools/src/plugins/bundleCss.ts
```diff
@@ -50,56 +53,76 @@ function normalizeId( id: string ): string {
...
-	return CSS_ID_REGEXP.test( id );
+	return CSS_ID_REGEXP.test( id ) && !RAW_QUERY_REGEXP.test( id );
 }
 
 /**
  * Returns whether the import specifier references an external resource.
  */
 function isExternalImport( specifier: string ): boolean {
-	return URL_PROTOCOL_REGEXP.test( specifier ) || PROTOCOL_RELATIVE_URL_REGEXP.test( specifier );
+	return URL.canParse( specifier ) || specifier.startsWith( '//' );
```
przemyslaw-zan, line 67: Missing test for `specifier.startsWith( '//' )`.

packages/ckeditor5-dev-build-tools/tests/plugins/bundleCss/bundleCss.test.ts
```diff
@@ -49,123 +43,169 @@ function getAsset( output: RolldownOutput[ 'output' ], fileName: string ): Outpu
...
+		'.aggregate-content-leaf',
+		'.dependency-editor-leaf',
+		'.dependency-content-leaf'
+	].map( selector => combined.indexOf( selector ) );
+
+	expect( combinedOrder ).toEqual( [ ...combinedOrder ].sort( ( a, b ) => a - b ) );
+	expect( combinedOrder.every( index => index >= 0 ) ).toBe( true );
+
+	// CSS entry points imported multiple times must be bundled only once.
+	expect( editor.match( /\.dependency-editor-leaf/g ) ).toHaveLength( 1 );
```
przemyslaw-zan, line 90: There is no test to verify ordering enforced by `[ ...new Set( ids.reverse() ) ].reverse()` -  e.g. an input that imports `dependency-package` first and then `aggregate-package` (which re-imports dependency), plus an assertion like `combined.indexOf( '.aggregate-editor-leaf' ) < combined.indexOf( '.dependency-editor-leaf' )`. That passes only under keep-last.

packages/ckeditor5-dev-build-tools/tests/build/build.test.ts
przemyslaw-zan, line 400: The `Minify` test builds `src/banner.js`, which contains no CSS at all, and asserts only that the JS came out minified, so the `minify` flag's journey from `build()` options through `getRolldownConfig()` into `bundleCss( { minify } )` and Lightning CSS is never verified - if someone dropped `minify` from the plugin options in `config.ts`, no test would fail. We should give the Minify test (or a sibling) a fixture that imports `index-editor.css/index-content.css` so the three CSS assets are non-empty, then assert those assets are actually minified - e.g. an exact-string match, or at least "no newlines/indentation" - for all three (`styles.css`, `-editor`, `-content`). (There is a plugin-level minify test, but it only asserts the minified output is shorter, which almost anything would satisfy.)

packages/ckeditor5-dev-build-tools/src/plugins/bundleCss.ts
```diff
@@ -204,158 +202,166 @@ export function bundleCss( pluginOptions: RollupBundleCssOptions ): Plugin {
...
+								return virtualEntry;
+							}
 
-						// Fallback to raw filesystem reads when transform() did not run.
-						this.addWatchFile( normalizedPath );
+							const normalizedPath = normalizeId( filePath );
 
-						return fs.readFileSync( normalizedPath, 'utf-8' );
-					},
+							this.addWatchFile( normalizedPath );
```
przemyslaw-zan, line 300: We should add a test that runs `generateBundle()` over a fixture with a nested `@import` and asserts `addWatchFile()` was called with both the entry CSS path and the nested file's path.

przemyslaw-zan (changes requested):

packages/ckeditor5-dev-build-tools/tests/plugins/bundleCss/bundleCss.test.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/3399eeb30c5d9af0e99d5485018c3e9aacf90d2d

packages/ckeditor5-dev-build-tools/tests/plugins/bundleCss/bundleCss.test.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/3399eeb30c5d9af0e99d5485018c3e9aacf90d2d

packages/ckeditor5-dev-build-tools/tests/plugins/bundleCss/bundleCss.mocked.test.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/3399eeb30c5d9af0e99d5485018c3e9aacf90d2d

packages/ckeditor5-dev-build-tools/src/plugins/bundleCss.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/3399eeb30c5d9af0e99d5485018c3e9aacf90d2d

packages/ckeditor5-dev-build-tools/tests/plugins/bundleCss/bundleCss.test.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/3399eeb30c5d9af0e99d5485018c3e9aacf90d2d

packages/ckeditor5-dev-build-tools/tests/build/build.test.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/3399eeb30c5d9af0e99d5485018c3e9aacf90d2d

packages/ckeditor5-dev-build-tools/src/plugins/bundleCss.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/3399eeb30c5d9af0e99d5485018c3e9aacf90d2d

packages/ckeditor5-dev-build-tools/src/plugins/bundleCss.ts
```diff
@@ -50,56 +53,76 @@ function normalizeId( id: string ): string {
...
+
+	for ( const moduleId of moduleIds ) {
+		for ( const importedId of getModuleInfo( moduleId )?.dynamicallyImportedIds || [] ) {
+			if ( importedId in chunk.modules ) {
+				dynamicImportTargets.add( importedId );
+			}
+		}
+	}
+
+	return moduleIds.filter( moduleId => moduleId === chunk.facadeModuleId || dynamicImportTargets.has( moduleId ) );
```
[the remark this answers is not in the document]
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/e7275ad4942e1d6fbe8df499289c1fbc2cc7edd6

przemyslaw-zan (approved):

PR #1448, opened 2026-07-28
Discover manual tests in the root-level `manual/` directory

filipsobol:
### 🚀 Summary

`@ckeditor/ckeditor5-dev-manual-server` now discovers manual tests in a package's root-level `manual/` directory instead of `tests/manual/`.

The `manualStaticAssetsPlugin()` plugin has been removed. Static assets are served with Vite's built-in `publicDir` option instead.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-internal/issues/4540

---

### 💡 Additional information

None.


filipsobol:
I think that the lower coverage is fine, since it's caused by deleting code, not introducing new code without test coverage.

martnpaneq (approved):

PR #1451, opened 2026-07-29
Fix manual test pages missing package theme styles and the test header covering sticky editor UI

filipsobol:
### 🚀 Summary

Fixes two issues in the Vite-based manual test server:

* The manual test entry script transform now injects `theme/index-editor.css` and `theme/index-content.css` (each only when present), restoring package styles on manual test pages (broken style/table dropdowns, missing show-blocks outlines, unstyled pagination, minimap, code block labels, editor chrome, and more).
* The header bar now follows normal document flow instead of being fixed.

---

### 📌 Related issues

* https://github.com/ckeditor/ckeditor5-commercial/issues/11139

---

### 💡 Additional information

N/A

### DIFF
--- a/packages/ckeditor5-dev-manual-server/theme/manual-bootstrap.ts
+++ b/packages/ckeditor5-dev-manual-server/theme/manual-bootstrap.ts
@@ -23,6 +23,8 @@ const globalTarget = window as any;
  *
  * - sets the global license key so tests do not have to pass `licenseKey` explicitly;
  * - resets the `window.editor` named DOM property and auto-attaches the CKEditor inspector;
+ * - offsets the editor viewport below the fixed `<ck-manual-header>` bar (if the page has one),
+ *   so sticky toolbars, menu bars, and balloons are not covered by the test chrome when scrolling;
  * - renders the "source changed" refresh prompt driven by the dev server (`refreshPlugin`).
  *
  * The refresh prompt is part of the dev-server contract, not test chrome, so it belongs here.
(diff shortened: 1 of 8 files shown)


packages/ckeditor5-dev-manual-server/theme/manual-bootstrap.ts
```diff
@@ -61,6 +64,35 @@ function autoAttachInspector(): void {
...
+const viewportOffsetEditors = new WeakSet<object>();
+
+/**
+ * Pushes the editor's viewport offset below the fixed `<ck-manual-header>` bar so sticky
+ * toolbars, menu bars, and balloon panels are not covered by the test chrome when the page
+ * is scrolled. Pages without the header chrome are left untouched, and an offset configured
+ * by the test itself is preserved (the header height is added on top of it).
+ */
+function offsetViewportBelowManualHeader( editor: any ): void {
+	const header = document.querySelector( 'ck-manual-header' );
```
martnpaneq: This looks hacky. Is there no other way to style the editor in manual tests with CSS only?

filipsobol: I updated the test header to follow normal document flow instead of being fixed. This should resolve the original issue in a better way.

packages/ckeditor5-dev-manual-server/theme/manual-instructions.css
```diff
@@ -11,12 +11,20 @@
...
+/*
+ * `:where()` keeps zero specificity so anything a test styles itself takes precedence.
+ */
+:where(*, *::before, *::after) {
+	box-sizing: border-box;
+}
+
+:where(body) {
+	margin: 0;
+	min-height: 100vh;
```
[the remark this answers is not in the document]
filipsobol: Specificity is compared before source order, so the adopted stylesheet only wins ties against other zero-specificity selectors — any `body { ... }`, class, or id rule in a test still overrides the `:where()` reset. A tie could only occur for a bare `*` / `:where()` rule in a test, and the only such rule in the whole suite (`* { box-sizing: border-box }` in one RTC manual test) sets the same value as the reset. The reset applying to the whole page is intentional — it is part of the new convention where manual tests keep `<body>` unconstrained and build column layouts with a wrapper element instead.

martnpaneq:
Looks like this page needs some further fixing: http://localhost:8125/external/ckeditor5/packages/ckeditor5-editor-classic/manual/stickypanel.manual.html

image

martnpaneq:
And have a look at the sticky header here - I'm not sure if it behaves correctly: http://localhost:8125/external/ckeditor5/packages/ckeditor5-ui/manual/tickets/9892/1.manual.html

image

martnpaneq:
Also the manual test nav-bar does not look right here: http://localhost:8125/external/ckeditor5/packages/ckeditor5-ui/manual/poweredby/poweredby.manual.html

image

filipsobol:
All three issues are fixed. Make sure to use the branch from PR https://github.com/ckeditor/ckeditor5-commercial/pull/11184 in the commercial repository.

martnpaneq:
I can confirm the 3 issues reported above are fixed. I had an outdated branch in `cke5-commercial`.

martnpaneq (approved):

PR #1456, opened 2026-08-04
Add the typedoc-plugin-hierarchy-fixer plugin restoring the class hierarchy in API docs.

pomek:
### 🚀 Summary

Added the `typedoc-plugin-hierarchy-fixer` plugin. It fixes the API docs data for classes extending a mixin constant:

```ts
const ClassicEditorBase = ElementApiMixin( Editor );

export class ClassicEditor extends ClassicEditorBase {}
```

TypeDoc does not document the non-exported `*Base` constant, so the hierarchy breaks: `Editor` loses its "Subclasses", `ClassicEditor` shows a dead `ClassicEditorBase` entry, and mixin interfaces are missing from "Implements".

The plugin reads the real base type from the TypeScript checker (here: `Editor & ElementApi`) and:

* points `extendedTypes` at the actual base class and adds the class to its `extendedBy`,
* adds mixin interfaces to `implementedTypes`, also for base-less mixins like `class Plugin extends ObservableMixin()`,
* skips bases from outside the project, like `extends Error`,
* after the private API purge, removes dangling `extendedBy`/`implementedBy` references, so private class names do not leak into "Subclasses".

Ordering matters: the plugin must run after `typedoc-plugin-event-inheritance-fixer`, which resolves inherited events through the original (broken) references. Running it earlier strips the `set:{property}`/`change:{property}` events from classes like `Command`.

---

### 📌 Related issues

* Closes https://github.com/cksource/umberto/issues/1140

---

### 💡 Additional information

Verified on a full CKEditor 5 docs build: 69 of 72 broken references fixed. The remaining 3 are `extends Error`, left as is. `Editor` lists all five editors as subclasses, `ClassicEditor` renders `Editor → ClassicEditor`, and `Plugin` shows `Observable` in "Implements".

### DIFF
--- a/.changelog/20260804133745_umberto_1140_restore_subclasses_in_api_docs.md
+++ b/.changelog/20260804133745_umberto_1140_restore_subclasses_in_api_docs.md
@@ -1,0 +1,7 @@
+---
+type: Fix
+scope:
+  - typedoc-plugins
+---
+
+Restored the class hierarchy in the generated API documentation data for classes extending an intermediate mixin constant, for example `const ClassicEditorBase = ElementApiMixin( Editor )`. The new `typedoc-plugin-hierarchy-fixer` plugin replaces the unresolvable base class reference with the actual base class, restores the "Subclasses" relation on the base class page, and lists the mixin interfaces in the "Implements" section. References to reflections removed by the private API purge mechanism no longer leak into the documentation output.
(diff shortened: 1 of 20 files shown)


pomek:
**Before:**

https://ckeditor.com/docs/ckeditor5/48.3.1/api/module_core_editor_editor-Editor.html

image

https://ckeditor.com/docs/ckeditor5/48.3.1/api/module_core_plugin-Plugin.html

image

**After:**

https://127.0.0.1:8080/ckeditor5/48.3.1/api/module_core_editor_editor-Editor.html

image

https://127.0.0.1:8080/ckeditor5/48.3.1/api/module_core_plugin-Plugin.html

image

psmyrek (approved):

psmyrek (approved):

PR #1459, opened 2026-08-10
Replace PO translation tooling with TypeScript modules

filipsobol:
### 🚀 Summary

Replaces PO translation handling in `ckeditor5-dev-build-tools` and `ckeditor5-dev-translations` with generated TypeScript modules.

Adds a `translationsTypeImportSource` synchronization option so integrator packages can import the generated `Translations` type from their existing public dependency.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-internal/issues/4552.

---

### 💡 Additional information

N/A


packages/ckeditor5-dev-translations/lib/utils/translationfile.js
```diff
@@ -0,0 +1,146 @@
...
+ * dictionary: Record<string, string|Array<string>>,
+ * getPluralForm?: Function,
+ * preamble: string,
+ * translationsTypeImportSource: string
+ * }>}
+ */
+export async function readTranslationFile( filePath ) {
+	const content = fs.readFileSync( filePath, 'utf-8' );
+	const { preamble, translationsTypeImportSource } = readTranslationFileMetadata( filePath, content );
+	const { default: translations } = await import( pathToFileURL( filePath ).href );
```
przemyslaw-zan, line 30: This loads the dictionary with a bare `await import( pathToFileURL( filePath ).href )`. The Node ESM cache is permanent per URL. I reproduced two effects:

- Write V1, read, write V2, then read again. The second read returns V1.
- `r1.dictionary === r2.dictionary`. Every caller shares one mutable object, and a mutation through one result appears in a later read.

Commit `a9fe4811` fixed only the metadata half. Lines 28-29 re-read the preamble and `translationsTypeImportSource` fresh through `oxc-parser`. So one call can now return the new metadata of a file paired with its old dictionary.

The shipped flows survive **by accident**. Sync reads each file strictly before its single write. `movetranslationsbetweenpackages.js:53-54` stays consistent with the disk across several config entries only because it mutates the cached module dictionary in place, which keeps the cache and the disk in undocumented lockstep.

Two reasons I would fix this before release rather than after. `readTranslationFile` is a public export at `lib/index.js:9`, and a consumer already calls it. The invariant above is undocumented and untested, so a later refactor of move to build a fresh dictionary, a watch task, or a `synchronizeTranslations` call followed by `moveTranslations` on overlapping packages, silently reverts or drops translations.

One caution for the fix. A change that clones the dictionary but does not bust the cache corrupts multi-entry moves, and no test catches that today. The clean fix is to extract the dictionary with `oxc-parser`, which the function already loads for the metadata, instead of evaluating the module.

The same mechanism affects `packages/ckeditor5-dev-build-tools/src/plugins/translations.ts:106` in a watch build or any long-lived process. Master read the `.po` files with `readFileSync` on every `generateBundle`, so each build saw fresh content. Exposure is narrow, because translation files change rarely, but the regression is mechanical.

packages/ckeditor5-dev-translations/lib/utils/movetranslationsbetweenpackages.js
```diff
@@ -5,84 +5,92 @@
...
+			const sourceTranslations = await readTranslationFile( sourceTranslationFilePath );
+			const destinationTranslations = destinationTranslationFileExists ?
+				await readTranslationFile( destinationTranslationFilePath ) :
+				{
+					language: sourceTranslations.language,
+					dictionary: {},
+					preamble: sourceTranslations.preamble,
+					translationsTypeImportSource: sourceTranslations.translationsTypeImportSource
+				};
+			const value = sourceTranslations.dictionary[ messageId ];
```
przemyslaw-zan, line 77: If a source language file lacks the moved message, the code assigns `undefined` into the destination dictionary, and `quote()` throws `Cannot read properties of undefined (reading 'replace')`. The throw happens inside the per-language loop, after earlier files were rewritten, and before either `contexts.json` is written. The tree is then half migrated with no record of what changed. This is **not a regression**, because the PO path on master fails the same way, so treat it as optional robustness work. The precondition is realistic, because sync never back-fills a context-only entry into a language file that lacks it. Either validate up front in `assertContextsExist`, which today checks only `contexts.json`, or map a missing value to `''`.

packages/ckeditor5-dev-translations/lib/utils/synchronizetranslationsbasedoncontext.js
```diff
@@ -5,133 +5,114 @@
...
 
-		// Then, for each translation file in a package:
 		for ( const translationFilePath of translationFilePaths ) {
-			const { translations, translationFile } = parseTranslationFile( translationFilePath );
-
-			// (5.1) Update file headers.
-			const { languageCode, localeCode } = languages.find( language => language.localeCode === translations.headers.Language );
+			const originalFile = fs.readFileSync( translationFilePath, 'utf-8' );
+			const { language, dictionary, preamble } = await readTranslationFile( translationFilePath );
+			const { languageCode } = languages.find( item => item.languageFileName === language );
```
przemyslaw-zan: `synchronizetranslationsbasedoncontext.js:55` and `movetranslationsbetweenpackages.js:83-84`. An unsupported language key destructures the `undefined` result of `Array.prototype.find`, and the message names neither the file nor the language. This is reachable when somebody hand-creates a locale file before the locale enters `SUPPORTED_LOCALES`, or when a stray `.ts` file lands in `lang/translations`, because the glob picks up any `*.ts`.

packages/ckeditor5-dev-build-tools/src/plugins/translations.ts
```diff
@@ -89,53 +64,49 @@ function getEsmCode( code: string ): string {
...
-					.filter( Boolean )
-					// Map files to desired structure.
-					.map( content => ( {
-						dictionary: getDictionary( content ),
-						getPluralForm: getPluralFunction( content )
-					} ) );
+				const translations: Array<Translation> = await Promise.all( paths.map( async filePath => {
+					const translationModule = await import( pathToFileURL( filePath ).href );
+
+					return translationModule.default[ language ] as Translation;
```
przemyslaw-zan, line 112: The plugin takes the language from the filename and then indexes `translationModule.default[ language ]` with no check. On a key and filename mismatch it emits `{"en":{"dictionary":undefined}}`, which is valid JavaScript, so the build succeeds and `t()` crashes at runtime instead. A file with no default export fails with `Cannot read properties of undefined (reading 'pl')` and no path. `this.error()` with the file path would cover both.

packages/ckeditor5-dev-build-tools/src/plugins/translations.ts
przemyslaw-zan, line 15: The TYPINGS constant hardcodes `import type { Translations } from '@ckeditor/ckeditor5-utils'` in every emitted `<lang>.d.ts`, with no option matching the new `translationsTypeImportSource`. A generated package declares neither a dependency nor a peer dependency on that package, so the published typings point at something a strict pnpm layout may not resolve. This matches master, so it is pre-existing, but it is now inconsistent with configurable sources.

packages/ckeditor5-dev-build-tools/src/plugins/translations.ts
```diff
@@ -89,53 +64,49 @@ function getEsmCode( code: string ): string {
...
 		name: 'cke5-translations',
 
 		async generateBundle() {
-			// Get the paths to the PO files based on provided pattern.
+			// Get the paths to the translation files based on provided pattern.
 			const filePaths = await glob( options.source, {
 				cwd: process.cwd(),
-				ignore: 'node_modules/**'
+				absolute: true,
+				ignore: [ '**/node_modules/**', '**/dist/**', '**/*.d.ts' ]
```
przemyslaw-zan, line 97: The glob now always ignores `**/node_modules/**`. On master the pattern was relative and never matched the absolute paths that an absolute `source` produces, so an umbrella build could gather translations from installed packages. That now matches zero files with no warning. Node also refuses to type-strip a `.ts` file under `node_modules`, so the capability is gone by construction. A warning on a zero-match `source` would help, and the changelog should mention it.

packages/ckeditor5-dev-translations/lib/utils/translationfile.js
```diff
@@ -0,0 +1,146 @@
...
+		'\t\tdictionary: {'
+	);
+
+	const entries = Object.keys( contexts )
+		.filter( messageId => Object.hasOwn( dictionary, messageId ) );
+
+	entries.forEach( ( messageId, index ) => {
+		const context = contexts[ messageId ];
+
+		for ( const line of context.split( /\r?\n/ ) ) {
```
przemyslaw-zan, line 81: A context that holds U+2028, U+2029 or a lone CR produces an **unparseable** file. All three are ECMAScript line terminators, so they end the line comment and the rest of the context becomes code. `quote()` escapes these characters for keys and values, but the comment path does not. Contexts are hand-authored English, so this is exotic, but a serializer should never emit a module it cannot read back. Splitting on those characters at lines 81-83 fixes it.

packages/ckeditor5-dev-translations/lib/utils/synchronizetranslationsbasedoncontext.js
```diff
@@ -5,133 +5,114 @@
...
-
-		// (5) Prepare English translations needed to check whether their wording have been changed in the source messages.
-		const englishTranslationFilePath = translationFilePaths.find( filePath => filePath.endsWith( 'en.po' ) );
-		const { translations: englishTranslations } = parseTranslationFile( englishTranslationFilePath );
-
+		const translationFilePaths = glob.sync( upath.join( packagePath, TRANSLATION_FILES_PATH, '*.ts' ) );
+		const englishTranslationFilePath = translationFilePaths.find( filePath => filePath.endsWith( 'en.ts' ) );
+		const { dictionary: englishTranslations } = await readTranslationFile( englishTranslationFilePath );
 		const changedEnglishTranslations = getChangedEnglishTranslations( englishTranslations, sourceMessagesForPackage );
+		const isCorePackage = upath.basename( packagePath ) === 'ckeditor5-core';
```
przemyslaw-zan: Core-package detection is `upath.basename( packagePath ) === 'ckeditor5-core'`, duplicated in `synchronizetranslationsbasedoncontext.js:50`, `createmissingpackagetranslations.js:26` and `movetranslationsbetweenpackages.js:84`. `synchronizeTranslations` already receives and validates `corePackagePath` but never threads it into the write path. This works in both current consumers. A differently named core directory would silently lose every plural function with no error.

przemyslaw-zan (commented):
### Changelog

The build-tools entry says only that the build tools now collect TypeScript translation sources. It omits four things an upgrading integrator needs:

- The `source` default changed from `**/*.po` to `**/lang/translations/*.ts`.
- The module contract: a default export keyed by the language code, and the key must match the filename.
- The dist-asset change: a dictionary-only source emits no `getPluralForm`, no `"getPluralForm":null` appears in ESM output, and UMD no longer overwrites a registered function.
- Sources load through a native `import()` of TypeScript.

The entry is correctly typed as a major breaking change and does state the input break. The runtime consequence is documented for integrators in two guides. So this is a completeness gap, and one or two sentences close it. Migration steps for packages that still hold `.po` files belong here too, and I raised that in the issue because it needs matching changes elsewhere.

The dev-translations entry has a factual error. It lists `readTranslationFile()` among the functions that are "now asynchronous", but that function is new in this PR. `serializeTranslationFile()` is not mentioned at all. `packages/ckeditor5-dev-translations/README.md:13-17` also still lists only the three old functions.


### Tests

- `tests/utils/translationfile.js` has no escaping test, no byte round-trip test, and no rewrite-then-reread test. The last one would have caught item 1 immediately. No test input holds a quote, a backslash or a newline inside a key or value, so `quote()` has zero behavioral cover. `skipLicenseHeader: true`, `preamble: ''`, a file with no default export, and a multi-line context are all untested.
- Every move test uses a single-entry config, at `tests/utils/movetranslationsbetweenpackages.js` lines 35, 64, 93 and 110. So the multi-entry path, whose correctness depends on the accidental aliasing in item 1, has no cover. Neither does a move of a message that one language file lacks.
- Four assertions from the old suite are gone: a changed English **plural** form, the no-op write skip at `synchronizetranslationsbasedoncontext.js:98`, a move that overwrites an existing destination entry, and any unmocked create-missing to sync path. That last one is the fresh-package scenario, which is exactly what a community migration hits.
- `tests/utils/synchronizetranslationsbasedoncontext.js:80-84` passes only because vitest equality ignores a property whose value is `undefined`, so the import-source override is unasserted.
- `tests/plugins/translations/translations.ts:52` asserts that the output lacks `translations/ignored.js`. The plugin groups by `parse( path ).name`, and for `ignored.d.ts` that is `ignored.d`, so the asserted name can never appear and the assertion passes either way. The exclusion is still covered indirectly, because importing the fixture would throw. Assert `translations/ignored.d.js`, or assert the emitted file set exactly.

packages/ckeditor5-dev-translations/lib/utils/translationfile.js
filipsobol: The report is valid, but we do not need to worry about it here: these commands run once per process, do not use HMR, and do not reread a file after rewriting it.

packages/ckeditor5-dev-build-tools/src/plugins/translations.ts
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/fe0c0c7993d18cb20be7154db0d6b30cd2e5e388.

packages/ckeditor5-dev-translations/lib/utils/synchronizetranslationsbasedoncontext.js
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/fe0c0c7993d18cb20be7154db0d6b30cd2e5e388.

packages/ckeditor5-dev-translations/lib/utils/synchronizetranslationsbasedoncontext.js
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/fe0c0c7993d18cb20be7154db0d6b30cd2e5e388.

packages/ckeditor5-dev-translations/lib/utils/movetranslationsbetweenpackages.js
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/fe0c0c7993d18cb20be7154db0d6b30cd2e5e388.

packages/ckeditor5-dev-build-tools/src/plugins/translations.ts
filipsobol: Not addressed here: this behavior predates this PR and remains outside its scope.

packages/ckeditor5-dev-build-tools/src/plugins/translations.ts
filipsobol: Not addressed here: this is existing intentional behavior outside this PR scope.

packages/ckeditor5-dev-translations/lib/utils/translationfile.js
filipsobol: Not addressed here: this is an exotic context-only edge case outside this PR scope.

packages/ckeditor5-dev-translations/lib/utils/movetranslationsbetweenpackages.js
```diff
@@ -5,84 +5,127 @@
...
+	}
+}
 
-		fs.writeFileSync(
-			destinationPackageContext.contextFilePath,
-			JSON.stringify( destinationPackageContext.contextContent, null, '\t' ),
-			'utf-8'
-		);
+function writePackageTranslation( packagePath, filePath, translations, contexts ) {
+	const language = getTranslationLanguage( filePath, translations );
```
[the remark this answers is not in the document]
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/commit/de6bf6db.

psmyrek (approved):

PR #1466, opened 2026-08-18
Verify GitHub token permissions before releases

### LINKED ISSUE #20214: [dev-release-tools] Verify the GitHub token instead of validating its length
## 📝 Provide a description of the improvement[/+]

[`provideToken()`](https://github.com/ckeditor/ckeditor5-dev/blob/0c15d0ae0fde790e23a0876c75563057ea1f3b19/packages/ckeditor5-dev-release-tools/lib/utils/providetoken.js#L18-L20) validates the GitHub token by counting its characters:

**It forces releasers onto an over-privileged token.** 40 is the length of a legacy hex PAT and – coincidentally – of a modern classic `ghp_` token. Everything else GitHub issues is rejected:

| Token | Length | Accepted? |
| --- | --- | --- |
| Legacy hex PAT | 40 | ✅ |
| Classic PAT (`ghp_…`) | 40 | ✅ |
| Fine-grained PAT (`github_pat_…`) | ~93 | ❌ |
| App installation token (`ghs_…`, the Actions `GITHUB_TOKEN`) | variable | ❌ |

### Definition of Done

- A fine-grained PAT with `Contents: Read and write` on the released repository is accepted.
- A token that authenticates but lacks write access is rejected **at the prompt**, naming the repository – not at `createRelease()`.
- Rejection messages state the cause.

### Technical details

Replace the length check with an authorization check against the API:

```js
async function validateToken( token, { owner, repo } ) {
	const value = token.trim();

	if ( !value ) {
		return 'The token cannot be empty.';
	}

	const { data } = await new Octokit( { auth: value } ).repos.get( { owner, repo } );

	if ( !data.permissions?.push ) {
		return `The token has no write access to the "${ owner }/${ repo }" repository.`;
	}

	return true;
}
```

`repos.get()` returns `permissions` for classic and fine-grained tokens alike, and an unusable token fails with a `401`.

filipsobol:
### 🚀 Summary

Replaces GitHub token length validation with an API preflight against the target repository. Both prompted and environment-provided tokens are checked for repository access and `Contents: write` permission before release tasks begin.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/20214

---

### 💡 Additional information

The write-permission probe uses GitHub’s generate-release-notes endpoint, which computes and returns proposed notes without persisting a release, tag, or other repository data. The validator was also exercised successfully in CI in [#1467](https://github.com/ckeditor/ckeditor5-dev/pull/1467).

### DIFF
--- a/packages/ckeditor5-dev-release-tools/lib/utils/providetoken.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/providetoken.js
@@ -4,23 +4,32 @@
  */
 
 import inquirer from 'inquirer';
+import validateGithubToken from './validategithubtoken.js';
 
 /**
  * Asks a user for providing the GitHub token.
  *
+ * @param {object} [options]
+ * @param {string} [options.cwd=process.cwd()] Current working directory from which the repository will be resolved.
  * @returns {Promise.<string>}
  */
-export default async function provideToken() {
+export default async function provideToken( options ) {
 	const tokenQuestion = {
 		type: 'password',
 		name: 'token',
 		message: 'Provide the GitHub token:',
-		validate( input ) {
-			return input.length === 40 ? true : 'Please provide a valid token.';
+		async validate( input ) {
+			try {
+				await validateGithubToken( input, options );
+
+				return true;
+			} catch ( error ) {
+				return error.message;
+			}
 		}
 	};
 
 	const { token } = await inquirer.prompt( [ tokenQuestion ] );
 
-	return token;
+	return token.trim();
 }
--- a/packages/ckeditor5-dev-release-tools/lib/utils/providetoken.js
+++ b/packages/ckeditor5-dev-release-tools/lib/utils/providetoken.js
@@ -13,14 +13,14 @@ import validateGithubToken from './validategithubtoken.js';
  * @param {string} [options.cwd=process.cwd()] Current working directory from which the repository will be resolved.
  * @returns {Promise.<string>}
  */
-export default async function provideToken( options ) {
+export default async function provideToken( { cwd = process.cwd() } = {} ) {
 	const tokenQuestion = {
 		type: 'password',
 		name: 'token',
 		message: 'Provide the GitHub token:',
 		async validate( input ) {
 			try {
-				await validateGithubToken( input, options );
+				await validateGithubToken( input, { cwd } );
 
 				return true;
 			} catch ( error ) {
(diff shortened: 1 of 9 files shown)


packages/ckeditor5-dev-release-tools/lib/utils/providetoken.js
```diff
@@ -4,23 +4,32 @@
...
 
 /**
  * Asks a user for providing the GitHub token.
  *
+ * @param {object} [options]
+ * @param {string} [options.cwd=process.cwd()] Current working directory from which the repository will be resolved.
  * @returns {Promise.<string>}
  */
-export default async function provideToken() {
+export default async function provideToken( options ) {
```
pomek:
```suggestion
export default async function provideToken( { cwd = process.cwd() } = {} ) {
```

This way, the code will match the corresponding JSDoc annotation.

packages/ckeditor5-dev-release-tools/lib/utils/validategithubtoken.js
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Verifies that a GitHub token can create a release in the current repository.
+ *
+ * @param {string} token Token used to authenticate with GitHub.
+ * @param {object} [options]
+ * @param {string} [options.cwd=process.cwd()] Current working directory from which the repository will be resolved.
+ * @returns {Promise.<string>} The normalized token.
+ */
+export default async function validateGithubToken( token, { cwd = process.cwd() } = {} ) {
```
pomek:
```suggestion
/**
 * Verifies that a GitHub token can create a release in the current repository.
 *
 * @param {string} token Token used to authenticate with GitHub.
 * @param {object} options
 * @param {string} options.cwd=process.cwd() Current working directory from which the repository will be resolved.
 * @returns {Promise.<string>} The normalized token.
 */
export default async function validateGithubToken( token, { cwd } ) {
```

Since this util is a _sub-util_ for `provideToken()`, I would either:

- move the default arguments resolution a layer higher,
- duplicate resolving the default arguments.

Both of my comments follow the first point.

packages/ckeditor5-dev-release-tools/lib/utils/providetoken.js
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/pull/1466/changes/2290793c1a667b785ee1c750fa3eedf18991d41a.

packages/ckeditor5-dev-release-tools/lib/utils/validategithubtoken.js
filipsobol: Addressed in https://github.com/ckeditor/ckeditor5-dev/pull/1466/changes/2290793c1a667b785ee1c750fa3eedf18991d41a.

pomek (approved):
LGTM.

PR #1470, opened 2026-08-20
Preserve root workspace dependencies during releases

filipsobol:
### 🚀 Summary

Fix the release preparation flow so that it converts workspace dependencies only in package copies under `release/*`, while preserving `workspace:*` dependencies in the repository root.

`updateDependencies()` previously always included `cwd/package.json`, even when `packagesDirectory` pointed to `release`. As a result, the v61 release changed the root `@ckeditor/ckeditor5-dev-changelog` dependency from `workspace:*` to `^61.0.0` without regenerating the lockfile. Updating only the lockfile would hide that release-script issue and would violate the syncpack rule requiring local packages to use the workspace protocol.

This PR:

- adds an optional `includeCwd` setting to `updateDependencies()`, defaulting to `true` for backward compatibility;
- sets `includeCwd: false` in the ckeditor5-dev release flow;
- restores the root manifest and lockfile specifier to `workspace:*`;
- adds regression tests and a changelog entry.

The release copies are still converted from `workspace:*` to the released version range before publication. The root package version is also still updated by the separate `updateVersions()` step.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5-commercial/issues/11395

---

### 💡 Additional information

Consumer impact was checked in both `ckeditor/ckeditor5` and `ckeditor/ckeditor5-commercial`. Their release scripts omit `includeCwd`, so the default value preserves their existing behavior. Their root manifests also contain no eligible local CKEditor package dependency that would be affected by this option. No changes are required in either repository.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -24,7 +24,7 @@ importers:
   .:
     devDependencies:
       '@ckeditor/ckeditor5-dev-changelog':
-        specifier: workspace:*
+        specifier: ^61.0.0
         version: link:packages/ckeditor5-dev-changelog
       '@inquirer/prompts':
         specifier: ^7.10.1
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -24,7 +24,7 @@ importers:
   .:
     devDependencies:
       '@ckeditor/ckeditor5-dev-changelog':
-        specifier: ^61.0.0
+        specifier: workspace:*
         version: link:packages/ckeditor5-dev-changelog
       '@inquirer/prompts':
         specifier: ^7.10.1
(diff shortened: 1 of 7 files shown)


pnpm-lock.yaml
pomek, line 1: Since the release process does not update the lockfile, maybe we should update `devDeps` of the root package?

https://github.com/ckeditor/ckeditor5-dev/blob/6c1ea622f56b708206b7e911d2487a8737634544/package.json#L19

pomek (approved):
LGTM.

pnpm-lock.yaml
filipsobol: I updated the implementation to fix the actual root cause. The root `package.json` should not have been updated during the release process in the first place.

pomek: That's awesome. I am a bit sad due to the increased scope, but we will finally solve it properly. Thanks!

PR #1473, opened 2026-08-26
Update dependencies.

martnpaneq:
### 🚀 Summary

Routine dependency maintenance in `pnpm-workspace.yaml`.

Raised existing override values so the pinned versions are current:

* `brace-expansion@^1` → `^1.1.18`, `brace-expansion@^2` → `^2.1.4`, `brace-expansion@^5` → `^5.0.9`
* `js-yaml@^3` and `js-yaml@^4` → `^4.3.1`

Added two new overrides:

* `'nanoid@^3': '^3.3.17'` — reaches `postcss > nanoid` under `vite` in `ckeditor5-dev-manual-server`.
* `'undici@^6': '^6.28.0'` — reaches `pacote > @npmcli/run-script > node-gyp > undici`.

Housekeeping — dropped seven overrides that proved redundant: `diff@^7`, `lodash@^4`, `serialize-javascript@6`, `ip-address@^10`, `postcss@^8`, `pacote@^21` and `tar@^7`. With them removed nothing downgrades (`diff@8.0.4`, `ip-address@10.3.1`, `pacote@21.5.1`, `tar@7.5.22` unchanged; `postcss` actually resolves higher, at `8.5.26`), and `lodash` / `serialize-javascript` are no longer in the graph at all.

`@puppeteer/browsers@^2` was tested the same way and **kept** — removing it lets `@puppeteer/browsers@2.13.2` back in under `puppeteer`, along with its older `extract-zip@2.0.1` dependency. The override is what keeps the tree on the `3.x` line.

---

### 📌 Related issues

See: https://github.com/ckeditor/ckeditor5-internal/issues/4684

---

### 💡 Additional information

* **Changelog intentionally omitted** — dependency-only update, nothing user-facing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@ckeditor/ckeditor5-dev-changelog": "workspace:*",
     "@inquirer/prompts": "^7.10.1",
     "@listr2/prompt-adapter-inquirer": "^2.0.22",
-    "@vitest/coverage-v8": "^4.1.2",
+    "@vitest/coverage-v8": "^4.1.11",
     "@vitest/eslint-plugin": "^1.6.24",
     "eslint": "^10.0.0",
     "eslint-config-ckeditor5": "^20.0.0",
@@ -37,7 +37,7 @@
     "typescript": "5.5.4",
     "upath": "^2.0.1",
     "vite": "^8.2.1",
-    "vitest": "^4.1.2"
+    "vitest": "^4.1.11"
   },
   "scripts": {
     "nice": "ckeditor5-dev-changelog-create-entry",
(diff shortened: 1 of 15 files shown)


pomek (commented):
LGTM. One small addition worth folding in here: the lockfile still resolves `vitest` / `@vitest/coverage-v8` at `4.1.7`, which Snyk flags (dev scope). The manifests already allow it, so a plain `pnpm update vitest @vitest/coverage-v8` should be enough.

martnpaneq:
Done — `vitest` and `@vitest/coverage-v8` are at `4.1.11`, applied with `pnpm update -r` so all 13 packages declaring it move together. Full test run passes (no runner split), and `snyk test --all-projects --dev` is clean. Good catch: I had been scanning without `--dev`, so I'd have missed this — there are dev-scope findings in ckeditor5-builder and ckeditor5-collaboration-samples too, which look outside this report's scope. Happy to track those separately if you want them covered.

pomek (approved):
LGTM.

PR #1476, opened 2026-09-02
Fix stale manual test HTML after edits

filipsobol:
### 🚀 Summary

Updates Vite to 8.2.2 and Rolldown to 1.2.6.

The manual test server now relies on Rolldown to re-emit changed HTML assets, so it no longer patches Vite's in-memory HTML or intercepts Rolldown full reloads. It retains the targeted JavaScript refresh prompt and force-ships body-only HTML changes. The bundled development integration test verifies that both `<head>` and `<body>` are regenerated.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-commercial/issues/11491

---

### 💡 Additional information

N/A

### DIFF
--- a/.changelog/20260902123108_cc_11491_fix_stale_manual_html_head.md
+++ b/.changelog/20260902123108_cc_11491_fix_stale_manual_html_head.md
@@ -1,0 +1,7 @@
+---
+type: Fix
+scope:
+  - ckeditor5-dev-manual-server
+---
+
+The manual test server now reloads complete HTML documents after edits, including changes made inside `<head>`.
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "syncpack": "^15.3.2",
     "typescript": "5.5.4",
     "upath": "^2.0.1",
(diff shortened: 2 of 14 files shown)


psmyrek (commented):
It works as expected 🎉 

Two things before merging it:

1. The integration test changes `<head>` and `<body>` at the same time. This does not independently protect the body-only path. I suggest adding a separate test which changes only the body, for example:

    ```js
    const newHtml = oldHtml.replace( '<h2>OLD</h2>', '<h2>NEW</h2>' );
    ```
    
    This would catch a future regression in the `hotUpdate()` logic.

2. This PR alone does not fix the https://github.com/ckeditor/ckeditor5-commercial/issues/11491 issue, so it would be clearer to just reference the issue here and let `#11492` close it after the new dev tools are released and consumed.

filipsobol:
Thanks. I added a new test that covers this and changed “Closes” to “See” in the PR description.

psmyrek (approved):

PR #1484, opened 2026-09-14
Store the worker module of `executeInParallel()` in a configurable project directory

### LINKED ISSUE #18939: [dev-release-tools] Put worker files in a directory that can be added to `.gitignore`
When using `executeInParallel` function, a worker file that is tracked by git is generated. It may happen that the script fails or is cancelled before finishing. In such case, the worker file is left behind:

```bash
11:57:55  ~/repositories/ckeditor5 master ● git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        6a410868-6bba-4db2-8abe-d05a323db00a.mjs

nothing added to commit but untracked files present (use "git add" to track)
```

We could generate worker files inside some directory, eg `.workerFiles` that could be then added to `.gitignore`, so the generated files are not tracked. I believe that adding a directory to this path would be enough:

https://github.com/ckeditor/ckeditor5-dev/blob/a5dd220271755a9694b74d549a814921ec4d59b5/packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js#L56

pomek:
### 🚀 Summary

The `executeInParallel()` function now stores the temporary worker module inside the project instead of in its root directory. The location is configurable via the new `workerDirectory` option, which is resolved from `cwd` and defaults to `build`.

Storing the module in a directory ignored by git prevents a leftover module from polluting the working tree when the process is aborted or fails.

---

### 📌 Related issues

* Closes https://github.com/ckeditor/ckeditor5/issues/18939

---

### 💡 Additional information

**Why not `os.tmpdir()`?** The worker module is an ES module that the worker imports dynamically. Every callback passed to `executeInParallel()` in our repositories does `await import( '@ckeditor/ckeditor5-dev-utils' )` or similar. Node resolves bare specifiers relative to the location of the importing file, so the module must live inside the project to reach its `node_modules`. A file in the OS temporary directory would fail with `ERR_MODULE_NOT_FOUND`.

**`.gitignore`:** `build/` was already ignored in `ckeditor5` and `ckeditor5-commercial`. This PR adds `/build/` to the `.gitignore` of this repository because `scripts/preparepackages.js` uses `executeInParallel()` from the repository root.

**Manual verification:** a real `executeInParallel()` run with a callback importing `@ckeditor/ckeditor5-dev-utils` and `upath` works with the default and a custom `workerDirectory`. After a `kill -9` mid-run, the module is left in `build/` and `git status` stays clean.


packages/ckeditor5-dev-release-tools/lib/utils/executeinparallel.js
```diff
@@ -31,6 +31,9 @@ const WORKER_SCRIPT = new URL( './parallelworker.js', import.meta.url );
  * @param {ExecuteInParallelPackagesDirectoryFilter|null} [options.packagesDirectoryFilter=null] An optional callback allowing filtering out
  * directories/packages that should not be touched by the task.
  * @param {string} [options.cwd=process.cwd()] Current working directory from which all paths will be resolved.
+ * @param {string} [options.workerDirectory='build'] A directory, relative to `options.cwd`, where temporary worker modules are stored.
+ * It must be inside the project so workers can resolve dependencies from `node_modules`.The default `build` directory is usually ignored
```
przemyslaw-zan:
```suggestion
 * It must be inside the project so workers can resolve dependencies from `node_modules`. The default `build` directory is usually ignored
```

.changelog/20260914105416_ck_18939_worker_files_directory.md
```diff
@@ -0,0 +1,11 @@
...
+type: Feature
+scope:
+  - ckeditor5-dev-release-tools
+closes:
+  - ckeditor/ckeditor5#18939
+---
+
+The `executeInParallel()` function now stores the temporary worker module inside the project instead of in its root directory. The location is configurable via the new `workerDirectory` option, which is resolved from `cwd` and defaults to `build`.
+
+The module must stay inside the project so that `import()` and `require()` calls in the executed callback still resolve dependencies from the project's `node_modules`. Storing it in a directory ignored by git prevents a leftover module from polluting the working tree when the process is aborted or fails.
```
przemyslaw-zan: `require()` is not applicable for `.mjs`, we can drop this from the changelog.

przemyslaw-zan (changes requested):
The new option is not exposed when running the `publishPackages` task.

https://github.com/ckeditor/ckeditor5-dev/blob/c18b1ef52d74bdfb872dbbfa5aad51c57f9b2f62/packages/ckeditor5-dev-release-tools/lib/tasks/publishpackages.js#L121-L131

.changelog/20260914105416_ck_18939_worker_files_directory.md
pomek:
```suggestion
The module must stay inside the project so that `import()` calls in the executed callback resolve dependencies from the project's `node_modules`. Storing it in a directory Git ignores prevents a leftover module from polluting the working tree if the process is aborted or fails.
```

przemyslaw-zan (approved):
Requests addressed, and changes already confirmed to work.

PR #1488, opened 2026-09-23
Fix favicon 404 errors on manual test pages

filipsobol:
### 🚀 Summary

Fix spurious `/favicon.ico` 404 errors on manual test pages by providing an inline favicon. This is a correctness fix.

---

### 📌 Related issues

* See https://github.com/ckeditor/ckeditor5-commercial/issues/11757.

---

### 💡 Additional information

The favicon is injected alongside the existing manual-page bootstrap, including pages without `<ck-manual-header>`. It uses a data URL, so no extra asset or HTTP request is needed. The catalog already has an inline favicon.


psmyrek (dismissed):
LGTM 🎉 

One suggestion: please add a short comment next to the new `<link>` tag explaining why it is there. Something like:

> Chrome requests `/favicon.ico` automatically when a page has no icon. The crawler reports the 404 as a response failure and retries the page. The data URL is only the PNG signature and prevents the request without adding an asset.

Without this, the tag with `iVBORw0KGgo=` value looks like a magic thing.

psmyrek (approved):

